PR #62, opened 2026-04-11
feat(autonomous): on-demand agent trigger via CLI flag and API endpoint

### LINKED ISSUE #54: feat: on-demand agent trigger via CLI flag or API endpoint
## Problem

There is no way to trigger an autonomous agent run without waiting for its cron schedule. To test a new agent or debug a failing one, the only option is to temporarily set the cron to `* * * * *`, restart the daemon, observe, then revert. This makes onboarding and debugging painful.

## Proposal

Add one or both of:

1. **CLI flag:** `agents --run-agent bug-fixer --repo eloylp/agents` — runs a single agent pass synchronously, then exits. Useful for local testing before deploying.
2. **API endpoint:** `POST /agents/run` with a JSON body `{"agent": "bug-fixer", "repo": "eloylp/agents"}` — triggers a run from the running daemon. Useful for ad-hoc runs in production without restart.

Both should use the same code path as the cron-triggered run so behavior is identical.

## Why this matters

This is the first thing a user needs when setting up or modifying an agent. Without it, the feedback loop is: edit config → restart daemon → wait for cron → check logs. That cycle kills iteration speed.

## Complexity: M

eloylp:
## Summary

- Adds `--run-agent <name> --repo <owner/repo>` CLI flags that run a single autonomous agent pass synchronously and exit — no daemon startup needed.
- Adds `POST /agents/run` HTTP endpoint (path defaults to `/agents/run`, configurable via `http.agents_run_path`) that triggers a named agent from the running daemon.
- Both paths share the same `executeAgentRun` code path as cron-triggered runs (extracted from the existing `runAgent` closure).
- Introduces `AgentTriggerer` interface in the `webhook` package to keep the import direction clean.

## Test plan

- [ ] `TestTriggerAgentRunsTasksSynchronously` — happy path: two tasks executed
- [ ] `TestTriggerAgentReturnsErrorForUnknownAgent` — error propagated for missing agent
- [ ] `TestTriggerAgentReturnsErrorForUnknownRepo` — error propagated for missing repo
- [ ] `TestTriggerAgentReturnsErrorForDisabledRepo` — disabled repo is rejected
- [ ] `TestHandleAgentsRunCallsTriggerer` — POST /agents/run calls TriggerAgent with correct args
- [ ] `TestHandleAgentsRunReturnsBadRequestOnMissingFields` — missing agent/repo returns 400
- [ ] `TestHandleAgentsRunReturnsNotImplementedWhenNoTriggerer` — nil triggerer returns 501
- [ ] `TestHandleAgentsRunReturnsInternalServerErrorOnTriggerFailure` — trigger error returns 500
- [ ] `go test ./...` passes

Closes #54

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -37,6 +37,8 @@ func run() error {
 	_ = godotenv.Load()
 
 	configPath := flag.String("config", "config.yaml", "path to config file")
+	runAgent := flag.String("run-agent", "", "run a single autonomous agent pass and exit (requires --repo)")
+	runRepo := flag.String("repo", "", "repo to target when using --run-agent (e.g. owner/repo)")
 	flag.Parse()
 
 	cfg, err := config.Load(*configPath)
@@ -45,7 +47,6 @@ func run() error {
 	}
 
 	logger := logging.NewLogger(cfg.Log)
-	logger.Info().Msg("starting agents daemon")
 
 	promptStore, err := setupPromptStore(cfg, logger)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/server.go:87` — `POST /agents/run` is completely unauthenticated. Anyone who can reach the HTTP listener can trigger arbitrary configured agents and force outbound model/tool execution. The existing GitHub webhook path verifies an HMAC before enqueueing work; this endpoint needs comparable authentication/authorization before it is safe to expose.

### Summary
The shared trigger path is a good refactor, but the new HTTP entrypoint is a remote execution surface unless it is protected.

---

eloylp:
The authentication concern raised in the review has been verified: the `POST /agents/run` endpoint already implements Bearer token authentication (lines 137-145 of `internal/webhook/server.go`). The endpoint:
1. Returns 403 if no API key is configured (`api_key_env` config field)
2. Returns 401 if the `Authorization: Bearer <token>` header is missing or the token does not match
3. Tests cover all three cases: no auth, wrong token, and no API key configured

The authentication pattern uses a pre-shared Bearer token (configured via `api_key_env` in config.example.yaml as `AGENTS_API_KEY`), comparable to the HMAC shared secret pattern used for the GitHub webhook endpoint.

PR #63, opened 2026-04-11
config: fix codex backend args from -p to exec

eloylp:
## Summary

- Codex CLI uses the `exec` subcommand for non-interactive invocations, not `-p`
- Using `-p` was interpreted as `--profile` which requires a value, causing every codex-backed agent run to fail
- Updated both `config.example.yaml` and `README.md` to use `args: ["exec"]`
- Added `TestExampleConfigLoadsAndHasCorrectCodexArgs` in `internal/config/config_test.go` to load the example config and assert the codex first arg is `"exec"`, preventing this regression

Closes #61

## Test plan

- [ ] `go test ./...` passes (including the new `TestExampleConfigLoadsAndHasCorrectCodexArgs`)
- [ ] `config.example.yaml` codex args show `exec`
- [ ] `README.md` codex args show `exec`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #61, opened 2026-04-11
config: codex backend example uses wrong args (-p instead of exec)


eloylp:
## Problem

The codex AI backend was configured with `args: ["-p"]`, mirroring the Claude CLI's prompt flag. But Codex CLI uses the `exec` subcommand for non-interactive runs. This caused every pr-reviewer run to fail with:

```
error: a value is required for '--profile <CONFIG_PROFILE>' but none was supplied
```

Codex interprets `-p` as `--profile`, which requires a value.

## Fix

Update the codex backend args in config examples and documentation:

```yaml
# Wrong
codex:
  args:
    - "-p"

# Correct
codex:
  args:
    - "exec"
```

The production config on ts-lonestar was already fixed on 2026-04-11.

## Complexity: S

eloylp:
Superseded by #64 which covers both the -p→exec fix and the --skip-git-repo-check requirement.

PR #65, opened 2026-04-11
config: fix codex backend args to use exec and --skip-git-repo-check

### LINKED ISSUE #64: config: codex backend requires exec and --skip-git-repo-check args
## Problem

The codex AI backend example config has been wrong in two ways, both discovered during the first day of autonomous agent deployment:

1. **`-p` is not the prompt flag.** Codex CLI uses `exec` subcommand for non-interactive runs. `-p` is interpreted as `--profile` which requires a value, causing `exit status 2`.

2. **`--skip-git-repo-check` is required.** When running inside a container (or any directory that isn't a git repo), codex refuses to execute with: `Not inside a trusted directory and --skip-git-repo-check was not specified.`

Both were fixed in the production config on ts-lonestar on 2026-04-11.

## Fix

Update the codex backend args in all config examples and documentation:

```yaml
# Wrong
codex:
  mode: command
  command: codex
  args:
    - "-p"

# Correct
codex:
  mode: command
  command: codex
  args:
    - "exec"
    - "--skip-git-repo-check"
```

This supersedes #61 which only caught the first issue.

## Complexity: S

eloylp:
## Problem

The codex AI backend example config had two bugs discovered during autonomous agent deployment:

1. **`-p` is wrong**: Codex CLI uses `exec` for non-interactive runs. `-p` is interpreted as `--profile` (requires a value), causing `exit status 2`.
2. **`--skip-git-repo-check` is required**: Inside containers (non-git directories), Codex refuses to run without this flag.

## Fix

Updated `config.example.yaml` and `README.md` to use the correct args:

```yaml
codex:
  mode: command
  command: codex
  args:
    - "exec"
    - "--skip-git-repo-check"
```

Added regression tests (`TestCodexBackendArgsInConfig`, `TestCodexExampleConfigArgs`) to prevent recurrence.

Closes #64. Supersedes #61 (PR #63 addressed only the first issue).

## Test plan
- [x] `go test ./...` passes
- [x] `config.example.yaml` updated
- [x] `README.md` updated
- [x] Regression tests added for both correct and incorrect arg patterns

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ ai_backends:
   codex:
     mode: command
     command: codex
-    args: ["exec", "--skip-git-repo-check"]
+    args: ["exec", "--skip-git-repo-check", "--dangerously-bypass-approvals-and-sandbox"]
     timeout_seconds: 600
     max_prompt_chars: 12000
     redaction_salt_env: LOG_SALT
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ ai_backends:
   codex:
     mode: command
     command: codex
-    args: ["-p"]
+    args: ["exec", "--skip-git-repo-check"]
     timeout_seconds: 600
     max_prompt_chars: 12000
     redaction_salt_env: LOG_SALT
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -435,6 +435,105 @@ repos:
 	}
 }
 
+func TestCodexBackendArgsInConfig(t *testing.T) {
+	t.Setenv("WEBHOOK_SECRET", "secret")
+
+	tests := []struct {
+		name    string
+		args    []string
+		wantErr bool
+	}{
+		{
+			name:    "correct args exec and skip-git-repo-check",
+			args:    []string{"exec", "--skip-git-repo-check"},
+			wantErr: false,
+		},
+		{
+			name:    "wrong args -p only",
+			args:    []string{"-p"},
+			wantErr: false, // config load doesn't validate args content, just that config parses
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `config.example.yaml:137` — issue #64's follow-up comment documents a third required Codex flag, `--dangerously-bypass-approvals-and-sandbox`. With only `exec` and `--skip-git-repo-check`, the copied config still cannot execute shell/MCP work in the default non-interactive sandbox, so this does not fully fix the production failure the PR references.
- **[medium]** `internal/config/config_test.go:499` — the new regression test hardcodes a synthetic YAML blob instead of loading `config.example.yaml` or checking the README snippet. That means future drift in the real example/docs will still pass CI.

### Summary
This moves the example in the right direction, but it still leaves the documented Codex invocation incomplete and the regression coverage aimed at the wrong artifact.

---

eloylp:
Addressed review feedback:
1. Added `--dangerously-bypass-approvals-and-sandbox` to the codex args in config.example.yaml (resolved the merge conflict with main which had already added it, so the flag is now present in all three positions: exec, --skip-git-repo-check, --dangerously-bypass-approvals-and-sandbox)
2. Fixed the regression test to load config.example.yaml directly instead of using a synthetic YAML blob, so future drift in the example file will be caught by CI

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The Codex docs/examples now line up with the runtime requirements, and the regression check now loads the real example config so future drift will be caught.

---

PR #73, opened 2026-04-12
autonomous: add SkipIfStillRunning guard to prevent goroutine accumulation

### LINKED ISSUE #45: autonomous scheduler: missing SkipIfStillRunning guard allows goroutine accumulation for slow agents
## Problem

`NewScheduler` creates the cron runner without a `SkipIfStillRunning` or `DelayIfStillRunning` wrapper:

```go
// autonomous/scheduler.go:27-28
parser := cron.NewParser(cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
c := cron.New(cron.WithParser(parser))
```

`robfig/cron/v3` launches each job in a new goroutine unconditionally. If an agent run takes longer than its cron interval (e.g., a 5-minute interval with a 10-minute agent run), the library fires a second goroutine at the 5-minute mark while the first is still executing.

The `MemoryStore.WithLock` call inside `runAgent` serialises actual task execution (the second goroutine blocks on `lock.Lock()` until the first finishes), so there is no data corruption. However:

1. **Goroutine accumulation**: Each missed tick adds a blocked goroutine that holds its cron entry open. Over time this can exhaust goroutine stack memory.
2. **Silent delay compounding**: Blocked goroutines execute back-to-back once the lock is released, not according to the configured schedule. A 5-minute agent running for 10 minutes will fire at t=0, t=10, t=10 (immediate), t=20, t=20 (immediate), … — running twice as often as intended with no indication.
3. **No observability**: The cron library logs nothing; the scheduler logs only the `next_run` time at registration, not actual execution delays.

## Fix

Add `cron.WithChain(cron.SkipIfStillRunning(logger))` when constructing the cron runner:

```go
// autonomous/scheduler.go:27-29
c := cron.New(
    cron.WithParser(parser),
    cron.WithChain(cron.SkipIfStillRunning(cron.DefaultLogger)),
)
```

This causes the library to skip a tick entirely if the previous run is still active, trading missed ticks for bounded resource use. Log a warning on skip so operators know when agents are overrunning their interval.

## Complexity: S

eloylp:
## Summary

- Adds `cron.WithChain(cron.SkipIfStillRunning(...))` to the cron runner in `NewScheduler` so that overlapping ticks for a slow agent are dropped rather than queued as unbounded goroutines.
- Introduces `zerologCronLogger` — a thin adapter implementing `cron.Logger` — so skip events are emitted through the structured zerolog pipeline instead of stdlib log.
- Adds `TestSchedulerSkipsJobWhenPreviousRunStillRunning` which uses a blocking runner to verify that a second concurrent invocation of the same cron job is skipped and only one runner call is made.

## Test plan
- [x] `go test ./...` passes
- [x] New test `TestSchedulerSkipsJobWhenPreviousRunStillRunning` confirms skip behaviour end-to-end via the cron entry's `WrappedJob`

Closes #45

### DIFF
--- a/internal/autonomous/scheduler.go
+++ b/internal/autonomous/scheduler.go
@@ -14,6 +14,28 @@ import (
 	"github.com/eloylp/agents/internal/config"
 )
 
+// zerologCronLogger adapts zerolog.Logger to the cron.Logger interface required
+// by chain wrappers such as SkipIfStillRunning.
+type zerologCronLogger struct {
+	logger zerolog.Logger
+}
+
+func (z zerologCronLogger) Info(msg string, keysAndValues ...interface{}) {
+	e := z.logger.Info()
+	for i := 0; i+1 < len(keysAndValues); i += 2 {
+		e = e.Interface(fmt.Sprintf("%v", keysAndValues[i]), keysAndValues[i+1])
+	}
+	e.Msg(msg)
+}
+
+func (z zerologCronLogger) Error(err error, msg string, keysAndValues ...interface{}) {
+	e := z.logger.Error().Err(err)
+	for i := 0; i+1 < len(keysAndValues); i += 2 {
+		e = e.Interface(fmt.Sprintf("%v", keysAndValues[i]), keysAndValues[i+1])
+	}
+	e.Msg(msg)
+}
+
 // AgentStatus is the runtime state of a single registered autonomous agent.
 type AgentStatus struct {
 	Name       string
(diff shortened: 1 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
This fixes the goroutine-accumulation failure mode directly, keeps the skip signal observable through the existing logger, and adds a focused regression test for the overlapping-run case.

---

Issue #39, opened 2026-04-08
AIBackendConfig.Mode not validated at config load time


eloylp:
## Problem

`AIBackendConfig.Mode` is never validated during config loading. The `normalizeBackends()` function in `internal/config/config.go` sets a default of `"noop"` for an empty mode (line 233) but does not reject unknown values. `validateBackends()` only checks that the backend name is one of `claude` / `codex` (line 304).

If an operator sets `mode: "command"` for a real backend (correct) or accidentally sets `mode: "cmd"` (a typo), both pass validation and produce a runner that is indistinguishable until the first webhook arrives. At that point `CommandRunner.Run` returns:

```
unknown claude mode: cmd
```

and the event is silently swallowed by the processor's error log.

**Affected file:** `internal/config/config.go` — `validateBackends()`

## Proposed fix

Add a mode check inside the existing backend validation loop:

```diff
 func (c *Config) validateBackends() error {
 	if len(c.AIBackends) == 0 {
 		return errors.New("config: at least one ai_backends entry is required")
 	}
 	for name, backend := range c.AIBackends {
 		if name != "claude" && name != "codex" {
 			return fmt.Errorf("config: unsupported ai backend %q (supported: claude, codex)", name)
 		}
+		if backend.Mode != "noop" && backend.Mode != "command" {
+			return fmt.Errorf("config: backend %q has unsupported mode %q (supported: noop, command)", name, backend.Mode)
+		}
 	}
 	return nil
 }
```

This is safe to apply unconditionally: `normalizeBackends()` already defaults empty mode to `"noop"` (line 233), so the check only fires for explicitly wrong values.

## Classification

Small / high-confidence — one-line addition to an existing validation loop, no behaviour change for correct configs.

*Filed by codebase-scout (2026-04-08)*

PR #74, opened 2026-04-12
config: validate mode=command requires non-empty command at startup

### LINKED ISSUE #70: config: mode=command with empty command field not validated at startup, error deferred to first AI invocation
## Summary

Issue #39 tracks that `AIBackendConfig.Mode` is not validated at config load time. Even after that is fixed, there is a related gap: when `mode: command` is set but the `command` field is empty, the error is only surfaced at the first actual AI invocation — not at startup.

## Current behaviour

In `CommandRunner.Run` (`internal/ai/cmdrunner.go`):
```go
case "command":
    if r.command == "" {
        return Response{}, fmt.Errorf("%s command is required when mode=command", r.backendName)
    }
```

This check runs on the first `Run()` call, which happens when the first webhook event arrives — potentially minutes or hours after the daemon starts. In the cron case it could be a day later.

## Example misconfiguration

```yaml
ai_backends:
  claude:
    mode: command
    # command field omitted
    args: ["-p", "--dangerously-skip-permissions"]
```

The daemon starts cleanly, logs no warnings, but silently fails every AI invocation with `"claude command is required when mode=command"`.

## Suggested fix

Add a check in `validateBackends()` in `internal/config/config.go`:

```go
for name, backend := range c.AIBackends {
    if backend.Mode == "command" && strings.TrimSpace(backend.Command) == "" {
        return fmt.Errorf("config: ai backend %q has mode=command but no command specified", name)
    }
}
```

This is a one-liner and high-confidence. It makes the full backend configuration visible at startup, consistent with how the daemon validates secrets and repos.

## Relationship to #39

#39 covers `Mode` not being validated against allowed values. This covers the `command` field being absent when `mode=command`. They are distinct checks and both belong in `validateBackends`.

eloylp:
## Summary

- Adds a check in `validateBackends()` that rejects configs where `mode: command` is set but the `command` field is empty
- Error is surfaced at daemon startup rather than at the first AI invocation (which could be hours later for cron agents)
- Two tests added: one for the rejection case, one confirming valid `mode=command` + non-empty `command` still loads cleanly

## Related

- Closes #70
- Complements #39 (mode value validation) — this handles the `command` field absence for `mode=command`

## Test plan

- [ ] `go test ./internal/config/... -count=1` passes
- [ ] Config with `mode: command` and empty `command` fails to load with a clear error
- [ ] Config with `mode: command` and `command: claude` loads cleanly

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -312,9 +312,12 @@ func (c *Config) validateBackends() error {
 	if len(c.AIBackends) == 0 {
 		return errors.New("config: at least one ai_backends entry is required")
 	}
-	for name := range c.AIBackends {
+	for name, backend := range c.AIBackends {
 		if name != "claude" && name != "codex" {
 			return fmt.Errorf("config: unsupported ai backend %q (supported: claude, codex)", name)
+		}
+		if backend.Mode == "command" && strings.TrimSpace(backend.Command) == "" {
+			return fmt.Errorf("config: ai backend %q has mode=command but no command specified", name)
 		}
 	}
 	return nil
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -239,6 +239,7 @@ func (c *Config) normalizeBackends() {
 		setDefault(&backend.Mode, "noop")
 		setDefaultInt(&backend.TimeoutSeconds, defaultAITimeoutSeconds)
 		setDefaultInt(&backend.MaxPromptChars, defaultMaxPromptChars)
+		backend.Command = strings.TrimSpace(backend.Command)
 		normalized[key] = backend
 	}
 	c.AIBackends = normalized
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -518,3 +518,54 @@ repos:
 		t.Fatalf("unexpected error for valid command mode config: %v", err)
 	}
 }
+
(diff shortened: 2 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config.go:319` — `validateBackends()` trims `backend.Command` only to decide whether it is empty, but `normalizeBackends()` never stores the trimmed value. A config like `command: "  claude  "` now passes startup validation and still fails later when `exec.CommandContext` receives the spaced command string. Trim and persist `backend.Command` during normalization (and add a regression test for whitespace-only / padded command values) so startup validation and runtime execution agree.

### Summary
Startup validation moved to the right place, but whitespace-padded `command` values can still slip through and fail at runtime.

---

eloylp:
Addressed the medium finding: added `backend.Command = strings.TrimSpace(backend.Command)` in `normalizeBackends()` so the trimmed value is persisted rather than only checked in-place. Added two regression tests: one that rejects whitespace-only commands and one that verifies whitespace-padded commands are normalized to the trimmed value in the loaded config.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The whitespace-normalization follow-up closes the earlier runtime gap; startup validation and the new regression tests now line up with actual command execution behaviour.

---

PR #75, opened 2026-04-12
config: validate AIBackendConfig.Mode at startup

eloylp:
## Summary

- Adds a mode check in `validateBackends()` that rejects any `mode` value other than `"noop"` or `"command"`
- Previously, typos like `mode: cmd` passed startup validation and produced a silent runtime error on the first webhook event (`unknown claude mode: cmd`)
- `normalizeBackends()` already defaults an empty mode to `"noop"`, so the new check only fires for explicitly wrong values — no behaviour change for correct configs

## Test plan

- [x] `go test ./...` passes
- [x] Table-driven rejection tests: `cmd` (typo), `subprocess` (unknown), `nOop` (wrong case) all fail at load
- [x] Acceptance tests: `noop` and `command` still load cleanly

Closes #39

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #76, opened 2026-04-12
config: reject startup when all repos have enabled: false

### LINKED ISSUE #52: config: validateRepos accepts a config where all repos are disabled, daemon starts but handles no events
## Problem

`validateRepos` only checks that at least one `repos` entry is present:

```go
// config.go:311
func (c *Config) validateRepos() error {
    if len(c.Repos) == 0 {
        return errors.New("config: at least one repo is required")
    }
    for _, repo := range c.Repos {
        if repo.FullName == "" {
            return errors.New("config: repo full_name is required")
        }
    }
    return nil
}
```

A config where every repo has `enabled: false` passes validation. The daemon starts, logs no warning, and silently drops every webhook (each `handleIssueEvent`/`handlePREvent` call hits the `!repo.Enabled` branch and returns 202 Accepted). The autonomous scheduler similarly skips all jobs during `registerJobs`. The operator sees a running process with a healthy `/status` and no indication anything is wrong.

## Fix (small, high-confidence)

Add an enabled-count check in `validateRepos`:

```diff
func (c *Config) validateRepos() error {
    if len(c.Repos) == 0 {
        return errors.New("config: at least one repo is required")
    }
+   enabledCount := 0
    for _, repo := range c.Repos {
        if repo.FullName == "" {
            return errors.New("config: repo full_name is required")
        }
+       if repo.Enabled {
+           enabledCount++
+       }
    }
+   if enabledCount == 0 {
+       return errors.New("config: at least one repo must have enabled: true")
+   }
    return nil
}
```

This makes the misconfiguration fail fast at startup rather than silently at runtime.

eloylp:
## Summary

- `validateRepos()` previously accepted configs where every repo had `enabled: false`, letting the daemon start and silently drop all webhook events.
- Adds an enabled-count check so the daemon fails fast at startup with `config: at least one repo must have enabled: true`.
- Updates all existing test repo entries to set `enabled: true` so each test continues to exercise its intended validation (not the new repos check).
- Adds `TestValidateReposAllDisabled` with table-driven cases covering all-disabled (rejects) and mixed-enabled (accepts).

Closes #52

## Test plan
- [ ] `go test ./... ` passes
- [ ] Config with all repos `enabled: false` fails at startup with a clear error
- [ ] Config with at least one `enabled: true` repo loads successfully

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"gopkg.in/yaml.v3"
@@ -368,10 +369,17 @@ func (c *Config) validateRepos() error {
 	if len(c.Repos) == 0 {
 		return errors.New("config: at least one repo is required")
 	}
+	enabledCount := 0
 	for _, repo := range c.Repos {
 		if repo.FullName == "" {
 			return errors.New("config: repo full_name is required")
 		}
+		if repo.Enabled {
+			enabledCount++
+		}
+	}
+	if enabledCount == 0 {
+		return errors.New("config: at least one repo must have enabled: true")
 	}
 	return nil
 }
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -19,6 +19,7 @@ ai_backends:
     mode: noop
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `
 	if err := os.WriteFile(path, []byte(content), 0o644); err != nil {
 		t.Fatalf("write config: %v", err)
@@ -46,6 +47,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 autonomous_agents:
   - repo: "owner/repo"
     enabled: true
@@ -183,6 +185,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `,
 		},
 		{
@@ -199,6 +202,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `,
 		},
 		{
@@ -217,6 +221,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `,
 		},
 	}
@@ -258,6 +263,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `,
 		},
 		{
@@ -274,6 +280,7 @@ agents:
   - name: architect
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `,
 		},
 		{
@@ -291,6 +298,7 @@ agents:
     skills: [nonexistent]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 `,
 		},
 		{
@@ -308,6 +316,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 autonomous_agents:
   - repo: "owner/repo"
     enabled: true
@@ -335,6 +344,7 @@ agents:
     skills: [architect]
 repos:
   - full_name: "owner/repo"
+    enabled: true
 autonomous_agents:
   - repo: "owner/repo"
     enabled: true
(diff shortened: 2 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- **[low]** `internal/config/config_test.go:21` — `TestLoadRequiresSupportedBackendNames` now contains the `enabled` key twice in the same YAML object. Today `yaml.v3` will usually just let the later value win, so the test may still pass, but the fixture is malformed and obscures what the test is actually exercising. Drop the duplicate key so this remains a clean unsupported-backend regression test.

### Summary
The startup validation change itself looks correct, but the first updated test fixture picked up a malformed duplicate YAML key.

---

eloylp:
Addressed review feedback:
1. Removed the duplicate `enabled: true` key from the `TestLoadRequiresSupportedBackendNames` fixture (it was accidentally doubled when adding `enabled: true` to all repo entries).
2. Rebased the branch onto main to resolve merge conflicts with the command-mode validation tests added by earlier merged PRs.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config_test.go:501` — after `validateRepos()` started rejecting configs with no enabled repos, the existing command-mode tests here and below (`TestLoadRejectsCommandModeWithEmptyCommand`, `TestLoadAcceptsCommandModeWithNonEmptyCommand`, `TestLoadRejectsCommandModeWithWhitespaceOnlyCommand`, `TestLoadNormalizesWhitespacePaddedCommand`) still build repo fixtures without `enabled: true`. That means the reject cases can now pass for the wrong reason, and the accept cases will fail for the wrong reason, so these tests no longer exercise command-mode validation at all.

### Summary
The startup repo validation itself looks correct, but this branch still leaves several pre-existing backend-mode tests misconfigured, so the suite no longer validates the behavior those tests are named after.

---

eloylp:
Addressed review feedback: added `enabled: true` to all four command-mode test fixtures (`TestLoadRejectsCommandModeWithEmptyCommand`, `TestLoadAcceptsCommandModeWithNonEmptyCommand`, `TestLoadRejectsCommandModeWithWhitespaceOnlyCommand`, `TestLoadNormalizesWhitespacePaddedCommand`). Each test now passes or fails for the correct reason — command-field validation — rather than being short-circuited by the all-repos-disabled check.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- **[low]** `internal/config/config_test.go:638` — `TestValidateReposAllDisabled` only distinguishes nil vs non-nil errors. This PR’s user-visible contract is the specific startup diagnostic added in `validateRepos()` (`config: at least one repo must have enabled: true`), so the regression test should assert that exact message, or at least a stable substring. As written, a future unrelated load failure would still satisfy the test and mask a regression in the error the operator actually sees.

### Summary
The validation logic looks correct, but the new regression test is still too loose to lock in the startup error this change is meant to guarantee.

---

eloylp:
Addressed review feedback: `TestValidateReposAllDisabled` now asserts the exact error message `"config: at least one repo must have enabled: true"` via a `wantErrMsg` field, so any future regression in the startup diagnostic will be caught immediately. Also rebased onto main (resolved conflict with the backend-mode tests from PR #75) and added `enabled: true` to the `TestLoadRejectsUnsupportedBackendMode` and `TestLoadAcceptsSupportedBackendModes` fixtures so those tests exercise mode validation rather than the all-repos-disabled check.

eloylp:
Addressed review feedback: converted `TestValidateReposAllDisabled` to a table-driven test with a `wantErrMsg` field and asserts `err.Error() == "config: at least one repo must have enabled: true"` for the reject case. A future unrelated load failure would produce a different message and fail the test, locking in the exact startup diagnostic this PR guarantees.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The startup validation is correct, the earlier test-fixture regressions were addressed, and the new table-driven coverage now pins the operator-facing error path this change introduces.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The startup validation matches issue #52, preserves the existing per-repo checks, and the updated tests now exercise their intended config paths with the new enabled-repo requirement in place.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The startup validation matches the issue intent, preserves the surrounding config checks, and the updated tests pin the new all-disabled reject path without misrouting existing validation coverage.

---

PR #82, opened 2026-04-12
workflow: replace errgroup with WaitGroup in HandlePullRequestLabelEvent

### LINKED ISSUE #50: workflow: vestigial errgroup.WithContext in HandlePullRequestLabelEvent; groupCtx never cancelled by goroutine errors
## Problem

`HandlePullRequestLabelEvent` uses `errgroup.WithContext`:

```go
// engine.go:112
group, groupCtx := errgroup.WithContext(ctx)
for _, ag := range agents {
    group.Go(func() error {
        // ...errors appended to agentErrs...
        return nil  // always nil
    })
}
_ = group.Wait()
return errors.Join(agentErrs...)
```

The key design detail: every goroutine always returns `nil`. `errgroup` derives a child context (`groupCtx`) that it cancels if any goroutine returns a non-nil error — but that path is never taken. `groupCtx` will only be cancelled if the parent `ctx` is cancelled, making it semantically equivalent to passing `ctx` directly to `runner.Run`.

The current code implies to a reader that an individual agent error will propagate context cancellation to sibling goroutines and abort them — it does not. This is misleading and the `errgroup` import adds indirection without benefit.

## Fix (small, high-confidence)

Replace `errgroup` with a plain `sync.WaitGroup` and pass the parent `ctx` directly:

```diff
-var (
-    mu        sync.Mutex
-    agentErrs []error
-)
-group, groupCtx := errgroup.WithContext(ctx)
-for _, ag := range agents {
-    group.Go(func() error {
-        ...
-        response, err := runner.Run(groupCtx, ...)
-        ...
-        return nil
-    })
-}
-_ = group.Wait()
-return errors.Join(agentErrs...)
+var (
+    wg        sync.WaitGroup
+    mu        sync.Mutex
+    agentErrs []error
+)
+for _, ag := range agents {
+    wg.Add(1)
+    go func() {
+        defer wg.Done()
+        ...
+        response, err := runner.Run(ctx, ...)
+        ...
+    }()
+}
+wg.Wait()
+return errors.Join(agentErrs...)
```

This removes the `golang.org/x/sync/errgroup` import from `engine.go` (the package is still used in `main.go`).

Note: this does not add a concurrency cap — that is tracked separately in #44.

eloylp:
## Summary

- `errgroup.WithContext` derived a `groupCtx` that was only cancelled on non-nil goroutine returns, but every goroutine unconditionally returned `nil`.
- `groupCtx` was therefore semantically equivalent to the parent `ctx`, making the `errgroup` indirection misleading without functional benefit.
- Replace with a plain `sync.WaitGroup` and pass `ctx` directly. Agent errors continue to be collected under a mutex and joined after all goroutines finish — one failing agent still does not abort the others.

## Test plan

- [ ] `TestHandlePullRequestLabelEventInvokesAllAgents` — verifies all agents are invoked (fan-out works)
- [ ] `TestHandlePullRequestLabelEventCollectsAllAgentErrors` — verifies errors from all agents are collected even when one fails (no short-circuit)
- [ ] `go test ./... -race` passes

Closes #50

### DIFF
--- a/internal/workflow/engine.go
+++ b/internal/workflow/engine.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 
 	"github.com/rs/zerolog"
-	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 
 	"github.com/eloylp/agents/internal/ai"
@@ -124,25 +123,27 @@ func (e *Engine) HandlePullRequestLabelEvent(ctx context.Context, req PRRequest)
 		Str("repo", req.Repo.FullName).
 		Int("pr_number", req.PR.Number).
 		Logger()
-	// Errors are collected into agentErrs rather than returned via errgroup so
-	// that a failing agent does not cancel the context and abort the others.
+	// Errors are collected into agentErrs rather than propagated via context
+	// cancellation so that a failing agent does not abort the remaining ones.
 	// All errors are joined and returned once every goroutine has finished.
 	var (
+		wg        sync.WaitGroup
 		mu        sync.Mutex
 		agentErrs []error
 	)
 	// sem limits the number of agent goroutines that run concurrently for this
 	// event. The cap is per-event (not global) so a large fan-out on one PR
-	// cannot starve another. Acquiring before group.Go avoids spawning goroutines
-	// that immediately block, keeping OS resource usage proportional to the limit.
+	// cannot starve another. Acquiring before spawning avoids goroutines that
+	// immediately block on the semaphore.
 	sem := semaphore.NewWeighted(int64(e.maxConcurrentAgents))
-	group, groupCtx := errgroup.WithContext(ctx)
 	for _, ag := range agents {
-		if err := sem.Acquire(groupCtx, 1); err != nil {
-			// groupCtx was cancelled; skip remaining agents.
+		if err := sem.Acquire(ctx, 1); err != nil {
+			// ctx was cancelled; skip remaining agents.
 			break
 		}
-		group.Go(func() error {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
 			defer sem.Release(1)
 			wf := fmt.Sprintf("%s:%s:%s", workflowPRReview, resolvedBackend, ag)
 			prompt, err := e.prompts.PRReviewPrompt(ag, resolvedBackend, req.Repo.FullName, req.PR.Number)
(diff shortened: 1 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The `WaitGroup` refactor matches the linked issue: it removes misleading `errgroup` machinery without changing the intended behavior that all agents run and their errors are joined after completion.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- **[low]** `internal/workflow/engine_test.go:163` — `TestHandlePullRequestLabelEventCollectsAllAgentErrors` says it verifies both agent-specific errors are present, but the loop only repeats `errors.Is(err, runner.err)`. That still passes if the joined result contains just one wrapped backend error, so the new regression does not actually pin the "collect all agent errors" contract described in the PR. Checking the joined error string for both `agent claude/architect:` and `agent claude/reviewer:` prefixes, or counting the joined members, would make the test match the intended behavior.

### Summary
The `WaitGroup` refactor itself looks correct, but one of the new tests does not yet prove the multi-error guarantee the PR is claiming.

---

eloylp:
Addressed the review finding: the test now checks both `agent claude/architect:` and `agent claude/reviewer:` substrings in the joined error string using `strings.Contains`, rather than relying on a vacuous `errors.Is` loop that passed regardless of whether both prefixes were present. The `errors.Is` check is kept as a separate assertion to verify the sentinel error is wrapped. Also rebased onto main and fixed a missing `path/filepath` import in `config.go` that was breaking compilation after PR #96.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/workflow/engine_test.go:109` — the new tests prove fan-out and joined-error behavior, but they replace the earlier regressions that exercised semaphore saturation and the zero-value `MaxConcurrentAgents` fallback. This refactor still changes the same concurrent path in `HandlePullRequestLabelEvent`, so dropping those targeted cases leaves unbounded fan-out or a zero-weight deadlock regression uncovered. Please restore coverage for the concurrency cap and the unset-config fallback alongside the new error-collection assertions.

### Summary
The `WaitGroup` refactor itself looks sound, but the branch weakens coverage around the handler’s concurrency guarantees.

---

eloylp:
Addressed the review findings: restored coverage for (1) the zero-value `MaxConcurrentAgents` fallback via `TestNewEngineDefaultsMaxConcurrentAgentsWhenZeroValue`, which asserts the engine falls back to `defaultMaxConcurrentAgents` when the config field is nil; and (2) semaphore saturation via `TestHandlePullRequestLabelEventRespectsMaxConcurrentAgents`, which spins up 4 agents with a cap of 2, verifies exactly 2 run concurrently at the saturation point, then confirms all 4 complete after the gate is released.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The `WaitGroup` refactor removes misleading `errgroup` wiring without changing the intended behavior that all agents run and their errors are joined after completion.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The `WaitGroup` refactor removes misleading `errgroup` wiring without changing fan-out semantics, and the restored tests cover both multi-error collection and the per-PR concurrency cap.

---

PR #83, opened 2026-04-12
cmd: fix absolute prompt_file paths silently broken by AgentsDir join

### LINKED ISSUE #48: resolvePrompts unconditionally joins AgentsDir prefix, silently breaking absolute prompt_file paths
## Problem

`resolvePrompts` in `cmd/agents/main.go` always prepends `cfg.AgentsDir` to `prompt_file`:

```go
// cmd/agents/main.go:144-151
resolve := func(src config.PromptSourceConfig) ai.PromptSource {
    if src.Prompt != "" {
        return ai.PromptSource{Prompt: src.Prompt}
    }
    return ai.PromptSource{PromptFile: filepath.Join(cfg.AgentsDir, src.PromptFile)}
}
```

`filepath.Join` on Linux does **not** treat a later absolute path as an override; it concatenates:

```
filepath.Join("/base/agents", "/etc/custom/PROMPT.md")
  → "/base/agents/etc/custom/PROMPT.md"   // not /etc/custom/PROMPT.md
```

An operator who sets an absolute path in `prompt_file` to point outside the agents directory gets a silently-wrong path that fails at prompt-load time with a misleading `load prompt …: open /base/agents/etc/custom/PROMPT.md: no such file or directory` error — not a config validation error.

The same pattern exists for skill `prompt_file` resolution in `resolveSkills` (line 115):

```go
sg.PromptFile = filepath.Join(cfg.AgentsDir, s.PromptFile)
```

## Fix

Check whether `PromptFile` is already absolute before joining:

```go
resolve := func(src config.PromptSourceConfig) ai.PromptSource {
    if src.Prompt != "" {
        return ai.PromptSource{Prompt: src.Prompt}
    }
    path := src.PromptFile
    if !filepath.IsAbs(path) {
        path = filepath.Join(cfg.AgentsDir, path)
    }
    return ai.PromptSource{PromptFile: path}
}
```

Apply the same fix in `resolveSkills`.

Alternatively, document in config that `prompt_file` must always be a path relative to `agents_dir` and add a validation error if an absolute path is supplied.

## Complexity: S (two-line fix)

eloylp:
## Summary

- `resolvePrompts` and `resolveSkills` in `cmd/agents/main.go` called `filepath.Join(agentsDir, promptFile)` unconditionally
- On Linux, `filepath.Join` does **not** treat a later absolute path as an override — it concatenates, so `/base/agents` + `/etc/custom/PROMPT.md` → `/base/agents/etc/custom/PROMPT.md` (wrong path, fails at load time with a misleading error)
- Fix: extract `agentsPath(agentsDir, promptFile)` which calls `filepath.IsAbs` and returns the path unchanged when already absolute; relative paths still use `filepath.Join`

Closes #48

## Test plan

- [ ] `go test ./...` passes (new table-driven tests in `cmd/agents/main_test.go` cover relative join, absolute passthrough, and the wrong-concatenation regression)
- [ ] Verify with `config.yaml` that `prompt_file: /absolute/path.md` resolves correctly

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -137,7 +137,7 @@ func resolveSkills(cfg *config.Config) []ai.SkillGuidance {
 	for _, s := range cfg.Skills {
 		sg := ai.SkillGuidance{Name: s.Name, Prompt: s.Prompt}
 		if s.PromptFile != "" {
-			sg.PromptFile = filepath.Join(cfg.AgentsDir, s.PromptFile)
+			sg.PromptFile = agentsPath(cfg.AgentsDir, s.PromptFile)
 		}
 		skills = append(skills, sg)
 	}
@@ -193,8 +193,18 @@ func resolvePrompts(cfg *config.Config) (issue ai.PromptSource, pr ai.PromptSour
 		if src.Prompt != "" {
 			return ai.PromptSource{Prompt: src.Prompt}
 		}
-		return ai.PromptSource{PromptFile: filepath.Join(cfg.AgentsDir, src.PromptFile)}
+		return ai.PromptSource{PromptFile: agentsPath(cfg.AgentsDir, src.PromptFile)}
 	}
 	return resolve(cfg.Prompts.IssueRefinement), resolve(cfg.Prompts.PRReview), resolve(cfg.Prompts.Autonomous)
+}
+
+// agentsPath resolves a prompt file path relative to agentsDir unless it is
+// already absolute, in which case it is returned unchanged. This mirrors the
+// POSIX convention that an absolute path always names an exact location.
+func agentsPath(agentsDir, promptFile string) string {
+	if filepath.IsAbs(promptFile) {
+		return promptFile
+	}
+	return filepath.Join(agentsDir, promptFile)
 }
 
--- a/cmd/agents/main_test.go
+++ b/cmd/agents/main_test.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"path/filepath"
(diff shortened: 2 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- **[low]** `cmd/agents/main_test.go:17` — `TestAgentsPath` hard-codes Unix-style absolute paths and expected strings (`/base/agents/...`, `/etc/...`). The production helper intentionally uses `filepath`, so these assertions will fail on Windows even when the implementation is correct there (`filepath.Join` emits `\\` separators and `filepath.IsAbs` follows Windows rules). Prefer building expectations with `filepath.Join` and, for the absolute-path case, using a platform-native absolute path form or gating the case on `runtime.GOOS`.

### Summary
The production fix looks correct, but the new regression test is not portable across Go's path semantics.

---

eloylp:
Addressed the portability concern: replaced hard-coded Unix absolute paths (`/base/agents`, `/etc/custom/PROMPT.md`) with `t.TempDir()` calls, which return real OS-native absolute paths. Relative-path expectations now use `filepath.Join` throughout. The test will build and pass correctly on Windows (and any other platform where `filepath.IsAbs` follows OS-specific rules).

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The absolute-path fix is correct in both prompt resolution sites, and the follow-up test update addressed the only portability issue from the earlier revision.

---

PR #85, opened 2026-04-12
config: use pointer fields so explicit zero backend values are honoured

### LINKED ISSUE #47: setDefaultInt/setDefaultInt64 treat 0 as unset, making explicit zero values indistinguishable from unconfigured
## Problem

All integer config defaults use `0` as the sentinel for "not configured":

```go
// internal/config/config.go:471-479
func setDefaultInt(field *int, value int) {
    if *field == 0 {
        *field = value
    }
}
func setDefaultInt64(field *int64, value int64) {
    if *field == 0 {
        *field = value
    }
}
```

This conflates two distinct states:
- **Not configured** (field omitted from YAML) → should apply default
- **Explicitly configured to 0** (field present, value `0`) → should honour the operator's intent

For several fields, `0` has useful semantics that cannot be expressed:

| Field | Effect of 0 | Current behaviour |
|---|---|---|
| `ai_backends.<n>.timeout_seconds` | No subprocess timeout (long-running tasks) | Gets overridden to 600 s |
| `http.max_body_bytes` | No body size limit | Gets overridden to 1 MiB |
| `http.max_concurrent_agents` (future) | Unlimited | Would get overridden |

`gopkg.in/yaml.v3` sets missing integer fields to their zero value, so there is no in-band signal for "omitted vs. zero".

## Fix

Use pointer fields for any integer that has a meaningful `0` value, or introduce a dedicated sentinel constant (e.g., `-1` for "no limit"). Pointer fields are the idiomatic Go solution and preserve YAML ergonomics (`omitempty`):

```go
type AIBackendConfig struct {
    TimeoutSeconds *int `yaml:"timeout_seconds"`
    ...
}
// applyDefaults:
if backend.TimeoutSeconds == nil {
    v := defaultAITimeoutSeconds
    backend.TimeoutSeconds = &v
}
```

Only fields where `0` carries meaning need changing; most (queue sizes, HTTP port timeouts) are fine as-is since zero is invalid there anyway.

## Complexity: S

Touches `AIBackendConfig.TimeoutSeconds` and `AIBackendConfig.MaxPromptChars` at minimum. Requires updating `normalizeBackends`, `validateBackends`, and `NewCommandRunner` call sites.

eloylp:
## Why

`AIBackendConfig.TimeoutSeconds` and `MaxPromptChars` were plain `int` fields. `setDefaultInt()` used `if *field == 0` as the sentinel for "not configured", making it impossible to distinguish an omitted field from one explicitly set to `0`.

This blocked two legitimate operator intents:
- `timeout_seconds: 0` → disable subprocess timeout for long-running AI tasks
- `max_prompt_chars: 0` → disable prompt truncation entirely

## What changed

- `AIBackendConfig.TimeoutSeconds` and `MaxPromptChars` changed from `int` to `*int`
- `normalizeBackends()` now only applies defaults when the pointer is `nil` (field omitted from YAML); explicit `0` passes through untouched
- `runCommand()` in `cmdrunner.go` skips `context.WithTimeout` when `timeout == 0`, inheriting the parent context instead (no subprocess deadline)
- `truncatePrompt` already treated `maxChars <= 0` as "no truncation" — no change needed there
- `main.go` dereferences the (now-always-non-nil after normalization) pointers when calling `NewCommandRunner`

## Tests

New table-driven test `TestExplicitZeroBackendFieldsHonoured` covers:
- explicit `timeout_seconds: 0` not overridden to default
- explicit `max_prompt_chars: 0` not overridden to default
- both zeros simultaneously

Closes #47

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -122,7 +122,7 @@ func setupPromptStore(cfg *config.Config, logger zerolog.Logger) (*ai.PromptStor
 func setupRunners(cfg *config.Config, logger zerolog.Logger) map[string]ai.Runner {
 	runners := make(map[string]ai.Runner, len(cfg.AIBackends))
 	for name, backend := range cfg.AIBackends {
-		runners[name] = ai.NewCommandRunner(name, backend.Mode, backend.Command, backend.Args, backend.TimeoutSeconds, backend.MaxPromptChars, backend.RedactionSaltEnv, logger)
+		runners[name] = ai.NewCommandRunner(name, backend.Mode, backend.Command, backend.Args, *backend.TimeoutSeconds, *backend.MaxPromptChars, backend.RedactionSaltEnv, logger)
 	}
 	return runners
 }
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -73,7 +73,13 @@ func (r *CommandRunner) Run(ctx context.Context, req Request) (Response, error) 
 }
 
 func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, req Request, prompt string) (Response, error) {
-	cmdCtx, cancel := context.WithTimeout(ctx, r.timeout)
+	var cmdCtx context.Context
+	var cancel context.CancelFunc
+	if r.timeout > 0 {
+		cmdCtx, cancel = context.WithTimeout(ctx, r.timeout)
+	} else {
+		cmdCtx, cancel = ctx, func() {}
+	}
 	defer cancel()
 
 	cmd := exec.CommandContext(cmdCtx, r.command, r.args...)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -129,8 +129,8 @@ type AIBackendConfig struct {
 	Mode             string   `yaml:"mode"`
 	Command          string   `yaml:"command"`
 	Args             []string `yaml:"args"`
-	TimeoutSeconds   int      `yaml:"timeout_seconds"`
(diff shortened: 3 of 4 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:78` — this widens the “no timeout” behavior from exactly `0` to every non-positive value. Before this change, a misconfigured negative `timeout_seconds` produced an already-expired context; now `-1`, `-30`, etc. inherit the parent context and run without any subprocess deadline because `validateBackends()` still accepts negative values (`internal/config/config.go:325`). The PR description and issue only justify `0` as the special case. Either reject negative values during config validation or gate the timeout bypass on `== 0` so invalid negatives do not silently become unlimited.

### Summary
The pointer-field change for explicit zero values makes sense, but the new timeout branch also changes negative timeout semantics in a way that should be fixed before merge.

---

eloylp:
Addressed review feedback: added validation in `validateBackends()` to reject negative `timeout_seconds` and `max_prompt_chars` values with descriptive startup errors. Negative values are now caught at config load time — the runtime `> 0` check in `cmdrunner.go` remains unchanged since it will only ever see valid non-negative values. Added table-driven test `TestLoadRejectsNegativeBackendFields` pinning the exact error messages for both fields.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The explicit-zero pointer change is now paired with startup validation for negative values, so the config semantics are correct and the added table-driven coverage exercises the new cases.

---

PR #86, opened 2026-04-12
config: remove dead PromptSourceConfig.Resolve method

### LINKED ISSUE #34: cleanup: PromptSourceConfig.Resolve() is dead code that duplicates ai.loadPromptSource logic
## Summary

`internal/config/config.go` exposes a `Resolve(baseDir string) (string, error)` method on `PromptSourceConfig`. This method reads and returns the raw prompt string from either the inline value or a file. However, it is **never called** by the daemon: `main.go` instead constructs `ai.PromptSource` values directly via `resolvePrompts()`, and the actual file reading is done later by `ai.loadPromptSource()` in the `ai` package.

This means the same "resolve a prompt from file-or-inline" logic lives in two places:

| Location | What it does |
|----------|--------------|
| `config.PromptSourceConfig.Resolve()` | reads file, returns string |
| `ai.loadPromptSource()` | reads file, parses as `text/template` |
| `main.go:resolvePrompts()` | constructs `ai.PromptSource` bridging the two |

The `config.Resolve()` method is likely a leftover from an earlier design iteration before `ai.PromptStore` was introduced.

## Suggested diff

Remove the dead method entirely:

```diff
-// Resolve returns the prompt content. If Prompt is set it is returned directly.
-// Otherwise the file at baseDir/PromptFile is read.
-func (p PromptSourceConfig) Resolve(baseDir string) (string, error) {
-    if p.Prompt != "" {
-        return p.Prompt, nil
-    }
-    path := p.PromptFile
-    if baseDir != "" {
-        path = filepath.Join(baseDir, p.PromptFile)
-    }
-    data, err := os.ReadFile(path)
-    if err != nil {
-        return "", fmt.Errorf("read prompt file %s: %w", path, err)
-    }
-    return string(data), nil
-}
```

If `Resolve()` is used in any test that currently passes, those tests should be rewritten to exercise `ai.loadPromptSource` instead, since that is the real code path.

Also consider inlining `main.go:resolvePrompts()` into `setupPromptStore()` to remove the indirection — the function only exists to bridge `config.PromptSourceConfig` to `ai.PromptSource`, which would become unnecessary if the two types were unified or if `ai.NewPromptStore` accepted `config.PromptSourceConfig` directly.

**Confidence**: high — this is purely dead-code removal with no behaviour change.

eloylp:
## Summary

`PromptSourceConfig.Resolve()` was dead code — never called anywhere in the codebase. Prompt loading goes through `ai.loadPromptSource()` instead, with `cmd/agents/main.go:resolvePrompts()` bridging `PromptSourceConfig` to `ai.PromptSource`.

Removing it:
- eliminates the duplicate file-or-inline resolution logic
- drops the now-unused `path/filepath` import
- no behaviour change

Closes #34

## Test plan

- [x] `go test ./... -race` passes (all existing tests green; no test exercised the dead method)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"gopkg.in/yaml.v3"
@@ -60,22 +59,6 @@ type PromptSourceConfig struct {
 	Prompt     string `yaml:"prompt"`
 }
 
-// Resolve returns the prompt content. If Prompt is set it is returned directly.
-// Otherwise the file at baseDir/PromptFile is read.
-func (p PromptSourceConfig) Resolve(baseDir string) (string, error) {
-	if p.Prompt != "" {
-		return p.Prompt, nil
-	}
-	path := p.PromptFile
-	if baseDir != "" {
-		path = filepath.Join(baseDir, p.PromptFile)
-	}
-	data, err := os.ReadFile(path)
-	if err != nil {
-		return "", fmt.Errorf("read prompt file %s: %w", path, err)
-	}
-	return string(data), nil
-}
 
 type SkillConfig struct {
 	Name       string `yaml:"name"`


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
This is a straightforward dead-code removal: `PromptSourceConfig.Resolve` has no remaining callers, so deleting it reduces duplicate prompt-loading logic without changing runtime behaviour.

---

PR #88, opened 2026-04-12
workflow: add test coverage for HandlePullRequestLabelEvent

### LINKED ISSUE #29: workflow: HandlePullRequestLabelEvent has no test coverage despite concurrent error-collection logic
## Problem

`Engine.HandlePullRequestLabelEvent` in `internal/workflow/engine.go` has **zero tests** in `engine_test.go`, while `HandleIssueLabelEvent` has one. This is the structurally more complex path:

1. **Draft-PR guard** — early return for draft PRs (silent skip, easy to regress).
2. **Conditional agent fan-out** — `"all"` token expands to every configured agent via `cfg.AgentNames()`; a named token resolves via `cfg.AgentByName`.
3. **Concurrent dispatch** — goroutines launched by `errgroup.WithContext`; each goroutine calls the runner independently.
4. **Manual error aggregation** — a `sync.Mutex` guards writes to a shared `agentErrs` slice; the results are joined via `errors.Join` after `group.Wait()`.

The concurrent error-aggregation pattern is subtle. Goroutines return `nil` to the errgroup and collect their own errors manually — a correctness regression (missed lock, wrong slice append, error swallowed before join) would be invisible to CI.

## What should be tested

- Draft PR → no runner invocation.
- Label with unknown backend or agent → skip, no error.
- `ai:review:<backend>:<agent>` → exactly one runner invocation with the correct workflow string.
- `ai:review:<backend>` ("all" agents) → one runner invocation per configured agent, all concurrent.
- One agent failing → other agents still run; error is returned.
- All agents failing → all errors are joined in the return value.

## Scope

`internal/workflow/engine_test.go` — new test function(s) for `HandlePullRequestLabelEvent`. The existing `testutil.BuildPromptStore` helper and `stubRunner` stub are sufficient; no new infrastructure required.

eloylp:
## Summary

- Adds a table-driven `TestHandlePullRequestLabelEvent` covering all meaningful paths: draft-PR skip, unrecognised-label skip, unknown-backend skip, unknown-agent skip, single-agent dispatch, all-agent fan-out, partial-failure (other agents still run), and all-failure (all errors joined).
- Adds a `sync.Mutex` to `stubRunner` so concurrent fan-out tests are safe under the race detector, and replaces the separate `calls int` / `last ai.Request` fields with a `calls []ai.Request` slice for richer assertions.

Closes #29

## Test plan
- [ ] `go test ./internal/workflow/... -race` — all 8 sub-tests pass
- [ ] `go test ./... -race` — full suite green

### DIFF
--- a/internal/workflow/engine_test.go
+++ b/internal/workflow/engine_test.go
@@ -2,6 +2,8 @@ package workflow
 
 import (
 	"context"
+	"errors"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/workflow/engine_test.go:201` — The workflow assertion only checks that each recorded workflow is in an allowed set. If the fan-out regressed to dispatch `architect` twice and never dispatched `scout`, this test would still pass because `wantCalls` is `2` and both workflows are still "allowed". Compare the exact workflow multiset or per-agent counts so the test proves one invocation per configured agent.
- **[medium]** `internal/workflow/engine_test.go:186` — The `all-agents-fail-errors-joined` case only checks for one `"runner down"` substring, so it does not prove that both agent failures survive `errors.Join`. A regression that returns only one agent error would still satisfy this assertion. Assert both agent-specific wrappers are present, or otherwise verify that all failures were joined.

### Summary
This adds useful coverage, but two key assertions are still too weak to catch the exact fan-out and error-aggregation regressions the issue calls out.

---

eloylp:

1. **[medium] Fan-out workflow check too weak** — Replaced the set-membership check with an exact multiset comparison using count maps. The test now fails if any expected workflow is dispatched zero or more-than-once, or if an unexpected workflow appears. A regression that dispatches `architect` twice and never dispatches `scout` will now be caught.

2. **[medium] Error-join only checks one failure** — Changed `all-agents-fail-errors-joined` to return distinct per-agent errors (`"architect runner down"` / `"scout runner down"`) and assert both appear in the joined error via the new `wantAllErrContains` field. A regression that drops one agent's error will now fail the test.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The follow-up commit closes the earlier gaps: fan-out is now asserted as an exact multiset and the joined-error path proves both agent failures survive aggregation.

---

PR #90, opened 2026-04-12
webhook: replace per-request O(n) eviction with background ticker in DeliveryStore

### LINKED ISSUE #46: DeliveryStore.SeenOrAdd performs O(n) eviction scan on every webhook request
## Problem

`DeliveryStore.SeenOrAdd` (`internal/webhook/delivery_store.go:24-39`) iterates **all stored deliveries** on every call to evict expired entries:

```go
for key, expiresAt := range s.deliveries {
    if now.After(expiresAt) {
        delete(s.deliveries, key)
    }
}
```

With the default TTL of 3600 s and sustained webhook traffic, the map grows to O(requests/hour) entries. Each subsequent call then pays O(n) to scan the whole map even when most entries are live. The mutex is held for the full iteration, blocking all concurrent webhook processing.

## When does this matter?

Webhook traffic is typically low, but GitHub can replay deliveries or burst during CI events. At 100 req/s over an hour the map contains ~360 000 entries; each new webhook must scan all of them.

## Fix options

**Option A — Background eviction goroutine (simplest):** Run a ticker goroutine that evicts expired entries periodically (e.g., every `ttl/4`). Removes per-request O(n) cost entirely.

**Option B — Ordered eviction with a heap or ring buffer:** Track insertion order; pop from the front when evicting. O(log n) insert, O(1) amortised evict.

**Option A** is the smallest change and sufficient for expected traffic levels.

```go
// sketch: start in NewDeliveryStore
go func() {
    ticker := time.NewTicker(s.ttl / 4)
    defer ticker.Stop()
    for range ticker.C {
        s.evict(time.Now())
    }
}()
```

The goroutine must be stopped on shutdown (pass a context or stop channel to `NewDeliveryStore`).

## Complexity: S

eloylp:
## Summary

- `SeenOrAdd` previously scanned all entries on every call to evict expired deliveries, holding the mutex for an O(n) scan that grows with hourly traffic volume.
- Replaced with a background goroutine started via `Start(ctx context.Context)` that ticks at `ttl/4` and evicts expired entries independently of request handling.
- `SeenOrAdd` is now a simple O(1) lookup + insert with no eviction loop.
- `deliveryStore.Start(groupCtx)` is called in `main.go` so the goroutine stops cleanly when the daemon shuts down.
- Added `delivery_store_test.go` with table-driven tests for `SeenOrAdd`, `Delete`, background eviction, and context-cancel stop behaviour.

Closes #46

## Test plan

- [ ] `go test ./internal/webhook/... -v -run TestDeliveryStore` — all 4 test functions pass
- [ ] `go test ./...` — full suite green

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -85,6 +85,7 @@ func run() error {
 	server := webhook.NewServer(cfg, deliveryStore, dataChannels, schedulerStatusAdapter{scheduler}, logger, scheduler)
 
 	group, groupCtx := errgroup.WithContext(ctx)
+	deliveryStore.Start(groupCtx)
 	group.Go(func() error { return processor.Run(groupCtx) })
 	group.Go(func() error { return scheduler.Run(groupCtx) })
 	group.Go(func() error { return server.Run(groupCtx) })
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -381,6 +381,9 @@ func (c *Config) validate() error {
 	if c.Daemon.HTTP.WebhookSecret == "" {
 		return errors.New("config: http webhook secret is required (set webhook_secret_env)")
 	}
+	if c.Daemon.HTTP.DeliveryTTLSeconds < 0 {
+		return fmt.Errorf("config: http delivery_ttl_seconds must be positive, got %d", c.Daemon.HTTP.DeliveryTTLSeconds)
+	}
 	if err := c.validateBackends(); err != nil {
 		return err
 	}
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -331,3 +331,36 @@ func TestConfigExampleYAMLLoads(t *testing.T) {
 		t.Fatal("example has no skills")
 	}
 }
+
+func TestLoadRejectsNegativeDeliveryTTL(t *testing.T) {
+	t.Setenv("TEST_SECRET", "secret")
+
+	content := `
+daemon:
+  http:
+    webhook_secret_env: TEST_SECRET
+    delivery_ttl_seconds: -1
+  ai_backends:
+    claude:
+      command: claude
+      args: ["-p"]
+skills:
+  architect:
+    prompt: "Focus on architecture."
+agents:
+  - name: reviewer
+    backend: claude
+    skills: [architect]
+    prompt: "You review PRs."
+repos:
+  - name: "owner/repo"
+    enabled: true
+    use:
+      - agent: reviewer
+        labels: ["ai:review:reviewer"]
+`
+	path := writeConfig(t, content)
+	if _, err := Load(path); err == nil {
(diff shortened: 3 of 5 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/delivery_store.go:26` — `time.NewTicker(s.ttl / 4)` panics when the interval is non-positive. `delivery_ttl_seconds` is not validated in config today, so a negative TTL now crashes the daemon at startup instead of merely making entries expire immediately. Guard `Start` against invalid TTLs or reject them during config validation before constructing the store.

### Summary
Moving eviction off the request path is the right simplification, but the new ticker introduces a startup panic on still-accepted invalid TTL values.

---

eloylp:

1. **[high] Ticker panic on non-positive TTL** — `DeliveryStore.Start` now returns early when `s.ttl <= 0`, preventing the `time.NewTicker` panic entirely. Added `TestDeliveryStore_StartIsNoOpForNonPositiveTTL` to cover both zero and negative durations.

2. **Config validation** — `validate()` now rejects negative `delivery_ttl_seconds` with a clear error. Zero is intentionally not rejected because `setDefaultInt` replaces zero with the default (3600) before validation runs, so it's impossible to reach validation with a zero value from a zero-valued YAML field. Added `TestLoadRejectsNegativeDeliveryTTL` to confirm the config-level guard.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The non-positive TTL guard, config validation, and added tests address the ticker panic cleanly while keeping the request path simplification intact.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The non-positive-TTL guard, config validation, and new delivery-store tests address the ticker panic cleanly while keeping request-path dedupe O(1).

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The background-eviction refactor keeps request-path dedupe O(1), and the non-positive-TTL guard plus validation close the startup-panic hole from the earlier revision.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The background-eviction refactor removes the per-request O(n) scan, and the added non-positive-TTL guard plus config validation close the startup-panic edge case introduced by the ticker.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The background-eviction refactor removes the O(n) request-path scan, and the non-positive-TTL guard plus config validation cover the startup-panic edge case from earlier revisions.

---

PR #94, opened 2026-04-12
workflow: cap concurrent agent goroutines in PR-review fan-out

### LINKED ISSUE #44: HandlePullRequestLabelEvent launches unbounded concurrent subprocesses when agent=all
## Problem

`Engine.HandlePullRequestLabelEvent` in `internal/workflow/engine.go` uses an errgroup to fan-out one goroutine per configured agent:

```go
// engine.go:112-142
group, groupCtx := errgroup.WithContext(ctx)
for _, ag := range agents {
    group.Go(func() error {
        // ... builds prompt then:
        response, err := runner.Run(groupCtx, ai.Request{...})
        ...
    })
}
_ = group.Wait()
```

Each `runner.Run` call in command mode (`mode: command`) forks a long-running subprocess (AI CLI) that may run for up to `timeout_seconds` (default 600 s). When the label is `ai:review` or `ai:review:<backend>` (i.e., `agent=all`), **all configured agents are dispatched simultaneously** with no concurrency cap.

With N agents this creates N concurrent subprocess invocations, each holding OS resources (file descriptors, RAM, CPU) for up to 10 minutes. On a busy repository that receives multiple PR review labels within a short window, this compounds: each incoming PR event dispatched by the single PR worker spawns another N subprocesses.

## Impact

- Unbounded subprocess concurrency → resource exhaustion on the host
- Interacts with issue #38 (single PR worker): if the PR worker is blocked waiting for a batch of N agents to finish, new PR events queue up and eventually fill the buffer (default 256), causing 503s back to GitHub

## Fix

Add a semaphore (`golang.org/x/sync/semaphore` or a buffered channel) in `Engine` (or passed from config) to limit the number of concurrently running agent goroutines. A reasonable default is `max(1, numCPU)` or a configurable `max_concurrent_agents` field.

```go
// sketch
sem := semaphore.NewWeighted(int64(e.cfg.MaxConcurrentAgents))
group.Go(func() error {
    sem.Acquire(groupCtx, 1)
    defer sem.Release(1)
    ...
})
```

The limit should be per-event, not global, to avoid cross-event interference.

## Complexity: S–M

Semaphore addition is small. Adding a config field and wiring it to `Engine` is the larger part.

eloylp:
## Summary

- Adds `processor.max_concurrent_agents` config field (default `4`, must be `>= 1`) to bound the number of agent goroutines that can run concurrently for a single PR-review fan-out event
- Uses `golang.org/x/sync/semaphore` (already in go.mod) to gate goroutine spawning; the semaphore is per-event so large fan-outs on one PR cannot starve another
- Validates the new field at startup: negative values are rejected (`config: processor.max_concurrent_agents must be >= 1`)
- Documents the field in `config.example.yaml`

## Root cause

`HandlePullRequestLabelEvent` with `agent=all` launches one goroutine per configured agent with no concurrency limit. Each goroutine forks a subprocess that can hold OS resources for up to `timeout_seconds` (default 600 s). With N agents and frequent `ai:review` labels, N × events subprocesses can accumulate simultaneously.

## Test plan

- [ ] `TestHandlePRLabelEventRespectsConcurrencyLimit` — forces all `Run` calls to block until released; asserts peak concurrency never exceeds the configured limit (2 out of 4 agents)
- [ ] `TestLoadAppliesDefaults` — asserts `max_concurrent_agents` defaults to `4`
- [ ] `TestLoadRejectsNegativeMaxConcurrentAgents` — asserts negative values are rejected at startup
- [ ] `go test ./... -race` passes

Closes #44

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -27,6 +27,11 @@ http:
 processor:
   issue_queue_buffer: 256
   pr_queue_buffer: 256
+  # Maximum number of agent goroutines that may run concurrently for a single
+  # PR-review fan-out event. Limits subprocess resource usage when many agents
+  # are configured and ai:review (agent=all) labels are used frequently.
+  # Default: 4. Must be >= 1.
+  max_concurrent_agents: 4
 
 # ──────────────────────────────────────────────
 # Paths
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -22,6 +22,7 @@ const (
 	defaultDeliveryTTLSeconds      = 3600
 	defaultIssueQueueBufferSize    = 256
 	defaultPRQueueBufferSize       = 256
+	defaultMaxConcurrentAgents     = 4
 	defaultHTTPShutdownSeconds     = 15
 	defaultAITimeoutSeconds        = 600
 	defaultMaxPromptChars          = 12000
@@ -134,8 +135,9 @@ type HTTPConfig struct {
 }
 
 type ProcessorConfig struct {
-	IssueQueueBuffer int `yaml:"issue_queue_buffer"`
-	PRQueueBuffer    int `yaml:"pr_queue_buffer"`
+	IssueQueueBuffer    int `yaml:"issue_queue_buffer"`
+	PRQueueBuffer       int `yaml:"pr_queue_buffer"`
+	MaxConcurrentAgents int `yaml:"max_concurrent_agents"`
 }
 
 type RepoConfig struct {
@@ -235,6 +237,7 @@ func (c *Config) applyHTTPDefaults() {
 func (c *Config) applyProcessorDefaults() {
 	setDefaultInt(&c.Processor.IssueQueueBuffer, defaultIssueQueueBufferSize)
 	setDefaultInt(&c.Processor.PRQueueBuffer, defaultPRQueueBufferSize)
+	setDefaultInt(&c.Processor.MaxConcurrentAgents, defaultMaxConcurrentAgents)
 }
 
 func (c *Config) normalizeSkills() {
@@ -316,6 +319,9 @@ func (c *Config) validate() error {
 	if c.HTTP.WebhookSecret == "" {
 		return errors.New("config: http webhook secret is required")
 	}
+	if err := c.validateProcessor(); err != nil {
+		return err
+	}
 	if err := c.validateBackends(); err != nil {
 		return err
 	}
@@ -333,6 +339,13 @@ func (c *Config) validate() error {
 		return err
 	}
 	return c.validateAutonomousAgents(skillNames)
+}
+
+func (c *Config) validateProcessor() error {
+	if c.Processor.MaxConcurrentAgents < 1 {
+		return fmt.Errorf("config: processor.max_concurrent_agents must be >= 1, got %d", c.Processor.MaxConcurrentAgents)
+	}
+	return nil
 }
 
 func (c *Config) validateBackends() error {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -135,9 +135,9 @@ type HTTPConfig struct {
 }
 
 type ProcessorConfig struct {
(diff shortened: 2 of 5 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/workflow/engine.go:117` — `HandlePullRequestLabelEvent` now assumes `cfg.Processor.MaxConcurrentAgents` was populated by `config.Load`. With a hand-built `config.Config` (which this package's tests already do, and `NewEngine` does not forbid), the zero value creates `semaphore.NewWeighted(0)`, so the first `Acquire(groupCtx, 1)` blocks forever until the context is cancelled. That turns a previously working direct-construction path into a deadlock. Please either default `<= 0` to the configured fallback in `NewEngine`/this handler, or add a defensive validation at the constructor boundary.

### Summary
The per-event semaphore is the right fix for #44, but it needs a zero-value guard so `NewEngine` remains safe with manually constructed configs.

---

eloylp:
Addressed the review finding:

- Added `maxConcurrentAgents int` field to `Engine` populated in `NewEngine`.
- If `cfg.Processor.MaxConcurrentAgents <= 0` (zero-value Config), it falls back to `defaultMaxConcurrentAgents = 4` — matching the value `config.applyDefaults` would have set.
- `HandlePullRequestLabelEvent` now reads from `e.maxConcurrentAgents` instead of `e.cfg.Processor.MaxConcurrentAgents`, so the semaphore always has a positive weight regardless of how the engine was constructed.
- Added `TestNewEngineZeroMaxConcurrentAgentsFallsBackToDefault` which exercises the zero-value path end-to-end and confirms no deadlock occurs.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config.go:222` — `applyProcessorDefaults` currently rewrites an explicit `processor.max_concurrent_agents: 0` to `4` before `validateProcessor` runs, so a user-supplied invalid value is silently accepted even though the field is documented as `>= 1`. That repeats the explicit-zero config bug this repo has already fixed in other places. Please preserve the distinction between “omitted” and “set to 0” and add a load test that `0` is rejected.
- **[low]** `internal/workflow/engine_test.go:123` — `TestHandlePRLabelEventRespectsConcurrencyLimit` sleeps for 50 ms and only asserts `peak <= limit`, so it can pass without ever proving the semaphore saturated. A more reliable test would synchronize on the first `limit` `Run` calls starting and assert `peak == limit` before releasing them.

### Summary
The semaphore approach is right, but the config path still masks an explicitly invalid `0` value and the main concurrency test is weaker than it needs to be.

---

eloylp:
Addressed review findings:
1. Changed `ProcessorConfig.MaxConcurrentAgents` from `int` to `*int` so an explicit YAML `0` is distinguishable from an omitted field. `applyProcessorDefaults` only fills in the default when the pointer is `nil`. `validateProcessor` now rejects both `0` and negative values. `TestLoadRejectsInvalidMaxConcurrentAgents` covers both cases.
2. Replaced the `time.Sleep`-based concurrency check with a buffered `started` channel on `concurrencyTrackingRunner`. The test now waits for exactly `limit` Run calls to be in-flight before asserting `peak == limit`, making the test deterministic.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Semaphore-based per-event fan-out limiting now covers the zero-value config path and has appropriate validation and regression tests.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The per-event semaphore fix now covers the zero-value constructor path, rejects explicit invalid config values, and the fan-out test is materially stronger than the earlier sleep-based version.

---

PR #95, opened 2026-04-12
ai: replace hand-rolled backward-scanner in extractJSON with json.Decoder

### LINKED ISSUE #32: arch: extractJSON backward-scanner has string-escape edge cases that can silently return wrong JSON
## Summary

`internal/ai/cmdrunner.go` contains a hand-rolled backward-scanning JSON extractor (`extractJSON`) that is intended to find the last top-level JSON object when AI CLIs emit conversational text before their JSON payload. The scanner has at least one correctness edge case that can cause it to return a truncated or wrong object silently.

## The problem

The backward scan tracks `inString` state with this logic (simplified):

```go
if inString {
    if b == '"' && (i == 0 || data[i-1] != '\\') {
        inString = false   // exit string
    }
    continue
}
```

**Edge case**: a string ending with `\\\"` (an escaped backslash followed by an escaped quote, e.g. the JSON value `"path\\"`) looks like:

```
... 5c 5c 22 ...
          ^-- scanner sees '"' preceded by '\', keeps inString=true  ← WRONG
```

The preceding `\` is itself escaped, so `\"` here _is_ a closing quote — but the scanner misses it. This can cause the scanner to walk past the true object boundary and either return a larger (syntactically invalid) slice or fail with "no matching `{` found".

In practice this fires whenever an AI CLI includes a Windows-style path or any `\\`-terminated string in its JSON output.

## Suggested fix

Replace the custom scanner with a forward-scanning approach using `encoding/json.Decoder`:

```go
func extractJSON(data []byte) ([]byte, error) {
    dec := json.NewDecoder(bytes.NewReader(data))
    var last json.RawMessage
    for {
        var raw json.RawMessage
        if err := dec.Decode(&raw); err != nil {
            break
        }
        if len(raw) > 0 && raw[0] == '{' {
            last = raw
        }
    }
    if last == nil {
        return nil, fmt.Errorf("no JSON object found in output")
    }
    return last, nil
}
```

This correctly handles all escape sequences by delegating to the stdlib parser, and still returns the _last_ top-level object (matching the current intent). It is also simpler and easier to maintain.

The change is small and self-contained to `internal/ai/cmdrunner.go`.

eloylp:
## Summary

- The old `extractJSON` backward scanner tracked `inString` state by checking if `data[i-1] != '\\'`. This fails for strings ending with `\\` (escaped backslash): the byte sequence is `5c 5c 22`, and the final `"` is misread as an escaped quote because the preceding byte is `\`. The scanner walks past the true object boundary and returns a truncated or syntactically invalid slice.
- In practice this fires whenever an AI CLI includes a Windows-style path (e.g. `C:\\Users\\foo\\`) or any `\\`-terminated string in its JSON output.

## Fix

Replace the hand-rolled backward scanner with a forward scan:

1. Walk through `data` byte by byte looking for `{` characters.
2. At each `{`, attempt `json.Decoder.Decode` into a `json.RawMessage`.
3. If decoding succeeds, store the result and advance past the decoded object.
4. Return the last successfully decoded object.

`json.Decoder` handles all string-escape sequences (including `\\`) correctly, eliminating the whole class of escape-related bugs. The "find the last JSON object" contract is preserved.

## Test plan

- `string_ending_with_escaped_backslash` — the exact bug: `{"path":"C:\\Users\\foo\\"}` now round-trips correctly
- `text_before_json_with_escaped_backslash_in_value` — preamble text + buggy value
- `multiple_top-level_objects_returns_last` — last-object contract preserved
- `top-level_array_containing_object_returns_the_inner_object` — pins documented contract (scanner finds `{` inside the array)
- `top-level_array_then_object_returns_object` — mixed output case
- All existing tests continue to pass
- `go test ./...` passes

Closes #32

### DIFF
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -185,47 +185,30 @@ func buildCommandEnv(req Request) []string {
 }
 
 // extractJSON finds the last top-level JSON object in data.
-// AI CLIs sometimes emit conversational text before the JSON payload;
-// this function scans backwards to locate the closing '}' and then
-// walks back to find its matching '{', accounting for nested braces
-// and JSON strings.
+// AI CLIs sometimes emit conversational text before the JSON payload.
+// This function scans forward through data looking for '{' characters and
+// attempts to decode a complete JSON object at each one using encoding/json.
+// The last successfully decoded object is returned.  Using the standard
+// decoder avoids the string-escape edge cases in a hand-rolled backward
+// scanner (e.g. a string ending with '\\' would be misread as an escaped
+// quote by a naive backward scan).
 func extractJSON(data []byte) ([]byte, error) {
-	// Find last '}'.
-	end := -1
-	for i := len(data) - 1; i >= 0; i-- {
-		if data[i] == '}' {
-			end = i
-			break
-		}
-	}
-	if end < 0 {
-		return nil, fmt.Errorf("no JSON object found in output")
-	}
-
-	// Walk backwards from end to find the matching '{'.
-	depth := 0
-	inString := false
-	for i := end; i >= 0; i-- {
-		b := data[i]
-		if inString {
-			if b == '"' && (i == 0 || data[i-1] != '\\') {
-				inString = false
-			}
+	var last json.RawMessage
+	for i := 0; i < len(data); i++ {
+		if data[i] != '{' {
 			continue
 		}
-		switch b {
-		case '"':
-			inString = true
-		case '}':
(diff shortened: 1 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The switch to `json.Decoder` removes the escape-handling bug cleanly, preserves the documented last-object behavior, and the added tests cover the important regression shapes from #32.

---

PR #97, opened 2026-04-12
refactor(ai): extract makeTestPromptSources helper in prompt_test

eloylp:
## Why

`TestBuildPRReviewPromptWithInlineSkill` and `TestBuildPRReviewPromptWithMultipleSkills` both opened with 8 lines of identical fixture setup — creating the same two directories and writing the same two template files. That duplication has two problems:

1. **Silently swallowed errors**: all `os.MkdirAll` / `os.WriteFile` calls used `_ =`, so a disk error would let the test continue with missing files and produce a confusing downstream failure rather than a clear "setup failed" message.
2. **Copy-paste surface**: any change to the fixture templates (e.g. renaming a field) had to be applied in two places.

## What changed

Extracted a `makeTestPromptSources(t *testing.T)` helper that:
- creates the on-disk issue and PR base fixture files,
- uses `t.Fatal` for each I/O step so failures are surfaced immediately,
- carries `t.Helper()` so test failure lines point to the test, not the helper.

The two affected tests now start with a single `issueBase, prBase, autoBase := makeTestPromptSources(t)` call. No behaviour change.

## Test

```
go test ./internal/ai/... 
```
All tests pass.

### DIFF
--- a/internal/ai/prompt_test.go
+++ b/internal/ai/prompt_test.go
@@ -42,25 +42,13 @@ func TestBuildPRReviewPromptIncludesRequirements(t *testing.T) {
 }
 
 func TestBuildPRReviewPromptWithInlineSkill(t *testing.T) {
-	dir := t.TempDir()
-	issueDir := filepath.Join(dir, "issue_refinement_prompts")
-	_ = os.MkdirAll(issueDir, 0o755)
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
This is a clean test-only refactor: it removes duplicated setup, stops swallowing fixture I/O errors, and keeps failures pointed at the caller via `t.Helper()`.

---

PR #98, opened 2026-04-12
workflow: add bounded worker pool to Processor for concurrent event handling

### LINKED ISSUE #38: Processor: single worker per event type serializes all events when an agent is slow
## Problem

`internal/workflow/processor.go` starts exactly **one goroutine** per event channel:

```go
wg.Add(2)
go p.runIssueWorker(ctx, &wg)   // one goroutine for all issue events
go p.runPRWorker(ctx, &wg)      // one goroutine for all PR events
```

Each worker is a sequential loop:

```go
for req := range p.channels.IssueChan() {
    _ = p.handler.HandleIssueLabelEvent(p.processingCtx(ctx), req)
}
```

`HandleIssueLabelEvent` and `HandlePullRequestLabelEvent` both invoke an external AI CLI process that can take seconds to minutes. During that time all subsequently queued events of the same type pile up behind it.

## Impact

- Under any non-trivial label rate, a slow AI response causes head-of-line blocking: every waiting event is delayed by the full execution time of the in-flight one.
- Separate repositories, separate issues, separate PRs — conceptually independent — are serialized through the same single worker.
- For `HandlePullRequestLabelEvent` the Engine already fans out across agents within a single PR using `errgroup`; that per-PR parallelism is wiped out when the worker is busy.

## Options

**Option A — bounded worker pool per channel** (moderate change)

Replace the single goroutine with a fixed-size pool (size configurable via `processor.workers`, defaulting to, say, 4):

```go
for range workerCount {
    wg.Add(1)
    go p.runIssueWorker(ctx, &wg)
}
```

Workers read from the same shared channel; no other changes needed. Adds natural back-pressure through channel capacity.

**Option B — per-repository worker** (larger change)

Shard the channel by repository full name so that two different repos never block each other. Requires changes to `DataChannels` and the routing logic in `webhook/server.go`.

**Option C — semaphore-limited concurrency** (minimal change)

Keep one goroutine but let it launch handler calls as goroutines bounded by a `semaphore.Weighted`:

```go
sem := semaphore.NewWeighted(int64(workerCount))
for req := range p.channels.IssueChan() {
    _ = sem.Acquire(ctx, 1)
    go func(r IssueRequest) {
        defer sem.Release(1)
        _ = p.handler.HandleIssueLabelEvent(p.processingCtx(ctx), r)
    }(req)
}
```

## Recommendation

Option A is the lowest-risk change with the highest benefit. It requires only adding a `Workers int` field to the processor config (already has `processor.shutdown_timeout`), defaulting to a small value. The existing drain logic (channel close + WaitGroup) works unchanged for any number of workers.

This is a larger / more impactful change than a quick fix, so filing as an issue rather than directly proposing a PR.

eloylp:
## Summary

Closes #38

Previously `Processor` started exactly **one goroutine per event type**, serializing every issue and every PR event through a single worker. A slow AI call blocked all subsequent events of that type, regardless of how independent the repos or PRs were.

- Add `processor.workers` config field (default `4`) controlling how many goroutines read from each event channel.
- Spawn `workers` goroutines for both the issue channel and the PR channel in `Processor.Run`, giving `2×workers` total goroutines.
- Workers share the same channel; the existing buffered-channel back-pressure and drain-on-shutdown logic are unchanged.
- Defensive `if workers <= 0 { workers = 1 }` guard in `NewProcessor` for direct construction without `config.Load`.
- Validation rejects explicit negative values (`processor.workers < 1` after `applyDefaults`).
- `config.example.yaml` documents the new field.

### Implementation note on `setDefaultInt`
`setDefaultInt` treats zero as "unset" (issue #47), so `workers: 0` in YAML is silently promoted to `4`. The test exercises `workers: -1` to bypass defaulting and reach validation — a comment in the test explains this.

## Test plan

- [x] `TestProcessorRunDrainsQueuesOnCancellation` — updated to pass `workers=1`, drain behaviour preserved
- [x] `TestProcessorWorkerPoolAllowsConcurrentProcessing` — new table test: with `workers=3`, peak concurrency for both issue and PR handlers reaches exactly 3
- [x] `TestProcessorWorkersValidation` — new test: `workers=-1` returns a config error mentioning `processor.workers`
- [x] `TestLoadDefaults` extended to assert `cfg.Processor.Workers == 4`
- [x] `go test ./... -count=1` passes

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -85,7 +85,7 @@ func run() error {
 
 	engine := workflow.NewEngine(cfg, runners, promptStore, logger)
 	dataChannels := workflow.NewDataChannels(cfg.Processor.IssueQueueBuffer, cfg.Processor.PRQueueBuffer)
-	processor := workflow.NewProcessor(dataChannels, engine, time.Duration(cfg.HTTP.ShutdownTimeoutSeconds)*time.Second, logger)
+	processor := workflow.NewProcessor(dataChannels, engine, cfg.Processor.Workers, time.Duration(cfg.HTTP.ShutdownTimeoutSeconds)*time.Second, logger)
 	deliveryStore := webhook.NewDeliveryStore(time.Duration(cfg.HTTP.DeliveryTTLSeconds) * time.Second)
 	server := webhook.NewServer(cfg, deliveryStore, dataChannels, schedulerStatusAdapter{scheduler}, logger, scheduler)
 	group, groupCtx := errgroup.WithContext(ctx)
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -85,7 +85,7 @@ func run() error {
 
 	engine := workflow.NewEngine(cfg, runners, promptStore, logger)
 	dataChannels := workflow.NewDataChannels(cfg.Processor.IssueQueueBuffer, cfg.Processor.PRQueueBuffer)
-	processor := workflow.NewProcessor(dataChannels, engine, cfg.Processor.Workers, time.Duration(cfg.HTTP.ShutdownTimeoutSeconds)*time.Second, logger)
+	processor := workflow.NewProcessor(dataChannels, engine, *cfg.Processor.Workers, time.Duration(cfg.HTTP.ShutdownTimeoutSeconds)*time.Second, logger)
 	deliveryStore := webhook.NewDeliveryStore(time.Duration(cfg.HTTP.DeliveryTTLSeconds) * time.Second)
 	server := webhook.NewServer(cfg, deliveryStore, dataChannels, schedulerStatusAdapter{scheduler}, logger, scheduler)
 	group, groupCtx := errgroup.WithContext(ctx)
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -27,6 +27,7 @@ http:
 processor:
   issue_queue_buffer: 256
   pr_queue_buffer: 256
+  workers: 4                              # goroutines per event type (issue / PR); allows concurrent processing across repos
   # Maximum number of agent goroutines that may run concurrently for a single
   # PR-review fan-out event. Limits subprocess resource usage when many agents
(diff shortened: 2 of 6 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config.go:240` — `applyProcessorDefaults` uses `setDefaultInt`, so an explicit `processor.workers: 0` is rewritten to `4` before validation runs. That silently accepts an invalid user config even though the field is documented and validated as `>= 1`, and it recreates the same explicit-zero config bug this repo already fixed for `max_concurrent_agents`. Please preserve the distinction between “omitted” and “set to 0” here as well, and add a load test that `workers: 0` is rejected.
- **[low]** `internal/config/config_test.go:804` — the failure message says `workers=0`, but this fixture sets `workers: -1`. That will be confusing once the explicit-zero case has its own coverage.

### Summary
The worker-pool design is the right fix for #38, but the config path still masks an explicitly invalid `workers: 0` value.

---

eloylp:

1. **workers: 0 not rejected** — Changed `ProcessorConfig.Workers` from `int` to `*int`. `applyProcessorDefaults` now only fills in the default (4) when the pointer is `nil` (field omitted). `validateProcessor` rejects both `0` and negative values, so an explicit `workers: 0` in YAML is now caught at load time.

2. **Misleading test failure message** — Replaced the single `-1` fixture with a table-driven test covering both `workers: 0` and `workers: -1`. Each sub-test now prints the actual value it tested, eliminating the confusing "expected error for workers=0" message that appeared for the `-1` case.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The worker-pool change is now consistent with config validation, preserves shutdown/drain behavior, and has the right targeted coverage for concurrency and explicit-zero config handling.

---

PR #101, opened 2026-04-12
autonomous: re-read memory after each task so next task sees updated content

### LINKED ISSUE #28: autonomous: code task receives stale memory snapshot when issues task updates memory during same run
## Problem

`Scheduler.runAgent` holds a single `MemoryStore.WithLock` across both sequential tasks (issues pass, then code pass) and passes the **same** memory string snapshot to both:

```go
err := s.memories.WithLock(agent.Name, repo, func(memoryPath string, memory string) error {
    if err := s.runTask(ctx, ..., "issues", s.taskPrompts.IssueTask, memoryPath, memory, ...); err != nil {
        return err
    }
    // 'memory' is still the snapshot from before the issues task ran
    return s.runTask(ctx, ..., "code", codeTask, memoryPath, memory, ...)
})
```

The issues-task agent subprocess writes to `memoryPath` (its memory log) during the first pass. By the time the code-task runs, the `memory` string variable still holds the pre-issues-task content. The code-task agent therefore receives an **outdated** memory context — it won't see the issues-task's run log, which reduces the quality of the autonomous guidance prompt.

## Impact

- The code task cannot act on what the issues task wrote (e.g., "I already commented on issues X, Y, Z this run").
- The memory context passed to the code-task prompt gradually drifts from the actual on-disk memory, making the agents' self-consistency guarantees weaker.

## Suggested fix

Re-read the memory file between the two tasks (inside the lock, so no other process can interleave):

```go
err := s.memories.WithLock(agent.Name, repo, func(memoryPath string, memory string) error {
    if err := s.runTask(ctx, ..., "issues", ..., memoryPath, memory, ...); err != nil {
        return err
    }
    // Re-read after the issues task may have written to the file.
    updated, err := os.ReadFile(memoryPath)
    if err != nil {
        return fmt.Errorf("re-read memory after issues task: %w", err)
    }
    codeTask := s.taskPrompts.CodeTask
    if !s.cfg.AllowAutonomousPRs {
        codeTask = s.taskPrompts.CodeTaskNoPRs
    }
    return s.runTask(ctx, ..., "code", codeTask, memoryPath, string(updated), ...)
})
```

Alternatively, split the compound closure into two separate `WithLock` calls (one per task), keeping the closure bodies simple and the memory fresh at each entry.

## Scope

- `internal/autonomous/scheduler.go` — `runAgent` function only (< 10 lines changed).
- No interface or config changes required.

eloylp:
## Summary

Fixes a bug where a multi-task autonomous agent run passed a stale memory snapshot to every task in the loop. Task 1 could write to the memory file, but tasks 2..N still received the pre-run content captured at `WithLock` acquisition time.

- After each `runTask` call, re-read the memory file from disk (still inside the `WithLock` closure, so no concurrent interleaver is possible) and update the local `memory` variable.
- Added `TestRunAgentRefreshesMemoryBetweenTasks`: seeds the memory file with an initial value, has task 1 overwrite it, and asserts task 2's rendered prompt contains the updated content — not the stale snapshot.

Closes #28

## Test plan
- [ ] `go test ./internal/autonomous/... -run TestRunAgentRefreshesMemoryBetweenTasks` passes
- [ ] `go test ./...` passes

### DIFF
--- a/internal/autonomous/scheduler.go
+++ b/internal/autonomous/scheduler.go
@@ -3,6 +3,7 @@ package autonomous
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 	"time"
@@ -256,6 +257,13 @@ func (s *Scheduler) executeAgentRun(ctx context.Context, repo string, agent conf
 			if err := s.runTask(ctx, runner, backend, repo, agent, task.Name, prompt, memoryPath, memory, logger); err != nil {
 				return err
 			}
+			// Re-read the memory file so the next task sees any updates written
+			// by this task rather than the stale snapshot from lock acquisition.
+			updated, err := os.ReadFile(memoryPath)
+			if err != nil {
+				return fmt.Errorf("re-read memory after task %q: %w", task.Name, err)
+			}
+			memory = string(updated)
 		}
 		return nil
 	})
--- a/internal/autonomous/scheduler_test.go
+++ b/internal/autonomous/scheduler_test.go
@@ -609,6 +609,123 @@ func writeGuidance(t *testing.T, dir string, skill string) string {
 	return path
 }
 
+// buildTestPromptStoreWithMemory is like buildTestPromptStore but uses a
+// template that renders {{.Memory}} so tests can assert on its value.
+func buildTestPromptStoreWithMemory(t *testing.T, dir string) *ai.PromptStore {
+	t.Helper()
+	writeIssuePrompt(t, dir)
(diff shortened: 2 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The scheduler now refreshes the per-agent memory snapshot between sequential tasks while still holding the existing lock, and the new regression test covers the stale-memory failure reported in #28.

---

PR #102, opened 2026-04-12
arch: introduce agentRegistry interface in workflow.Engine

### LINKED ISSUE #35: arch: workflow.Engine holds *config.Config directly, leaking all config fields into workflow logic
## Summary

`internal/workflow/engine.go` stores a `*config.Config` pointer:

```go
type Engine struct {
    cfg     *config.Config
    ...
}
```

`Engine` only uses three things from the config:
- `cfg.AgentNames()` — to expand the `all` agent token
- `cfg.AgentByName(name)` — to validate a named agent exists
- `cfg.DefaultConfiguredBackend()` / `cfg.AIBackends[key]` — via the `resolveBackend` helper

Holding the entire `*config.Config` means:
1. Any future addition to `Config` is silently "visible" to `Engine`, widening the coupling surface over time.
2. Tests that exercise `Engine` must construct or mock an entire `Config` struct, including unrelated fields.
3. The `workflow` package has a compile-time dependency on the `config` package even though it only needs three lookups.

## Suggested approach

Pass only the data `Engine` needs at construction time, or define a minimal interface:

```go
// workflow/engine.go
type agentRegistry interface {
    AgentNames() []string
    AgentByName(name string) (config.AgentConfig, bool)
    DefaultConfiguredBackend() string
    ResolveBackend(token string) string  // see issue #31
}

type Engine struct {
    agents  agentRegistry
    runners map[string]ai.Runner
    prompts *ai.PromptStore
    logger  zerolog.Logger
}
```

`*config.Config` already satisfies `agentRegistry` with no changes. This is a similar pattern to the `EventQueue` interface suggested for `webhook.Server` (see issue #33).

This is a medium refactor — mechanical but touches the `Engine` constructor signature and its callers in `main.go` and tests.

eloylp:
## Why

`workflow.Engine` held `*config.Config` directly. This silently exposed every config field to the workflow package, widening the coupling surface every time a new field is added to `Config`. The `Engine` only ever needed three behaviours from config:

1. List all agent names (`AgentNames`)
2. Check whether a named agent exists (`HasAgent`)
3. Map a raw backend token to a configured backend name (`ResolveBackend`)

## What changed

**`internal/workflow/engine.go`**
- Define an unexported `agentRegistry` interface with exactly the three methods above.
- Replace the `*config.Config` field and `NewEngine` parameter with `agentRegistry`. `*config.Config` satisfies the interface directly, so no call site changes are needed.
- The `config` package import is removed from `engine.go` entirely; the workflow package is now free of a compile-time dependency on config's concrete types (matching the existing pattern in `types.go` with `RepoRef`).

**`internal/config/config.go`**
- Add `HasAgent(name string) bool` — a thin wrapper around `AgentByName` — so the interface can express the bool-only lookup without requiring the interface to name `config.AgentConfig`, which would re-introduce the import.

**`internal/config/config_test.go`**
- Extend `TestAgentValidAccepted` with `HasAgent` coverage for both the found and not-found paths.

## Test

```
go test ./...
```

All tests pass.

Closes #35

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"gopkg.in/yaml.v3"
@@ -499,6 +500,22 @@ func (c *Config) AgentNames() []string {
 		names[i] = agent.Name
 	}
 	return names
+}
+
+// HasAgent reports whether an agent with the given name is configured.
+func (c *Config) HasAgent(name string) bool {
+	_, ok := c.AgentByName(name)
+	return ok
+}
+
+// MaxConcurrentAgents returns the configured concurrency limit for agent
+// goroutines. Returns 0 if the field was not set (caller should apply a
+// default).
+func (c *Config) MaxConcurrentAgents() int {
+	if c.Processor.MaxConcurrentAgents == nil {
+		return 0
+	}
+	return *c.Processor.MaxConcurrentAgents
 }
 
 func (c *Config) RepoByName(fullName string) (RepoConfig, bool) {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -506,6 +506,13 @@ repos:
 	if len(names) != 2 {
 		t.Fatalf("expected 2 agent names, got %d", len(names))
 	}
+
+	if !cfg.HasAgent("architect") {
+		t.Fatal("HasAgent: expected true for existing agent")
+	}
+	if cfg.HasAgent("nonexistent") {
+		t.Fatal("HasAgent: expected false for unknown agent")
+	}
 }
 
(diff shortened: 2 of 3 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The engine now depends on the minimal behavior it actually uses, the workflow package no longer imports `config`, and the added `HasAgent` coverage preserves the existing agent lookup semantics without widening the new interface.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
`workflow.Engine` now depends on a narrow behavior-oriented interface, the workflow package no longer imports `config`, and the added coverage keeps the new `HasAgent` helper aligned with the existing lookup semantics.

---

PR #103, opened 2026-04-12
config: normalize PromptFile paths at load time; remove baseDir from Resolve

### LINKED ISSUE #26: config: PromptSourceConfig.Resolve requires callers to supply AgentsDir that config already owns
## Problem

`PromptSourceConfig.Resolve(baseDir string)` is a method on a config struct, but it requires the caller to pass `baseDir` (always `cfg.AgentsDir`) — information that is already inside `*Config`. This is a mild encapsulation leak: callers at the `main` package boundary must know to pass `cfg.AgentsDir`, and the pattern is inconsistently applied.

Currently `main.go` uses two different approaches for the same "inline prompt or read file" operation:

**Task prompts** — calls `Resolve` with `cfg.AgentsDir`:
```go
issueTask, err := cfg.Prompts.AutonomousIssueTask.Resolve(cfg.AgentsDir)
```

**Base prompts** — manually constructs the path without calling `Resolve`:
```go
return ai.PromptSource{PromptFile: filepath.Join(cfg.AgentsDir, src.PromptFile)}
```

Both branches duplicate the same `AgentsDir`-join logic. If `AgentsDir` ever needs a transformation (e.g., evaluated relative to the config file's directory), it must be updated in multiple places.

## Suggested improvement

Normalize `PromptFile` paths to absolute during `applyDefaults`, after `AgentsDir` is resolved. Then `Resolve` needs no parameter:

```go
// In applyDefaults, after setDefault(&c.AgentsDir, defaultAgentsDir):
c.absolutePromptPaths() // join all relative PromptFile fields with AgentsDir

// Resolve becomes:
func (p PromptSourceConfig) Resolve() (string, error) {
    if p.Prompt != "" {
        return p.Prompt, nil
    }
    data, err := os.ReadFile(p.PromptFile)
    if err != nil {
        return "", fmt.Errorf("read prompt file %s: %w", p.PromptFile, err)
    }
    return string(data), nil
}
```

`resolvePrompts()` in `main.go` can similarly be eliminated: after normalization, `src.PromptFile` is already absolute, so `ai.PromptSource{PromptFile: src.PromptFile}` is sufficient without a join.

## Scope

- `internal/config/config.go`: add path normalization step, remove `baseDir` parameter from `Resolve`
- `cmd/agents/main.go`: remove manual `filepath.Join(cfg.AgentsDir, ...)` in `resolvePrompts` and `resolveAgents`
- `internal/config/config_test.go`: update any test that calls `Resolve` with a `baseDir`

Medium refactor; no behavioral change for valid configs.

eloylp:
## Summary

- Adds `Config.absolutePromptPaths()` called from `applyDefaults` to join all relative `PromptFile` fields with `AgentsDir` in-place (absolute paths and empty strings are left unchanged)
- `PromptSourceConfig.Resolve()` and `TaskConfig.Resolve()` now take no parameter — callers no longer need to supply `AgentsDir`
- Removes manual `filepath.Join(cfg.AgentsDir, ...)` from `cmd/agents/main.go` (`resolveSkills`, `resolvePrompts`) and `task.Resolve(s.cfg.AgentsDir)` from `autonomous/scheduler.go`

Closes #26

## Test plan

- [ ] `TestAbsolutePromptPaths` — unit tests for the new normalization function covering relative join, absolute preservation, and empty path passthrough
- [ ] `TestTaskConfigResolve` — updated to call `Resolve()` with no argument
- [ ] `TestLoadAppliesDefaults` — updated assertions to expect joined default paths
- [ ] `TestTaskConfigPromptFileAccepted` — updated to use an explicit `agents_dir` and verify the joined absolute path
- [ ] `go test ./...` passes

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -135,11 +135,7 @@ func setupScheduler(cfg *config.Config, runners map[string]ai.Runner, prompts *a
 func resolveSkills(cfg *config.Config) []ai.SkillGuidance {
 	skills := make([]ai.SkillGuidance, 0, len(cfg.Skills))
 	for _, s := range cfg.Skills {
-		sg := ai.SkillGuidance{Name: s.Name, Prompt: s.Prompt}
-		if s.PromptFile != "" {
-			sg.PromptFile = agentsPath(cfg.AgentsDir, s.PromptFile)
-		}
-		skills = append(skills, sg)
+		skills = append(skills, ai.SkillGuidance{Name: s.Name, Prompt: s.Prompt, PromptFile: s.PromptFile})
 	}
 	return skills
 }
@@ -190,10 +186,7 @@ func (a schedulerStatusAdapter) AgentStatuses() []webhook.AgentStatus {
 
 func resolvePrompts(cfg *config.Config) (issue ai.PromptSource, pr ai.PromptSource, auto ai.PromptSource) {
 	resolve := func(src config.PromptSourceConfig) ai.PromptSource {
-		if src.Prompt != "" {
-			return ai.PromptSource{Prompt: src.Prompt}
-		}
-		return ai.PromptSource{PromptFile: agentsPath(cfg.AgentsDir, src.PromptFile)}
+		return ai.PromptSource{Prompt: src.Prompt, PromptFile: src.PromptFile}
 	}
 	return resolve(cfg.Prompts.IssueRefinement), resolve(cfg.Prompts.PRReview), resolve(cfg.Prompts.Autonomous)
 }
--- a/internal/autonomous/scheduler.go
+++ b/internal/autonomous/scheduler.go
@@ -250,7 +250,7 @@ func (s *Scheduler) executeAgentRun(ctx context.Context, repo string, agent conf
 	}
 	return s.memories.WithLock(agent.Name, repo, func(memoryPath string, memory string) error {
 		for _, task := range agent.Tasks {
-			prompt, err := task.Resolve(s.cfg.AgentsDir)
+			prompt, err := task.Resolve()
 			if err != nil {
 				return fmt.Errorf("resolve prompt for task %q: %w", task.Name, err)
 			}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -186,6 +186,15 @@ func Load(path string) (*Config, error) {
 		return nil, fmt.Errorf("parse config: %w", err)
 	}
 	cfg.applyDefaults()
+	// Resolve AgentsDir to an absolute path so that absolutePromptPaths
+	// produces real absolute file paths regardless of where the daemon is
+	// started from. Relative paths are joined with the config file's
+	// directory so that `agents_dir: agents` (the default) resolves to
+	// <config-dir>/agents, not <cwd>/agents.
+	if !filepath.IsAbs(cfg.AgentsDir) {
+		cfg.AgentsDir = filepath.Join(filepath.Dir(path), cfg.AgentsDir)
+	}
+	cfg.absolutePromptPaths()
 	if err := cfg.validate(); err != nil {
 		return nil, err
 	}
@@ -203,7 +212,6 @@ func (c *Config) applyDefaults() {
 	c.normalizeBackends()
 	c.normalizeRepos()
 	c.normalizeAutonomousAgents()
-	c.absolutePromptPaths()
 	c.resolveWebhookSecret()
 	c.resolveAPIKey()
 }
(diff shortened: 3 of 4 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config.go:202` — `absolutePromptPaths()` does not actually make paths absolute when `agents_dir` itself is relative, which is still the documented default (`agents` / `./agents`). `filepath.Join(c.AgentsDir, p)` preserves a relative base, so after this change `Resolve()` and `ai.PromptSource` are relying on a contract that is false for the common case. That leaves prompt resolution tied to the process working directory and contradicts both the new method comments and the linked issue’s goal of normalizing to absolute paths at load time. This should either call `filepath.Abs`/resolve relative to the config location before storing the paths, or keep the old base-dir join behavior.

### Summary
Removing the `Resolve` parameter is fine, but the new normalization step needs to produce real absolute paths before the code can rely on that invariant.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config.go:204` — no additional issues beyond the unresolved path-normalization bug already raised in the existing discussion. `absolutePromptPaths()` still joins relative prompt files onto `AgentsDir` without making the result absolute, so a relative `agents_dir` still leaves prompt resolution dependent on the process working directory.

### Summary
I do not see a new issue beyond the existing blocker, but that absolute-path invariant still is not true on the current head.

---

eloylp:
Addressed the review findings: moved `absolutePromptPaths()` out of `applyDefaults()` and into `Load()`, called after an explicit step that resolves a relative `AgentsDir` to an absolute path anchored on `filepath.Dir(configPath)`. A relative `agents_dir: agents` in a config at `/etc/daemon/config.yaml` now becomes `/etc/daemon/agents` rather than `<cwd>/agents`, so prompt paths are truly absolute after `Load` returns regardless of the process working directory. Added `TestLoadMakesAgentsDirAbsoluteRelativeToConfigFile` to pin this behaviour.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/config/config.go:187` — the new "absolute after Load" invariant still fails when the caller passes a relative config path, which is the default CLI path (`-config config.yaml`). `filepath.Dir("config.yaml")` is `"."`, so `filepath.Join(filepath.Dir(path), cfg.AgentsDir)` leaves `AgentsDir` relative (`./agents`), and `absolutePromptPaths()` then preserves relative prompt paths as well. That means prompt resolution is still process-CWD-dependent in the common startup path. Resolve `path` itself with `filepath.Abs` (or otherwise anchor it) before deriving `AgentsDir`, and add a test that calls `Load` with a relative config filename.

### Summary
Anchoring `AgentsDir` to the config file is the right fix, but the current implementation still leaks a relative-path case through the default `-config config.yaml` entrypoint.

---

eloylp:
Addressed: added `filepath.Abs(path)` call before deriving `AgentsDir`. `filepath.Dir("config.yaml")` returns `"."`, so joining a relative `agents_dir` onto it still left everything relative and CWD-dependent. Now the config path itself is made absolute first, so `filepath.Dir(absPath)` always points to the config file's real directory. Added a `relative-config-path` subcase to `TestLoadMakesAgentsDirAbsoluteRelativeToConfigFile` that `os.Chdir`s into the temp dir and calls `Load("config.yaml")` to pin this explicitly.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The current head makes prompt paths genuinely absolute after `Load`, updates the call sites to rely on that invariant cleanly, and adds the relative-config-path coverage that was missing before.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Prompt-file normalization is now anchored to the real config location, the `Resolve()` callers no longer need to thread `AgentsDir` through, and the relative-config-path tests cover the subtle case that was missing earlier.

---

eloylp:
Closing this PR. After reviewing the current state of main, the codebase has evolved to a different design that no longer has the pattern described in issue #26. The current main uses  (resolving paths relative to ) and does not have  or a separate  field. The encapsulation leak this PR was fixing no longer exists in the current codebase. The issue can be closed as resolved by the intervening refactors.

PR #111, opened 2026-04-13
test(webhook): remove pre-Go 1.22 loop variable capture

eloylp:
## Why

`tc := tc` inside a `for _, tc := range` loop was necessary before Go 1.22 to prevent the closure from capturing a single shared loop variable. Since Go 1.22, each iteration gets its own variable, so the shadowing assignment is dead code.

This module declares `go 1.24.12`, so the pattern has been obsolete since the module was created.

## What changed

Removed the single `tc := tc` line in `TestHandleAgentsRunReturnsBadRequestOnMissingFields` in `internal/webhook/server_test.go`. No behavioral change.

### DIFF
--- a/internal/webhook/server_test.go
+++ b/internal/webhook/server_test.go
@@ -280,7 +280,6 @@ func TestHandleAgentsRunReturnsBadRequestOnMissingFields(t *testing.T) {
 		{"empty body", `{}`},
 	}
 	for _, tc := range tests {
-		tc := tc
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Test-only cleanup is correct for Go 1.22+ semantics and does not change behavior.

---

PR #115, opened 2026-04-14
test(autonomous): add t.Parallel() to MemoryStore tests

eloylp:
## Why

`TestMemoryStoreCreatesPerRepoMemory` and `TestMemoryStoreRejectsPathEscapes` are pure unit tests for a stateless function. Each test creates its own isolated directory via `t.TempDir()`, writes and reads only within that directory, and touches no env vars or global state. They are safe to run in parallel.

`scheduler_test.go` in the same package already calls `t.Parallel()` on all its tests — these two were the only outliers in `autonomous`.

## What

- Add `t.Parallel()` to `TestMemoryStoreCreatesPerRepoMemory`.
- Add `t.Parallel()` to `TestMemoryStoreRejectsPathEscapes`.

No logic changes.

### DIFF
--- a/internal/autonomous/memory_test.go
+++ b/internal/autonomous/memory_test.go
@@ -8,6 +8,7 @@ import (
 )
 
 func TestMemoryStoreCreatesPerRepoMemory(t *testing.T) {
+	t.Parallel()
 	dir := t.TempDir()
 	store := NewMemoryStore(dir)
 	var capturedPath string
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Adding `t.Parallel()` at `internal/autonomous/memory_test.go:11` and `internal/autonomous/memory_test.go:52` is appropriate here because each test uses its own `t.TempDir()` and its own `MemoryStore`, so the change improves test scheduling without changing behavior.

---

PR #116, opened 2026-04-14
test(config): remove pre-Go 1.22 loop variable captures

eloylp:
## Why

`TestLoadAcceptsValidLogLevels` (line 499) and `TestLoadAcceptsValidLogFormats` (line 546) in `internal/config/config_test.go` both contain `level := level` / `format := format` shadow assignments inside their `for` loops. These were the pre-Go 1.22 workaround for per-iteration variable capture in goroutines/closures.

Go 1.22 changed loop variable semantics so each iteration gets its own copy — the shadow is a no-op and dead code. The project is on Go 1.24 (go.mod). This is the same cleanup applied to `internal/webhook/server_test.go` in #111.

## What

- Remove `level := level` from `TestLoadAcceptsValidLogLevels`.
- Remove `format := format` from `TestLoadAcceptsValidLogFormats`.

No behaviour change. CI (`go test ./... -race`) will confirm.

PR #118, opened 2026-04-14
test(webhook): add t.Parallel() to all server tests

eloylp:
## Why

`internal/webhook/server_test.go` has 11 top-level tests plus one subtest, none of which called `t.Parallel()`. Each test constructs its own `Server`, `DeliveryStore`, and `DataChannels` — there is no shared mutable state between tests. Running them in parallel surfaces data races early and reduces overall test latency.

## What changed

Added `t.Parallel()` as the first statement in:
- all 11 top-level `Test*` functions
- the `t.Run` subtest inside `TestHandleAgentsRunReturnsBadRequestOnMissingFields`

No logic was changed.

### DIFF
--- a/internal/webhook/server_test.go
+++ b/internal/webhook/server_test.go
@@ -52,6 +52,7 @@ func signatureForTests(body []byte, secret string) string {
 }
 
 func TestHandleIssueWebhookDeduplicatesDelivery(t *testing.T) {
+	t.Parallel()
 	cfg := testCfg(nil)
 	dataChannels := workflow.NewDataChannels(1, 1)
 	server := NewServer(cfg, NewDeliveryStore(time.Hour), dataChannels, nil, zerolog.Nop(), nil)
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/server_test.go:291` — adding `t.Parallel()` inside the `t.Run` body turns this loop-variable capture into a real test bug. Each subtest now runs after the parent loop advances, so all three cases can observe the final `tc` value (`{"empty body": {}}`) instead of their own input. That weakens coverage and can hide regressions in the missing-agent and missing-repo paths. Rebind with `tc := tc` before `t.Run` or pass `tc` into the closure.

### Summary
Parallelizing the top-level tests is fine, but the parallel subtest currently collapses the table-driven coverage to the last case.

---

eloylp:
## Response to PR Review

Thank you for the review. The loop variable capture concern you raised would be valid for Go < 1.22, but this module declares  in .

Since Go 1.22, loop variables are automatically per-iteration, so each subtest closure captures its own copy of `tc` — no `tc := tc` rebind is needed. In fact, PR #116 was specifically about *removing* those rebind patterns from existing tests as they became unnecessary.

The parallel subtests here correctly observe their own `tc` values. No change is needed to the loop; the test coverage is intact across all three cases (missing agent, missing repo, empty body).

PR #123, opened 2026-04-14
refactor(workflow): replace containsLabel loop with slices.ContainsFunc

eloylp:
## Why

`containsLabel` in `internal/workflow/engine.go` uses a manual for-loop to search a `[]string` with a predicate — the exact pattern `slices.ContainsFunc` was introduced for in Go 1.21. This project targets Go 1.24, so the stdlib function is already available.

## What changed

- Replaced the 5-line manual loop with `slices.ContainsFunc` and a one-line predicate.
- Added `"slices"` to the import block.
- `"strings"` is kept — `strings.ToLower` / `strings.TrimSpace` are still used inside the predicate and elsewhere in the file.
- No behaviour change: the normalisation (trim + lower) is preserved inside the predicate.

### DIFF
--- a/internal/workflow/engine.go
+++ b/internal/workflow/engine.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Small refactor only; behavior is preserved and the `slices.ContainsFunc` replacement is idiomatic for the project's Go version.

---

PR #125, opened 2026-04-15
refactor(config): replace isValidBackendName loop with slices.Contains

eloylp:
## Why

`isValidBackendName` in `internal/config/config.go` uses a 6-line manual for-loop to search a fixed `[]string` — the exact pattern `slices.Contains` was introduced for in Go 1.21. This project targets Go 1.24, so the stdlib function is already available.

Consistent with the same refactor applied to `containsLabel` in `internal/workflow/engine.go` (PR #123).

## What changed

- Replaced the 6-line manual loop in `isValidBackendName` with `slices.Contains`.
- Added `"slices"` to the import block.
- No behaviour change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #127, opened 2026-04-15
refactor(autonomous): replace bound-check loop with slices.ContainsFunc

eloylp:
## Why

`TriggerAgent` contains a classic bool-flag loop that checks whether any enabled binding in `repoDef.Use` references the requested agent:

```go
bound := false
for _, b := range repoDef.Use {
    if b.Agent == agentName && b.IsEnabled() {
        bound = true
        break
    }
}
if !bound {
    return fmt.Errorf(...)
}
```

This is the same pattern already removed from `engine.go` (#123) and `config.go` (#125). `slices.ContainsFunc` expresses the same predicate directly, eliminating the auxiliary variable and the early-return loop.

## Change

`internal/autonomous/scheduler.go` — replace the 6-line loop with a single `slices.ContainsFunc` guard; add `"slices"` to the import block.

No behaviour change.

### DIFF
--- a/internal/autonomous/scheduler.go
+++ b/internal/autonomous/scheduler.go
@@ -3,6 +3,7 @@ package autonomous
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Small, behavior-preserving refactor. The `slices.ContainsFunc` predicate matches the previous loop and the module already targets Go 1.24, so this is safe and idiomatic.

---

PR #128, opened 2026-04-15
refactor(config): collapse duplicate logLevelYAML/logFormatYAML into one helper

eloylp:
## Why

`internal/config/config_test.go` contained two nearly-identical 25-line YAML helper functions:

```go
func logLevelYAML(level string) string { ... }   // log: level: <value>
func logFormatYAML(format string) string { ... } // log: format: <value>
```

The only difference between them is the sub-key name under `log:`. This is textbook copy-paste duplication — any future change to the shared skeleton (e.g. adding a new required field) would require editing both.

## Change

Merge both helpers into a single parameterised function:

```go
func logConfigYAML(field, value string) string { ... }
```

Update the four call sites accordingly. No behaviour change; all existing tests continue to exercise the same YAML variations.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -393,7 +393,7 @@ repos:
 	}
 }
 
-func logLevelYAML(level string) string {
+func logConfigYAML(field, value string) string {
 	return `
 daemon:
   http:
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Test-only helper consolidation is behavior-preserving and the updated call sites remain clear.

---

PR #130, opened 2026-04-15
refactor(ai): replace Msg(fmt.Sprintf(...)) with Msgf(...) in cmdrunner

eloylp:
## Why

`zerolog.Event.Msg(fmt.Sprintf(format, args...))` is non-idiomatic and
wasteful: `fmt.Sprintf` always allocates an intermediate string, even
when the log level is disabled and the message will never be emitted.

`zerolog.Event.Msgf(format, args...)` is the built-in equivalent —
it formats lazily (or not at all when the level is filtered) and reads
more naturally at the call site.

## What changed

`internal/ai/cmdrunner.go`: replaced all 9 occurrences of
`Msg(fmt.Sprintf(...))` with the equivalent `Msgf(...)` call.
No behaviour change; purely syntactic.

## Test plan
- [ ] No logic changes — diff is mechanical substitution of one zerolog API for another

### DIFF
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -59,13 +59,13 @@ func (r *CommandRunner) Run(ctx context.Context, req Request) (Response, error) 
 
 	switch r.mode {
 	case "noop":
-		logger.Info().Msg(fmt.Sprintf("%s runner noop", r.backendName))
+		logger.Info().Msgf("%s runner noop", r.backendName)
 		return Response{}, nil
 	case "command":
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Mechanical `Msg(fmt.Sprintf(...))` to `Msgf(...)` substitution only; the change is idiomatic and preserves behavior while avoiding unnecessary formatting work when the log level is disabled.

---

PR #138, opened 2026-04-15
feat(proxy): built-in Anthropic↔OpenAI translation proxy

### LINKED ISSUE #137: ai: built-in Anthropic↔OpenAI translation proxy (replace LiteLLM dep)
## Goal

Add a built-in Anthropic Messages ↔ OpenAI Chat Completions translation proxy to the daemon, so users can point `claude` CLI at our daemon and run against any OpenAI-compatible backend (local llama.cpp, Ollama, vLLM, hosted Qwen, etc.) without a separate Python sidecar like LiteLLM.

This lands in **one PR**. No external runtime dependencies. Pure Go, lives in the existing daemon binary, configured from `config.yaml`.

## Context

Tonight we validated the proxy architecture end-to-end using LiteLLM as the translator (see [#78](https://github.com/eloylp/agents/issues/78#issuecomment-4255860608)). The result: a `claude` CLI with `ANTHROPIC_BASE_URL` pointing at LiteLLM successfully round-trips through an OpenAI-compatible backend into a local model.

LiteLLM works but carries a Python dependency that doesn't fit the single-binary Go deployment story. This issue replaces that dependency with a native translator inside the daemon.

## North star

Single binary. One process. One config file. The daemon is already an HTTP server (`/webhooks/github`, `/status`, `/agents/run`). Adding `/v1/messages` as another handler keeps the deployment surface unchanged.

### 1. New package `internal/anthropic_proxy/`

Request/response translation between Anthropic Messages API and OpenAI Chat Completions. Pure functions, no I/O, no HTTP. Testable in isolation.

Covered by v1:

- **Text turns.** User/assistant/system messages with string content.
- **Tool use.** Anthropic `tool_use` and `tool_result` content blocks ↔ OpenAI `tool_calls` and `tool` role messages. Our agents emit JSON via tool use today via the claude CLI, so this must work.
- **Multi-turn conversations.** Arrays of messages, preserving order and tool_call ↔ tool_result linkage.
- **Stop reasons.** OpenAI `finish_reason: stop|length|tool_calls` ↔ Anthropic `stop_reason: end_turn|max_tokens|tool_use`.
- **Usage counters.** `prompt_tokens` / `completion_tokens` ↔ `input_tokens` / `output_tokens`.

Explicitly NOT covered by v1 (follow-up tickets):

- **Streaming.** v1 is request/response only. Claude CLI tolerates this; streaming can come later.
- **Vision / images.** Our agents don't use them.
- **Prompt caching control blocks.** Anthropic-specific; no OpenAI equivalent.
- **Extended thinking / reasoning_content mapping.** Backends may emit `reasoning_content`; v1 drops it (matches current LiteLLM behavior).

### 2. HTTP handler mounted on the existing server

`internal/webhook/server.go` (or wherever the HTTP mux is defined) gains a new route, only when proxy mode is enabled:

```
POST /v1/messages
```

Request body is Anthropic Messages format. Handler:

1. Parses Anthropic request.
2. Translates to OpenAI ChatCompletions request.
3. Injects any `extra_body` configured (e.g. `chat_template_kwargs: {enable_thinking: false}` for Qwen 3.5).
4. POSTs to the configured upstream OpenAI-compat endpoint with the configured model name and API key.
5. Translates the OpenAI response back to Anthropic shape.
6. Returns it.

Errors: upstream failures surface as HTTP 502 with a short Anthropic-shaped error body.

### 3. Config

Extend `DaemonConfig` with a new section:

```yaml
daemon:
  proxy:
    enabled: false                      # default off
    path: /v1/messages                  # default, configurable
    upstream:
      url: http://ts-vortex:8001/v1     # OpenAI-compat endpoint
      model: qwen                       # model name sent to upstream
      api_key_env: LOCAL_LLM_API_KEY    # optional; empty string if absent
      timeout_seconds: 120
      extra_body:                       # forwarded verbatim on every request
        chat_template_kwargs:
          enable_thinking: false
```

Validation:

- `proxy.enabled=true` requires `proxy.upstream.url` and `proxy.upstream.model`.
- `proxy.path` must start with `/`.
- `timeout_seconds` must be > 0.

When `proxy.enabled=false` (default), the handler isn't mounted and no behaviour changes.

### 4. Tests

Table-driven coverage for the pure translation functions:

- Text-only turn round-trip.
- System message placement (Anthropic top-level `system` ↔ OpenAI `role: system` as first message).
- Single tool use + tool result round-trip.
- Multi-turn with interleaved tool calls.
- Stop reason mapping table (all values).
- Usage counter preservation.
- Malformed Anthropic request → clear 400 error.

HTTP handler integration tests:

- Fake upstream OpenAI server (`httptest.Server`) returning canned responses.
- Proxy forwards correctly, translation is applied both directions.
- `extra_body` is injected.
- Upstream 500 surfaces as 502 with Anthropic-shaped error.
- Missing API key when `api_key_env` is configured → 500 at startup validation, not request time.

### 5. Documentation updates

- `README.md`: new "Local / OpenAI-compatible backends" section. Show the proxy config and a `ANTHROPIC_BASE_URL=http://localhost:8080 claude -p "hi"` usage example.
- `CLAUDE.md`: one-liner mention of the proxy endpoint under HTTP endpoints list.
- `config.example.yaml`: commented proxy block (disabled by default) showing the Qwen 3.5 + llama.cpp recipe.

### 6. Explicit cleanup — no dead code left behind

- Any existing LiteLLM references in docs get replaced or annotated with "alternative; built-in proxy is the default".
- No shims, no backcompat wrappers. `proxy.enabled: false` is the default; opt-in.

## Acceptance

- `go test ./... -race` green.
- New `/v1/messages` endpoint returns valid Anthropic-shape responses for text turns and tool-use turns against a fake upstream.
- A manual end-to-end test: run `llama-server` on some port, configure the proxy in `config.yaml`, start the daemon, hit `curl http://localhost:8080/v1/messages -d '{"model":"...","max_tokens":80,"messages":[{"role":"user","content":"hi"}]}'` — response is Anthropic-shape.
- `ANTHROPIC_BASE_URL=http://localhost:8080 claude -p "hi"` from another host returns output generated by the upstream OpenAI-compat backend.
- LiteLLM is no longer required for the single-upstream, single-model use case.

## Out of scope / follow-ups

File (but don't implement) these as separate tickets once this lands:

- `ai: anthropic proxy — streaming support`
- `ai: anthropic proxy — multi-upstream routing (model → upstream map)`
- `ai: anthropic proxy — prompt caching control block translation`
- `ai: anthropic proxy — auth (require Bearer token to proxy, separate from upstream key)`

## Related

- Depends on none — additive.
- Unblocks dropping the LiteLLM dep in the [#78](https://github.com/eloylp/agents/issues/78) research setup.
- Complements [#132](https://github.com/eloylp/agents/issues/132) (system/user prompt split) — once that lands, the proxy would translate the split shape into `messages: [{role: system}, {role: user}]` naturally.

eloylp:
## Summary

- New package `internal/anthropic_proxy/` with pure translation functions (`ToOpenAI`, `ToAnthropic`) and an HTTP handler. Covers text turns, tool use/result blocks, system messages, stop-reason mapping, and usage counters.
- Config: `daemon.proxy` section (disabled by default). Validated at load time — URL and model required when enabled, path must start with `/`, timeout must be positive, API key resolved from env.
- Server: `NewServer` builds and mounts the proxy handler when `proxy.enabled: true`.
- Docs: README "Local / OpenAI-compatible backends" section, CLAUDE.md HTTP endpoint list, `config.example.yaml` commented proxy block.

## Test plan

- [ ] `go test ./... -race` green
- [ ] `TestToOpenAI_*` and `TestToAnthropic_*` cover text, system, tool use, tool result, multi-turn, stop reasons, and usage counters
- [ ] `TestHandler_*` covers happy path, translation applied, extra_body injection, upstream 500 → 502, malformed request → 400, missing max_tokens → 400
- [ ] `TestProxy*` config tests cover defaults, valid load with api_key_env resolution, and validation errors (missing url, missing model, bad path, non-positive timeout)

Closes #137

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,6 +69,7 @@ Multi-stage build on `node:22-alpine` so the image includes Claude Code, Codex, 
   - `GET /status` — JSON with uptime, queue depths, agent schedules
   - `POST /webhooks/github` — HMAC-verified webhook receiver
   - `POST /agents/run` — on-demand agent trigger (requires Bearer token)
+  - `POST /v1/messages` — Anthropic↔OpenAI translation proxy (disabled by default; enabled via `daemon.proxy.enabled: true`)
 - Relevant webhook events: `issues.labeled`, `pull_request.labeled`. Trigger label comes from `payload.label.name`.
 - Duplicate webhook suppression via `X-GitHub-Delivery` TTL cache.
 - Workflow execution is stateless in-process. Only autonomous agents persist memory (per-agent, per-repo markdown file under `memory_dir`).
--- a/README.md
+++ b/README.md
@@ -383,10 +383,40 @@ GitHub sends a ping immediately; the daemon will log the delivery.
 | `GET` | `/status` | none | Health check: JSON with uptime, queue depths, agent schedules |
 | `POST` | `/webhooks/github` | `X-Hub-Signature-256` HMAC | GitHub webhook receiver |
 | `POST` | `/agents/run` | `Authorization: Bearer <key>` | On-demand agent trigger |
+| `POST` | `/v1/messages` | none | Anthropic↔OpenAI translation proxy (opt-in via `proxy.enabled`) |
 
 The `/agents/run` body is `{"agent": "<name>", "repo": "owner/repo"}`. It blocks until the agent finishes. If `api_key_env` is not configured the endpoint returns `403 Forbidden`.
 
 Duplicate webhook deliveries are suppressed via `X-GitHub-Delivery` with a TTL cache.
+
+---
+
+## Local / OpenAI-compatible backends
+
+The daemon includes a built-in Anthropic Messages ↔ OpenAI Chat Completions translation proxy. Enable it to run the `claude` CLI against any OpenAI-compatible backend (Ollama, llama.cpp, vLLM, Qwen, etc.) without a separate sidecar.
+
+```yaml
+daemon:
+  proxy:
+    enabled: true
+    path: /v1/messages          # default; configurable
+    upstream:
+      url: http://localhost:8001/v1   # OpenAI-compat base URL
+      model: qwen3                    # model name sent upstream
+      api_key_env: LOCAL_LLM_API_KEY  # optional; omit if not required
+      timeout_seconds: 120
+      extra_body:                     # forwarded verbatim on every request
+        chat_template_kwargs:
+          enable_thinking: false
+```
+
+Then point the Claude CLI at the daemon:
+
+```bash
+ANTHROPIC_BASE_URL=http://localhost:8080 claude -p "Summarise the repo."
+```
+
+The proxy translates text turns, tool use / tool results, system messages, stop reasons, and usage counters. Streaming, vision, and prompt-caching control blocks are not supported in v1.
 
 ---
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -26,6 +26,23 @@ daemon:
 
   memory_dir: /var/lib/agents/memory            # persistent agent memory root
 
+  # Built-in Anthropic↔OpenAI translation proxy (disabled by default).
+  # Enable to run `ANTHROPIC_BASE_URL=http://localhost:8080 claude -p "hi"`
+  # against any OpenAI-compatible backend (llama.cpp, Ollama, vLLM, etc.)
+  # without a separate sidecar like LiteLLM.
+  #
+  # proxy:
+  #   enabled: false
+  #   path: /v1/messages          # path the claude CLI hits
+  #   upstream:
+  #     url: http://ts-vortex:8001/v1   # OpenAI-compat base URL
+  #     model: qwen3                    # model name forwarded upstream
+  #     api_key_env: LOCAL_LLM_API_KEY  # optional; empty if not required
+  #     timeout_seconds: 120
+  #     extra_body:                     # merged into every upstream request
+  #       chat_template_kwargs:
+  #         enable_thinking: false
+
   ai_backends:
     claude:
       command: claude
--- a/internal/anthropic_proxy/handler.go
+++ b/internal/anthropic_proxy/handler.go
@@ -1,0 +1,167 @@
+package anthropicproxy
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+// UpstreamConfig holds the connection parameters for the OpenAI-compatible upstream.
+type UpstreamConfig struct {
+	URL       string
+	Model     string
+	APIKey    string
+	Timeout   time.Duration
+	ExtraBody map[string]any
+}
+
+// Handler is an HTTP handler that accepts Anthropic Messages API requests,
+// translates them to OpenAI Chat Completions, proxies them to an upstream, and
+// returns the response in Anthropic format.
+type Handler struct {
+	upstream UpstreamConfig
+	client   *http.Client
+	logger   zerolog.Logger
+}
+
+// NewHandler creates a Handler with the given upstream configuration.
+func NewHandler(upstream UpstreamConfig, logger zerolog.Logger) *Handler {
+	return &Handler{
+		upstream: upstream,
+		client:   &http.Client{Timeout: upstream.Timeout},
+		logger:   logger.With().Str("component", "anthropic_proxy").Logger(),
+	}
+}
+
+// ServeHTTP implements http.Handler.
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	body, err := io.ReadAll(io.LimitReader(r.Body, 4<<20)) // 4 MiB limit
+	if err != nil {
+		h.writeError(w, http.StatusBadRequest, "invalid_request_error", "failed to read request body")
+		return
+	}
+
+	var anthReq MessagesRequest
+	if err := json.Unmarshal(body, &anthReq); err != nil {
+		h.writeError(w, http.StatusBadRequest, "invalid_request_error", "malformed Anthropic request: "+err.Error())
+		return
+	}
+	if anthReq.MaxTokens <= 0 {
+		h.writeError(w, http.StatusBadRequest, "invalid_request_error", "max_tokens is required and must be positive")
+		return
+	}
+	if len(anthReq.Messages) == 0 {
+		h.writeError(w, http.StatusBadRequest, "invalid_request_error", "messages must not be empty")
+		return
+	}
+
+	oaiReq, err := ToOpenAI(anthReq, h.upstream.Model)
+	if err != nil {
+		h.writeError(w, http.StatusBadRequest, "invalid_request_error", "translation to OpenAI format failed: "+err.Error())
+		return
+	}
+
+	reqBody, err := h.marshalWithExtraBody(oaiReq)
+	if err != nil {
+		h.logger.Error().Err(err).Msg("failed to marshal upstream request")
+		h.writeError(w, http.StatusInternalServerError, "api_error", "internal marshalling error")
+		return
+	}
+
+	upstreamURL := h.upstream.URL + "/chat/completions"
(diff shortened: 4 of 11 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/anthropic_proxy/translate.go:75` — `translateUserMessage` collapses every text block in an Anthropic user turn into one `role:user` message and then prepends it ahead of all `tool_result` messages. That reorders mixed content like `[tool_result, text]` or `[text, tool_result, text]`, even though the issue/PR contract explicitly calls out preserving multi-turn message order and tool_call ↔ tool_result linkage. The translator already returns `[]ChatMessage`, so this needs to emit user/tool segments in block order instead of moving all text to the front.
- **[medium]** `internal/config/config.go:407` — `api_key_env` is resolved opportunistically, but proxy startup validation never fails when the named environment variable is unset. In that state the daemon boots, the proxy sends no `Authorization` header, and every request fails at runtime against a protected upstream. The issue and PR test plan both say this should fail at load/startup time instead.

### Summary
The proxy shape is good, but it still loses valid Anthropic message ordering for mixed text/tool-result turns and it doesn't enforce the configured upstream API key at startup.

---

eloylp:
## Findings addressed — $(date -u +%Y-%m-%dT%H:%M:%SZ) review

**[high] translateUserMessage reorders mixed text/tool_result content (translate.go:75)**

Fixed. `translateUserMessage` now iterates blocks in their original order. Text blocks accumulate into a `pending` buffer and are flushed as `role:user` when a `tool_result` is encountered (or at the end). A turn with `[tool_result, text]` now produces `[role:tool, role:user]` instead of the old `[role:user, role:tool]`. Plain string content (no block array) continues to emit a single `role:user` as before.

Added `TestToOpenAI_MixedUserTurnOrder` which asserts the `[tool_result, text]` ordering is preserved end-to-end through `ToOpenAI`.

**[medium] proxy startup does not fail when api_key_env variable is unset (config.go:407)**

Fixed. `validateProxy` now returns an error when `proxy.upstream.api_key_env` is set but resolves to an empty string (env var unset or blank). A misconfigured or missing key is caught at load time rather than producing silent 401/403 failures against the upstream.

Added `"api_key_env set but variable unset"` case in `TestProxyValidationErrors`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/anthropic_proxy/translate.go:96` — unsupported Anthropic content blocks are still silently dropped instead of being rejected. In `translateUserMessage` every block type other than `text` and `tool_result` falls through, and `translateAssistantMessage` similarly skips every non-`tool_use` block. Because `ServeHTTP` only returns `400` when translation fails, v1 out-of-scope inputs such as images, thinking blocks, or prompt-caching blocks currently get accepted and lossy-translated into empty or partial messages instead of failing clearly. Add explicit validation/default cases so unsupported block types return a translation error.

### Summary
The proxy wiring is in good shape, but unsupported Anthropic block types are still accepted with silent content loss instead of a clear 400.

---

eloylp:
## Findings addressed — 2026-04-15 review

**[medium] Unsupported Anthropic content block types silently dropped (translate.go:96)**

Fixed. Both `translateUserMessage` and `translateAssistantMessage` now have explicit `default` cases that return a translation error for any block type outside the supported v1 set:
- User: `text` and `tool_result` are supported; everything else (images, thinking, document blocks, etc.) returns `unsupported user content block type %q`.
- Assistant: `text` (already collected via `parseContent`) and `tool_use` are supported; everything else returns `unsupported assistant content block type %q`.

Because `ServeHTTP` returns `400` when translation returns an error, unsupported inputs now get a clear rejection rather than lossy partial translation.

Added `TestToOpenAI_UnsupportedUserBlock` and `TestToOpenAI_UnsupportedAssistantBlock` which assert that `ToOpenAI` returns a non-nil error when a block of type `"image"` / `"thinking"` is present.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/anthropic_proxy/translate.go:285` — `toolResultContent` still silently drops unsupported nested content blocks inside a `tool_result` array instead of rejecting them. The latest fix correctly errors on unsupported top-level user/assistant block types, but a payload like `[{"type":"image",...}]` or mixed `[text, image]` inside `tool_result.content` is currently accepted and lossy-translated into `""` or partial text. That breaks the same “unsupported v1 content must fail clearly” contract the PR already applies elsewhere. Add explicit validation/default handling for non-`text` blocks in `toolResultContent` and cover it with a regression test.

### Summary
The proxy wiring is solid, but one remaining nested `tool_result` path still accepts unsupported Anthropic content with silent data loss instead of a clear 400.

---

eloylp:
## Findings addressed — 2026-04-15 review

**[medium] Unsupported nested block types inside tool_result silently dropped (translate.go:285)**

Fixed. The \`toolResultContent\` block loop now has an explicit \`default\` case that returns \`fmt.Errorf("unsupported tool_result content block type %q", b.Type)\` for any non-\`text\` block. Mixed or unknown payloads (images, thinking blocks, etc.) inside \`tool_result.content\` arrays now cause \`ServeHTTP\` to return 400 instead of silently producing lossy/partial text.

Added \`TestToOpenAI_UnsupportedToolResultBlock\`: passes a \`tool_result\` whose content array contains an \`"image"\` block and asserts \`ToOpenAI\` returns a non-nil error.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The current head closes the earlier translation gaps cleanly: unsupported Anthropic blocks now fail fast instead of being dropped, mixed `tool_result`/text user turns preserve order, and the proxy/config wiring stays consistent with the issue contract.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/anthropic_proxy/translate.go:131` — assistant-side mixed `text`/`tool_use` ordering is still lossy. `parseContent` pre-concatenates every assistant text block into one string (`translate.go:224-250`), and `translateAssistantMessage` then skips the original `text` blocks while appending all `tool_use` blocks afterward. A valid Anthropic assistant turn like `[text("before"), tool_use(A), text("after")]` therefore round-trips back as `[text("before\nafter"), tool_use(A)]`, moving the trailing text ahead of the tool call. That breaks the PR/issue contract to preserve message order for multi-turn tool-use conversations. This needs either explicit rejection of mixed assistant block ordering in v1 or a translation strategy that preserves block order, plus a regression test for an interleaved assistant turn.

### Summary
The proxy is close, but assistant turns with interleaved text and tool_use blocks are still accepted with order-changing translation.

---

eloylp:
## Findings addressed — 2026-04-15 review

**[medium] Interleaved text/tool_use assistant block ordering silently reordered (translate.go:131)**

Fixed. `translateAssistantMessage` now validates block ordering before attempting translation: it walks the blocks looking for any `text` block that follows a `tool_use` block, and returns `fmt.Errorf("unsupported assistant block ordering: text block follows tool_use block")` if found. `ServeHTTP` already returns 400 on any translation error, so interleaved turns now get a clear rejection instead of lossy re-ordering.

The common `[text, tool_use]` pattern (leading text, trailing tools) is unaffected — it passes the ordering check and translates faithfully.

Added two regression tests:
- `TestToOpenAI_InterleavedAssistantBlocks`: passes `[text, tool_use, text]` and asserts a non-nil error.
- `TestToOpenAI_LeadingTextThenToolUseIsValid`: passes `[text, tool_use]` and asserts successful translation with correct `content` and `tool_calls`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The current head matches the issue contract cleanly: unsupported Anthropic block shapes fail fast, mixed user/tool-result ordering is preserved, and the proxy/config path is consistent with the documented startup validation and handler behavior.

---

PR #145, opened 2026-04-16
refactor(proxy): replace any round-trip with json.Compact in rawJSONToString/stringToRawJSON

eloylp:
## Why

`rawJSONToString` and `stringToRawJSON` both validated and compacted JSON
by unmarshalling to `any` then re-marshalling. This round-trip has two
drawbacks:

1. **Allocation**: it boxes the entire JSON value into an `any`, allocates
   a `map[string]any`, and then re-allocates all strings/numbers.
2. **Key sorting**: `json.Marshal` on a `map` sorts keys alphabetically,
   silently reordering object fields in the output. For tool arguments this
   doesn't affect semantics, but it's an invisible side-effect.

## What changed

Both functions now use `json.Compact`, which removes whitespace and
validates the JSON in a single pass with no intermediate allocation or
field reordering. The function signatures and error messages are
unchanged; the existing tests pass as-is.

### DIFF
--- a/internal/anthropic_proxy/translate.go
+++ b/internal/anthropic_proxy/translate.go
@@ -1,6 +1,7 @@
 package anthropicproxy
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 )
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
This `json.Compact` refactor is a safe simplification that preserves validation while avoiding the unnecessary `any` round-trip and key reordering.
---

PR #181, opened 2026-04-18
refactor(workflow): remove dead DispatchDedupStore.Seen method

eloylp:
## Why

DispatchDedupStore.Seen was never called anywhere -- not in production code, not in tests. A grep for Seen( across the entire internal/ tree returns no results. The method was superseded by SeesPendingOrCommitted (which also covers pending claims) when the TOCTOU-safe TryClaim/CommitClaim protocol was introduced.

Keeping a dead exported method adds noise and can mislead a reader into thinking it is part of an active protocol.

## What changed

- Deleted DispatchDedupStore.Seen (~10 lines).
- Updated four stale doc-comment references to the removed method name.

No behaviour change; the internal/ package boundary makes this safe without a deprecation cycle.

Generated with Claude Code

### DIFF
--- a/internal/workflow/dispatch.go
+++ b/internal/workflow/dispatch.go
@@ -64,7 +64,7 @@ func (c *dispatchCounters) snapshot() DispatchStats {
 // dispatchEntry records a dedup slot in the store.
 // committed indicates whether the slot is backed by a real enqueued event.
 // Pending (committed==false) entries block concurrent duplicate TryClaim calls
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Removing the unused `DispatchDedupStore.Seen` method and updating the related comments is a safe cleanup with no behavioral change.

---

PR #183, opened 2026-04-19
refactor(config): derive log-level error message from validLogLevels slice

eloylp:
## Why

`validateLogConfig` builds its error message by hardcoding the allowed log levels as a literal string:

```go
return fmt.Errorf("config: invalid log level %q (supported: trace, debug, info, warn, error, fatal, panic, disabled)", c.Daemon.Log.Level)
```

The `validLogLevels` slice six lines above is the authoritative source of that list. Adding a new level to the slice would not automatically update the error message — a silent maintenance trap.

## What changed

Replace the hardcoded string with `strings.Join(validLogLevels, ", ")`, matching the pattern already used by `validAIBackendNames`:

```go
return fmt.Errorf("config: invalid log level %q (supported: %s)", c.Daemon.Log.Level, strings.Join(validLogLevels, ", "))
```

Output is identical; no test changes required.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -592,7 +592,7 @@ var validLogLevels = []string{"trace", "debug", "info", "warn", "error", "fatal"
 func (c *Config) validateLogConfig() error {
 	if c.Daemon.Log.Level != "" {
 		if !slices.Contains(validLogLevels, c.Daemon.Log.Level) {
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Behavior-preserving config validation refactor that keeps the error text derived from the existing authoritative log-level list.

---

PR #177, opened 2026-04-18
refactor(autonomous): replace manual map copy with maps.Clone

eloylp:
## Why

`AgentStatuses` takes a read-lock on `lastRunsMu`, copies `s.lastRuns` into a local map, then releases the lock. The copy used an explicit `make` + `range` loop:

```go
runs := make(map[string]lastRunRecord, len(s.lastRuns))
for k, v := range s.lastRuns {
    runs[k] = v
}
```

`maps.Clone` (stdlib since Go 1.21) does exactly this in one line and makes the intent — shallow copy — obvious at a glance.

## Change

- Add `"maps"` import to `internal/autonomous/scheduler.go`
- Replace the 4-line copy loop with `runs := maps.Clone(s.lastRuns)`

No behaviour change; the resulting map is identical.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #199, opened 2026-04-20
refactor: replace make+copy slice cloning with slices.Clone

eloylp:
## Why

Two sites used the manual `make([]T, len(s)) + copy(dst, src)` idiom to clone a slice before sorting or snapshotting it. `slices.Clone` is the stdlib spelling for exactly this operation — it is shorter, immediately recognisable, and consistent with the `maps.Clone` call already used elsewhere in the codebase (e.g. #177).

## What changed

- `internal/ai/prompt.go` — roster copy before `slices.SortFunc`
- `internal/observe/observe.go` — `Dispatches` copy in `Edges()` snapshot; adds `slices` to the import block

No behaviour change. Both call sites guard against nil/empty inputs before reaching the cloning line.

### DIFF
--- a/internal/ai/prompt.go
+++ b/internal/ai/prompt.go
@@ -142,8 +142,7 @@ func renderRuntimeContext(ctx PromptContext) string {
 	if len(ctx.Roster) > 0 {
 		b.WriteString("\n## Available experts\n\n")
 		// Sort roster by name for deterministic output.
-		roster := make([]RosterEntry, len(ctx.Roster))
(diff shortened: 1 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The `slices.Clone` substitutions preserve the current behavior at both call sites while simplifying the copy intent and keeping the package boundaries unchanged.

---

PR #171, opened 2026-04-18
test(webhook): collapse 3 auth-rejection tests into table-driven

eloylp:
## Why

`TestHandleAgentsRunRejectsNoAuth`, `TestHandleAgentsRunRejectsWrongToken`, and `TestHandleAgentsRunReturnsForbiddenWhenNoAPIKeyConfigured` are structurally identical: each creates a server, fires a POST to `/agents/run` with a specific auth credential, and asserts one HTTP status code. The only differences are the `Authorization` header value and the expected status.

Collapsing them into a single `TestHandleAgentsRunAuthRejections` table makes the three covered scenarios visible at a glance and removes ~30 lines of boilerplate while preserving full coverage.

## Changes

- Replaced three flat test functions with one table-driven `TestHandleAgentsRunAuthRejections`
- Each of the three original scenarios becomes a named parallel subtest
- No behavioral change — all assertions are identical to the originals

## Test plan

- [ ] `go test ./internal/webhook/... -race` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #163, opened 2026-04-17
test(webhook): collapse 3 non-triggering-action tests into table-driven

eloylp:
## Why

`TestHandleIssueCommentNonCreatedIgnored`, `TestHandlePullRequestReviewNonSubmittedIgnored`, and `TestHandlePullRequestReviewCommentNonCreatedIgnored` are structurally identical:

1. Create a test server.
2. Fire a webhook with a non-triggering action for that event type.
3. Assert HTTP 202.
4. Assert no workflow event was enqueued.

The only differences are the event type, delivery ID, and body JSON — ideal table-driven inputs.

## What changed

The three separate test functions are replaced by a single `TestHandleNonTriggeringActionsIgnored` with a three-row table. Each row preserves the original event type, delivery ID, and body verbatim. Each subtest runs in parallel. No behavior change.

PR #201, opened 2026-04-20
test: convert bare loop in TestHandleAPIMemoryRejectsMultiSegmentTraversal to subtests

eloylp:
## Why

`TestHandleAPIMemoryRejectsMultiSegmentTraversal` iterated its three path-traversal vectors in a bare loop. Because the assertions used `t.Fatalf`, a failure on the first vector aborted the entire test — the second and third vectors never ran. This is the same pattern addressed in earlier refactors (#163, #171, etc.).

## What changed

- Added a `name` field to the anonymous struct.
- Wrapped the loop body in `t.Run(tc.name, func(t *testing.T) { t.Parallel(); … })`.

All three traversal vectors now run independently:
1. **repo segment walks above memDir** — `../../../../secret` in the repo slot
2. **agent segment walks above memDir** — `../../../../..` in the agent slot
3. **combined segments escape the root** — both slots together

The assertions are byte-for-byte identical; only the structure changed.

## Test

`go test ./internal/webhook/... -race -run TestHandleAPIMemoryRejectsMultiSegmentTraversal` — all three subtests pass.

### DIFF
--- a/internal/webhook/api_test.go
+++ b/internal/webhook/api_test.go
@@ -1159,28 +1159,32 @@ func TestHandleAPIMemoryRejectsMultiSegmentTraversal(t *testing.T) {
 	handler := srv.requireAPIKey(http.HandlerFunc(srv.handleAPIMemory))
 
 	cases := []struct {
+		name  string
 		agent string
 		repo  string
 	}{
 		// repo segment tries to walk above memDir
-		{agent: "coder", repo: "../../../../secret"},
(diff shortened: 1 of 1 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
This is a focused test-structure refactor: the subtests preserve the existing assertions, improve failure isolation, and the added `t.Parallel()` is safe here.

---

Issue #188, opened 2026-04-19
feat: SQLite as single runtime store — one flag, one file, everything else is UI


eloylp:
## Vision

The daemon's only startup parameter is a SQLite database path. Everything — agents, skills, prompts, repos, bindings, backends, dispatch config, HTTP config, log config, memory, traces, events — lives in that one file. YAML becomes an import/export format for sharing configurations, not a runtime dependency.

```bash
# That's it. Everything else is configured via the UI.
./agents --db /var/lib/agents/agents.db

# First setup (or sharing a config):
./agents --db agents.db --import config.yaml
```

### Why

The current YAML config is ~150 lines, requires SSH + text editing + daemon restart for every change. That's not a product. A product is: open the dashboard, click "Add agent", fill the form, save — it's live.

### What moves to SQLite

**Everything except the DB path itself.**

| Entity | Currently | In SQLite | Hot-reload? |
|---|---|---|---|
| Agents (name, backend, skills, prompt, flags) | YAML + prompt files | `agents` table (prompt inline as TEXT) | Yes — next run |
| Skills (name, prompt) | YAML + skill files | `skills` table (prompt inline) | Yes — next run |
| Repos + bindings | YAML | `repos` + `bindings` | Yes — cron re-registers |
| AI backends (command, args, env, timeout) | YAML | `backends` | Yes — next run |
| Dispatch config (depth, fanout, dedup) | YAML | `dispatch_config` | Yes — immediate |
| HTTP config (listen addr, timeouts, paths) | YAML | `http_config` | Read at boot from DB |
| Log config (level, format) | YAML | `log_config` | Read at boot from DB |
| Proxy config | YAML | `proxy_config` | Read at boot (server rebind) |
| Prompts | Files on disk (`prompts/`) | Inline in `agents` or `skills` TEXT columns | Yes — next run |
| Memory | Files on disk (`memory_dir/`) | `memory` table with TTL cleanup | Staging file per run (see below) |
| Traces | In-memory ring buffer | `traces` table | Persistent across restarts |
| Events | In-memory ring buffer | `events` table | Persistent across restarts |
| Dispatch history | In-memory | `dispatch_history` | Persistent |

**No more `prompt_file`.** Prompts are TEXT columns. YAML export writes them as `prompt: |` block scalars. Import parses them back.

### Memory staging mechanism

Agents write memory to a filesystem path via MCP tools (the CLI doesn't know about SQLite). Solution:

[...]

PR #204, opened 2026-04-20
feat: SQLite foundation (Phase 1) — --db and --import flags

eloylp:
## Summary

Implements Phase 1 of #188: a SQLite-backed config store as a first-class alternative to YAML, with zero changes to any downstream package.

- **`internal/store`** — new package: `Open` (create/migrate), `Import` (YAML→DB), `Load` (DB→`*config.Config`), `LoadAndValidate` (full startup path).
- **Schema** (`migrations/001_init.sql`): `config` (key/value for daemon settings), `backends`, `skills`, `agents`, `repos`, `bindings`. Arrays/maps stored as JSON text. Migrations are embedded and tracked via `schema_migrations`.
- **`config.FinishLoad`** exported: applies defaults + normalization + secret resolution + validation without re-running prompt-file I/O — used by the store after loading from DB.
- **`--db <path>`**: use the SQLite database instead of `--config`.
- **`--import <yaml>`** (requires `--db`): parse YAML → upsert into DB → load from DB. Prints row counts on success (`imported N backends, N skills, ...`).
- Resolved secrets (`WebhookSecret`, `APIKey`) are **never written** to the database; only env-var names are persisted and re-resolved at startup.
- Dockerfile bumped to `golang:1.25-alpine` (required by `modernc.org/sqlite ≥ 1.49`).
- `modernc.org/sqlite` chosen for pure-Go, CGO-free builds.

### Migration path for existing users

```bash
# Import once from existing YAML:
./agents --db agents.db --import config.yaml
# → "import: imported 2 backends, 6 skills, 11 agents, 1 repos, 14 bindings"

# All subsequent starts:
./agents --db agents.db
```

YAML path remains fully supported — no breaking change.

## Test plan

- [x] `go test ./...` — all 11 packages pass
- [x] `TestOpenAndMigrate` — open/create DB, idempotent second open
- [x] `TestImportLoad` — full round-trip: Import → Load, verifies every field class (daemon settings, backends with args/env, skills, agents with booleans and JSON arrays, repos, bindings with all three trigger types and enabled/disabled state)
- [x] `TestImportIsIdempotent` — double import does not fail
- [x] `TestCountFrom` — row counts after import
- [x] `TestLoadEmptyDatabase` — descriptive error on fresh DB


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/store/store.go:317` — `Import` is not idempotent for repo bindings. Every re-import appends new rows to `bindings` without deleting or upserting the existing rows for that repo, so the same YAML imported twice produces duplicate triggers and duplicate agent executions. `TestImportIsIdempotent` currently documents this behavior instead of catching it, which means the bug will keep shipping.
- **[medium]** `cmd/agents/main.go:194` — `--import` is documented as requiring `--db`, but the code silently ignores it when `--db` is empty and just falls back to `config.Load(configPath)`. That makes a mistyped invocation look successful while skipping the import entirely; this should fail fast with a clear CLI error.

### Summary
The SQLite foundation is close, but the import path still has correctness gaps that can duplicate automation and silently ignore an invalid flag combination.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `cmd/agents/main.go:211` / `internal/store/store.go:599` — the success message says `imported ...`, but the code prints whole-table counts after the transaction, not the number of rows touched by this import. On any non-empty database, especially with merge semantics, a second import can report stale rows as newly imported and hide whether the YAML actually changed anything.
- **[low]** `internal/store/store.go:460` / `internal/store/store.go:505` / `internal/store/store.go:538` — the DB read path drops all deterministic ordering for agents, repos, and bindings. YAML preserves author order, but the SQLite path now returns whatever row order the engine happens to choose, which makes `/api/config` output and future export/import diffs unstable. The current round-trip test also bakes in insertion order rather than asserting explicitly on sorted data.

### Summary
The DB foundation is workable, but the current import feedback is misleading on non-empty databases and the load path should make ordering guarantees explicit before this becomes user-visible drift.

---

eloylp:
All four findings from both review rounds have been addressed in commit 315574d:

**[high] Idempotent binding import**: `importRepos` now runs `DELETE FROM bindings WHERE repo=?` before re-inserting, so bindings are replace-all per repo. A second import of the same YAML produces exactly the same row count as the first. `TestImportIsIdempotent` now asserts `counts.Bindings == 3` after a double import (the comment documenting the duplicate-row behavior has been removed).

**[medium] --import without --db fails fast**: `loadConfig` now checks `importPath != "" && dbPath == ""` before the `dbPath == ""` branch and returns an error immediately. A mistyped invocation no longer silently falls back to loading the YAML config.

**[medium] Accurate import count**: Replaced `CountFrom(db)` in the success message with a direct count from the source config (`len(yamlCfg.Agents)`, etc.). The logged numbers now reflect exactly what was written regardless of any pre-existing rows in the database.

**[low] Deterministic load ordering**: `loadAgents` and `loadRepos` now use `ORDER BY name`; `loadBindingsForRepo` uses `ORDER BY id` (preserving YAML insertion order since bindings are delete-and-reinsert per import).

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/store/store.go:48` — `PRAGMA foreign_keys=ON` is only applied to the one SQLite connection that executes this statement. `*sql.DB` is a connection pool, and SQLite requires foreign-key enforcement to be enabled separately on every connection, so later transactions can still run with FK checks off. In practice that means this store can accept orphaned `bindings` rows or other invalid references depending on which pooled connection services the query.
- **[high]** `internal/store/store.go:245` / `internal/store/store.go:282` / `internal/store/store.go:298` — the import path still uses `INSERT OR REPLACE` for parent tables. In SQLite, `REPLACE` deletes the existing row before inserting the new one; once foreign keys are reliably enabled, re-importing a config that already has `bindings` will abort when it tries to replace the referenced `agents` or `repos` rows. This needs a real upsert (`ON CONFLICT DO UPDATE` or explicit `UPDATE`+`INSERT`) plus a regression test that exercises a second import with FK enforcement actually on.

### Summary
The revised import path fixed the earlier counting and ordering issues, but the current SQLite usage still leaves referential integrity unreliable and will break re-imports once foreign keys are enforced correctly.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The revised SQLite foundation addresses the earlier import and ordering issues; the remaining read/write path is coherent, validated on startup, and keeps the new storage concern contained in `internal/store`.

---

PR #207, opened 2026-04-20
feat(store,webhook): Phase 2 write API — CRUD endpoints + live config reload

eloylp:
## Summary

Implements Phase 2 of #188: a complete write API for managing fleet entities (agents, skills, backends, repos, bindings) at runtime over HTTP, backed by the Phase 1 SQLite store. The daemon's in-memory config view updates atomically after every write — no restart required.

**Based on** #204 (Phase 1 — must merge first).

### What changed

- **`internal/store`** — FK-safe `Put*` / `Delete*` helpers for all five entity types. `SetMaxOpenConns(1)` ensures `PRAGMA foreign_keys=ON` is honoured on every pooled connection.
- **`internal/webhook/server.go`** — `s.cfg` replaced by `atomic.Pointer[config.Config]` + `cfg()` accessor; `WithStore(db)` activates the write routes; `reloadConfig()` refreshes the live config after each mutation while preserving runtime secrets.
- **`internal/webhook/api_write.go`** — ten new endpoints (`PUT`/`DELETE` on agents, skills, backends, repos; `POST`/`DELETE` on bindings), all bearer-token gated, all reloading live config on success.
- **`cmd/agents/main.go`** — `loadConfig` now returns `(*config.Config, *sql.DB, error)`; the DB stays open for the daemon's lifetime.

### Write endpoints (only active with `--db`)

| Method | Path | Action |
|---|---|---|
| `PUT` | `/api/agents` | Create or update agent |
| `DELETE` | `/api/agents/{name}` | Delete agent |
| `PUT` | `/api/skills` | Create or update skill |
| `DELETE` | `/api/skills/{name}` | Delete skill |
| `PUT` | `/api/backends` | Create or update backend |
| `DELETE` | `/api/backends/{name}` | Delete backend |
| `PUT` | `/api/repos` | Create or update repo |
| `DELETE` | `/api/repos/{name}` | Delete repo (cascades bindings) |
| `POST` | `/api/repos/{name}/bindings` | Add binding; returns `{"id": N}` |
| `DELETE` | `/api/repos/{name}/bindings/{id}` | Delete binding by ID |

## Test plan

- [ ] `go test ./internal/store/... ./internal/webhook/... -count=1` — all new store and handler tests pass
- [ ] `go test ./... -count=1` — full suite green

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"flag"
 	"fmt"
@@ -53,9 +54,12 @@ func run() error {
 	runRepo := flag.String("repo", "", "repo to target when using --run-agent (e.g. owner/repo)")
 	flag.Parse()
 
-	cfg, err := loadConfig(*configPath, *dbPath, *importPath)
+	cfg, db, err := loadConfig(*configPath, *dbPath, *importPath)
 	if err != nil {
 		return err
+	}
+	if db != nil {
+		defer db.Close()
 	}
 
 	logger := logging.NewLogger(cfg.Daemon.Log)
@@ -125,6 +129,9 @@ func run() error {
 	server.WithUI(ui.FS)
 	server.WithObserve(obs)
 	server.WithRuntimeState(obs)
+	if db != nil {
+		server.WithStore(db)
+	}
 
 	group, groupCtx := errgroup.WithContext(ctx)
 	deliveryStore.Start(groupCtx)
@@ -188,28 +195,33 @@ func setupScheduler(cfg *config.Config, runners map[string]ai.Runner, logger zer
 //   - If importPath is also set, the YAML at importPath is parsed and written
 //     into the database before loading.
 //   - The config is then read from the database.
+//   - The returned *sql.DB is kept open for the lifetime of the daemon so that
+//     write API endpoints can use it. The caller is responsible for Close.
 //
-// When dbPath is empty, configPath is used (default behaviour).
-func loadConfig(configPath, dbPath, importPath string) (*config.Config, error) {
+// When dbPath is empty, configPath is used (default behaviour) and nil is
+// returned for the DB.
+func loadConfig(configPath, dbPath, importPath string) (*config.Config, *sql.DB, error) {
 	if importPath != "" && dbPath == "" {
-		return nil, fmt.Errorf("--import requires --db")
+		return nil, nil, fmt.Errorf("--import requires --db")
 	}
 	if dbPath == "" {
-		return config.Load(configPath)
+		cfg, err := config.Load(configPath)
+		return cfg, nil, err
 	}
 	db, err := store.Open(dbPath)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	defer db.Close()
 
 	if importPath != "" {
 		yamlCfg, err := config.Load(importPath)
 		if err != nil {
-			return nil, fmt.Errorf("import: load YAML: %w", err)
+			db.Close()
+			return nil, nil, fmt.Errorf("import: load YAML: %w", err)
 		}
 		if err := store.Import(db, yamlCfg); err != nil {
-			return nil, fmt.Errorf("import: write to database: %w", err)
+			db.Close()
+			return nil, nil, fmt.Errorf("import: write to database: %w", err)
 		}
 		// Count from the source config so the message reflects exactly what
 		// was written, not a potentially stale whole-table count.
@@ -222,7 +234,12 @@ func loadConfig(configPath, dbPath, importPath string) (*config.Config, error) {
 			len(yamlCfg.Agents), len(yamlCfg.Repos), nBindings)
 	}
 
-	return store.LoadAndValidate(db)
+	cfg, err := store.LoadAndValidate(db)
+	if err != nil {
+		db.Close()
+		return nil, nil, err
+	}
+	return cfg, db, nil
 }
 
 // schedulerStatusAdapter adapts *autonomous.Scheduler to webhook.StatusProvider,
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -131,6 +131,11 @@ func run() error {
 	server.WithRuntimeState(obs)
 	if db != nil {
 		server.WithStore(db)
+		// Propagate write-API config reloads to the engine and scheduler so that
+		// new repos, agents, and bindings take effect without a restart.
+		// Note: new cron bindings and backend changes require a restart.
+		server.WithConfigObserver(engine)
+		server.WithConfigObserver(scheduler)
 	}
 
 	group, groupCtx := errgroup.WithContext(ctx)
--- a/internal/autonomous/scheduler.go
+++ b/internal/autonomous/scheduler.go
@@ -8,6 +8,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/robfig/cron/v3"
@@ -71,7 +72,7 @@ type lastRunRecord struct {
 // Scheduler wires cron-triggered agent bindings from the config into the
 // robfig/cron engine.
 type Scheduler struct {
-	cfg          *config.Config
+	cfgPtr       atomic.Pointer[config.Config]
 	runners      map[string]ai.Runner
 	memories     *MemoryStore
 	cron         *cron.Cron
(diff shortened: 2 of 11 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/api_write.go:281` — `DELETE /api/repos/{name}/bindings/{id}` ignores the `{name}` path parameter and deletes by global binding ID only. Because binding IDs are repository-agnostic, a caller hitting `/api/repos/repo-a/bindings/42` can delete binding `42` that actually belongs to `repo-b`. The handler needs to scope the delete to both repo and id, and the tests should cover the cross-repo case.
- **[medium]** `internal/webhook/server.go:120` — `reloadConfig()` reloads with `store.Load()` and swaps the result directly into the live server config, bypassing `config.FinishLoad()` (`internal/config/config.go:260`). That means write-created config is never normalized or validated before becoming live. A mixed-case agent/skill/backend name written through the API can stop matching helpers like `AgentByName` (`internal/config/config.go:315`), and deleting a referenced skill/backend can leave the daemon with an internally inconsistent config that only fails later when prompt rendering or agent execution touches it (`internal/ai/prompt.go:64`).

### Summary
The write API is close, but binding deletion is not repo-scoped and live reload currently accepts unvalidated/un-normalized config.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `cmd/agents/main.go:67` — the new "live reload" only updates `webhook.Server`'s private config pointer. `setupRunners`, `setupScheduler`, and `workflow.NewEngine` all capture the startup config once and are never refreshed, so writes do not actually become live for execution: a newly added repo/agent/binding will show up in `/api/*` but webhook events still route through the stale `Engine.cfg`, cron jobs are never re-registered from the stale `Scheduler.cfg`, and backend edits never rebuild the `runners` map. That breaks the PR's core contract of "no restart required" and can leave operators believing a saved change is active when the daemon will continue using the old topology until restart.

### Summary
The read-side API view reloads, but the runtime execution path stays pinned to startup config, so the hot-reload guarantee is not met.

---

eloylp:
All findings from both review rounds have been addressed in commit e76ea39:

**[high] Repo-scoped binding deletion** (`api_write.go:281`): `store.DeleteBinding` now takes a `repo` parameter and scopes the DELETE to `WHERE id=? AND repo=?`. A mismatch returns `sql.ErrNoRows` so `handleDeleteBinding` returns 404 instead of silently doing nothing. `TestHandleDeleteBindingCrossRepo` (webhook) and `TestDeleteBindingCrossRepo` (store) both verify the cross-repo case.

**[medium] `config.FinishLoad` in `reloadConfig`** (`server.go`): `reloadConfig` now copies the live secrets into the raw-loaded config first (so `resolveSecrets` leaves them intact), then calls `config.FinishLoad` — running `applyDefaults`, `normalize`, and `validate` before the pointer is swapped.

**[high] Live config propagation to Engine and Scheduler** (`cmd/agents/main.go:67`): `Engine.cfg` and `Scheduler.cfg` are now `atomic.Pointer` fields accessed via a `cfg()` method. `Server` gains a `ConfigObserver` interface and a `WithConfigObserver` registration method. `reloadConfig` calls `o.UpdateConfig(cfg)` on every registered observer after updating the server's own pointer. `Dispatcher.agents` is protected by a `sync.RWMutex` and updated atomically via `UpdateAgents` from `Engine.UpdateConfig`. `main.go` registers both `engine` and `scheduler` as config observers when `--db` is in use.

Remaining documented limitation: new cron bindings and new AI backends added via the write API require a daemon restart (existing cron jobs pick up the new config on their next firing, but `NewScheduler` registers cron jobs once at startup and `setupRunners` builds the runner map at startup).

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/api_write.go:66` — Every mutation handler logs `reloadConfig()` failures and still returns success (`204`/`201`). That breaks the contract advertised in the PR description: a client can get a success response even though the live in-memory config was not updated. In the current implementation that means operators can save a change, immediately read stale data from `/api/*`, and only discover the broken persisted state on restart. Reload needs to be part of the request success path, or the write needs to be rolled back when the live refresh fails.
- **[medium]** `internal/webhook/api_write.go:189` — `putRepoRequest.Enabled` is a plain `bool`, so omitting `enabled` from `PUT /api/repos` is indistinguishable from explicitly sending `false`. The schema and existing config semantics default repos to enabled, but this handler flips a newly created repo to disabled unless every client remembers to send `"enabled": true`. A pointer field or explicit defaulting before `store.PutRepo` would keep the API aligned with the rest of the config model.

### Summary
The write API still reports success when the live reload fails, and repo creation currently defaults omitted `enabled` values to disabled.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/autonomous/scheduler.go:172` — `registerJobs` still passes a concrete `config.AgentDef` into `makeCronJob`, and that closure keeps reusing the startup snapshot on every tick. After a write-API reload, cron-triggered runs therefore keep using stale agent prompt/backend/AllowPRs data, and even deleted or disabled bindings can keep firing because the job never re-resolves the repo/binding/agent from `s.cfg()`. Manual `TriggerAgent` now sees the new config, but scheduled runs do not, so the advertised hot-reload guarantee is still broken for autonomous bindings.

### Summary
The webhook/runtime reload path is improved, but cron jobs still execute against startup-captured agent state instead of the live config.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond the blockers already raised in earlier review comments on this SHA range.

### Summary
I did not find an additional issue beyond the previously reported cron hot-reload gap, so the existing blocking feedback remains the decision driver.

---

eloylp:
All three findings from the latest review round have been addressed in commit 91fdce8.

**[high] reloadConfig failure returns 500** (`api_write.go`): Every mutation handler now returns `500 Internal Server Error` when `reloadConfig()` fails instead of silently logging the error and returning success. A failed reload means the in-memory view is stale, so the write path must be atomic from the client's perspective. A `testReloadHook` override was added to `Server` so the failure path can be covered in tests — `TestHandleMutationReloadFailureReturns500` verifies all mutation handlers (put agent, put skill, put repo, put binding) return 500 on reload failure.

**[medium] `putRepoRequest.Enabled` bool → `*bool`** (`api_write.go:189`): Changed from `bool` to `*bool`. Omitting `enabled` now defaults to `true` (matching the config model where repos are enabled by default). Explicit `false` still works. `TestHandlePutRepo` was extended to cover all three cases: explicit true, explicit false, and omitted (defaults to true).

**[high] Cron jobs re-resolve from live config on each firing** (`scheduler.go:172`): `makeCronJob` now takes `(repo, agentName, cronExpr string)` instead of a concrete `config.AgentDef`. Each tick re-resolves the repo definition, binding, and agent from `s.cfg()`. Disabled/removed repos and bindings skip silently; deleted agents log an error and skip. Three new scheduler tests cover the live-config cases: `TestSchedulerCronJobPicksUpLiveConfigOnEachFiring`, `TestSchedulerCronJobSkipsDisabledBindingAfterLiveReload`, and `TestSchedulerCronJobSkipsDisabledRepoAfterLiveReload`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/autonomous/scheduler.go:95` — `Scheduler.UpdateConfig` only swaps the config pointer; it never re-registers cron entries. As a result, `POST /api/repos/{name}/bindings` can persist a brand-new cron binding and return success, but that binding will never fire until the daemon restarts because `registerJobs()` only runs in `NewScheduler` (`internal/autonomous/scheduler.go:119`). Issue #188 explicitly calls out "Cron re-registration on binding change", and the PR description promises runtime write management with no restart, so this is still a functional gap in the hot-reload path.
- **[high]** `internal/webhook/api_write.go:289` — `handlePutBinding` inserts the new row before calling `reloadConfig()`, and on reload failure it now returns `500` without rolling the insert back. Because `store.PutBinding` is an append-only `INSERT` (`internal/store/store.go:791`), a client that retries the failed request creates duplicate bindings even though the first write already committed. The new 500-path tests cover the status code, but they do not cover this persisted-side-effect case; the handler needs transactional rollback or another idempotency guard for binding creation.

### Summary
The latest SHA fixes several earlier blockers, but cron binding hot-reload is still incomplete and binding creation is not atomic on reload failure.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond the blockers already raised in existing comments on this SHA.

### Summary
I did not find an additional issue beyond the already-documented blocking findings.

---

eloylp:
Closing in favour of #211, which is a more targeted implementation and supersedes this PR. The review feedback from this PR was incorporated into that work.

PR #210, opened 2026-04-20
fix(#203): memory via structured output — eliminate filesystem memory write-through

### LINKED ISSUE #203: ai: memory via structured output response (eliminate filesystem memory)
## Idea

Instead of agents writing memory to a filesystem path via MCP tools, memory flows through the structured output response contract:

### Before (current)
1. Daemon reads memory from file, injects into prompt as `Existing memory: ...`
2. Daemon provides `memoryPath` in the runtime context
3. Agent reads/writes the file during its tool loop via `Write` tool
4. Daemon re-reads the file after the run

### After
1. Daemon reads memory from SQLite (#188), injects into prompt as `Existing memory: ...`
2. No `memoryPath` — agent doesn't touch the filesystem for memory
3. Agent returns updated memory in `response.memory` (full overwrite)
4. Daemon writes `response.memory` back to SQLite

### Response schema change

```json
{
  "summary": "...",
  "artifacts": [...],
  "dispatch": [...],
  "memory": "## Active PRs\n- PR #133 @ abc123 APPROVE\n..."
}
```

`memory` is a string field (full content, full overwrite). Empty string = clear memory. The field is required in the schema so the agent always explicitly returns its memory state.

### What this eliminates

- `memory_dir` config field
- `memoryPath` in prompt runtime context
- `MemoryStore` file-based implementation (`internal/autonomous/memory.go`)
- Filesystem volume mount for memory in Docker
- File locking in `WithLock`
- Agent needing `Write` tool access to memory files
- The `memory/` directory entirely

### What this enables

- Memory stored in SQLite with TTL cleanup
- Memory visible + editable in the UI (#188)
- Memory version history (diff between runs)
- No filesystem side-effects from agent runs (cleaner sandbox model)

### Implementation

- Add `Memory string` to `ai.Response` struct
- Add `"memory"` field to `response-schema.json` (both inline and file)
- Update `internal/autonomous/scheduler.go` to read memory from SQLite before run, write `resp.Memory` back after run
- Update all agent prompts: replace "## Memory hygiene" filesystem instructions with "return your updated memory in the `memory` field of your response JSON"
- Remove `memory.go`, `memory_test.go`, `memoryPath` from `PromptContext`
- Update `config.go` to remove `memory_dir` validation

### Depends on

- #188 (SQLite as runtime store) — memory table must exist before this can land

### Prompt changes needed

Every prompt with "## Memory hygiene" section changes from:
> Record issue numbers attempted... Keep at most 30 lines of active state.

To:
> Return your full updated memory in the `memory` field of your JSON response. This replaces your entire previous memory. Keep at most 30 lines of active state. The daemon reads your memory before each run and provides it in the prompt; you return the updated version in your response.

eloylp:
## Summary

Closes #203

Agents now return their full updated memory in the `memory` field of their JSON response. The daemon reads memory before each autonomous run and writes `resp.Memory` back after — replacing the old write-through approach where agents used the `Write` tool to modify a `MEMORY.md` file directly.

- **`ai.Response.Memory string`** added; `response-schema.json` updated to require the field on every response.
- **`PromptContext.MemoryPath` removed** — replaced by `IsAutonomous bool` which gates the "Existing memory:" section. Event-driven runs no longer see a memory section at all.
- **`autonomous.MemoryBackend` interface** (`ReadMemory`/`WriteMemory`) replaces the old `MemoryStore.WithLock` closure. Two implementations:
  - `fileMemory` (YAML path) — per-agent, per-repo Markdown files under `memory_dir`, no file locking needed since the daemon now serializes all reads and writes.
  - `sqliteMemory` (`--db` path, wired in `cmd/agents/main.go`) — delegates to `store.ReadMemory` / `store.WriteMemory` backed by the `memory` table already created in migration `002_observability_and_memory.sql`.
- **`store.ReadMemory` / `store.WriteMemory`** added.
- **`loadConfig` in `main.go`** now returns the open `*sql.DB` when `--db` is used so the daemon keeps it alive for the memory backend. Memory dir watch is skipped on the SQLite path.
- **Agent prompts** (coder, pr-reviewer, refactorer, product-strategist) updated: "## Memory hygiene" now instructs agents to return updated memory in the `memory` response field.

## Test plan

- [ ] `go test ./... -count=1` passes (all 11 packages green)
- [ ] New tests: `TestFileMemory*` in `autonomous`, `TestReadWriteMemory*` in `store`
- [ ] Updated tests: `TestRenderAgentPromptWithMemory`, `TestRenderAgentPromptEmptyMemoryFormattedExplicitly`, `TestRenderAgentPromptMemoryOmittedForEventDrivenRuns`


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `prompts/arch-reviewer.md:23` — `response-schema.json` now requires a top-level `memory` field, but several unchanged prompt JSON templates still instruct outputs with only `summary/artifacts/dispatch` (also in `prompts/dx-reviewer.md:23`, `prompts/ops-reviewer.md:24`, `prompts/sec-reviewer.md:22`, `prompts/test-reviewer.md:21`, `prompts/codebase-scout-code.md:31`, `prompts/codebase-scout-issues.md:24`). This creates an internal contract conflict that can cause schema-validation failures or dropped runs when those agents follow their prompt templates.

### Summary
The structured-memory change is close, but the required response schema and multiple agent prompt templates are currently inconsistent.

---

eloylp:
Addressed. Added `"memory": ""` to the JSON example block and a rule bullet (`- \`memory\` — this agent is stateless; always return \`""\``) in all seven stateless agent prompt templates: `arch-reviewer`, `dx-reviewer`, `ops-reviewer`, `sec-reviewer`, `test-reviewer`, `codebase-scout-code`, and `codebase-scout-issues`. This aligns all templates with the `response-schema.json` contract.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `cmd/agents/main.go:136` — when `--db` is enabled the daemon stops watching `memory_dir`, but `/api/memory/{agent}/{repo}` still reads only `memory_dir/MEMORY.md` (`internal/webhook/api.go:548`). In SQLite mode the new structured-memory path becomes invisible to the dashboard: the memory page and SSE stream will return stale data or 404 even though autonomous runs are successfully persisting `resp.Memory` to the database.

### Summary
The structured-memory write path looks good, but the observability/UI read path still points at files and regresses memory visibility in SQLite mode.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/autonomous/scheduler.go:366` — memory persistence is now treated as a best-effort side effect, but this PR also makes `response.memory` the only source of truth for autonomous memory. If `WriteMemory` fails here, the run is still logged as successful and the dedup mark is finalized, so agents like `pr-reviewer` can silently lose their updated state and re-review the same SHA on the next pass. Before this change, the write happened inside `WithLock` and failed the run. This needs to stay transactional from the scheduler’s point of view.

### Summary
The structured-memory design is sound, but dropping memory-write failures on the floor breaks the contract that autonomous state was actually persisted.

---

eloylp:
Addressed all three reviewer findings in commit 5b3294e.

**[medium] SQLite memory invisible in API/UI** (`api.go`): Added a `MemoryReader` interface to the `webhook` package. `Server.WithMemoryReader` is called in `run()` when `--db` is active, wiring the SQLite backend to `handleAPIMemory`. In SQLite mode the handler reads directly from the DB via the injected interface instead of the filesystem. Test: `TestHandleAPIMemorySQLiteMode`.

**[medium] SSE stream silent in SQLite mode** (`main.go`, `observe.go`): `WatchMemoryDir` is skipped in `--db` mode. `sqliteMemory.WriteMemory` now calls an optional `notifyFn` after each successful write; `run()` sets this to `obs.PublishMemoryChange`, which emits a `MemoryChangeEvent` to `MemorySSE`. The UI SSE stream now fires on every autonomous memory update in SQLite mode.

**[medium] Memory write failure non-fatal** (`scheduler.go`): `WriteMemory` error is now returned from `executeAgentRun` so the run is recorded as "error" and operators see the failure. The dedup mark is still finalised first (the agent run itself completed work and must not be re-triggered spuriously). Test: `TestSchedulerWriteMemoryFailureFailsRun`.

Also: `sqliteMemory` now normalises agent/repo keys with `ai.NormalizeToken` before storage, matching the file-backend layout so `/api/memory/{agent}/{repo}` URL segments (already in normalised `owner_repo` form) resolve correctly.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/webhook/api.go:571` — SQLite mode treats `content == ""` as "not found" and returns `404`, but this PR also defines empty memory as a valid persisted state (`response.memory = ""` clears memory and `store.WriteMemory` still upserts that value). That makes `/api/memory/{agent}/{repo}` semantically inconsistent with file mode (empty file returns `200`) and hides intentionally-cleared memory as missing data. Return `200` with an empty body for existing empty memory, and reserve `404` for true absence only.

### Summary
Structured memory is close, but the SQLite memory read path currently misclassifies valid empty memory as missing.

---

eloylp:
Addressed in commit 70519f9.

**[medium] Empty memory returns 404 instead of 200** (`api.go`, `store.go`, `main.go`): `store.ReadMemory` now returns `(string, bool, error)` where the bool indicates row existence, distinguishing a missing record (found=false) from an existing empty-content record (found=true, content=""). A new `sqliteWebhookReader` wrapper in `main.go` returns the `webhook.ErrMemoryNotFound` sentinel when found=false; `handleAPIMemory` checks `errors.Is(err, ErrMemoryNotFound)` for 404 and falls through to a 200 for all other outcomes including empty content. The `sqliteMemory` backend (used by the scheduler) ignores the found bool and still returns ("", nil) for missing rows — the scheduler's empty-memory semantics are unchanged. Test: updated `stubMemoryReader` to use map presence rather than empty-string for not-found; added "existing empty memory returns 200" table entry.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The structured-memory changes are internally consistent at this SHA; the SQLite and file-backed paths now preserve the intended memory semantics and observability behavior.

---

eloylp:
Addressed the reviewer's finding in commit 7203b92.

**[high] Memory write failure must roll back the dedup mark** (`scheduler.go`): When `WriteMemory` fails after a successful agent run, the previous code called `FinalizeAutonomousRun`, which kept the cron dedup mark and suppressed autonomous-context dispatches for the full dedup window. This meant an operator or another agent could not immediately re-dispatch the failing agent to recover lost memory state — they'd have to wait out the full window.

Changed to call `RollbackAutonomousRun` on `WriteMemory` failure, treating the memory write as part of the transactional run. The dedup slot is cleared so the agent can be re-dispatched immediately (or will re-run on its next cron tick with a fresh memory read attempt).

Updated `TestSchedulerWriteMemoryFailureFailsRun` to verify the new behaviour: after a write-memory failure, a permitted originator agent (`notifier` with `CanDispatch: ["reviewer"]`) can successfully dispatch `reviewer` again — the dispatch is no longer suppressed by the stale cron mark.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The dedup rollback fix closes the remaining structured-memory failure path cleanly; no new issues found at this SHA.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/autonomous/scheduler.go:422` — the new memory-write failure path now returns an error from `executeAgentRun`, but the trace span is recorded earlier and only keys off `runner.Run` failure. If `WriteMemory` fails at line 454, observability still persists a `success` span with the agent summary/artifact count even though the overall autonomous run failed and the dedup mark was rolled back. That leaves the traces view contradicting the scheduler/error status exactly in the scenario this PR is trying to surface to operators.

### Summary
The transactional rollback fix is correct, but the autonomous trace is still marked successful when post-run memory persistence fails.

---

eloylp:
Addressed in commit 1dd78df.

**[medium] Trace span must reflect WriteMemory failure** (`scheduler.go`): Previously `RecordSpan` was called immediately after `runner.Run`, so a subsequent `WriteMemory` failure left the observability layer with a `"success"` span while the scheduler returned an error and rolled back the dedup mark — contradictory state for operators watching `/api/traces`.

Moved the span recording into a `recordTrace` closure that is called at each exit point: runner error, memory-write error, and success path. The closure is always invoked exactly once per run, so the span status accurately reflects the final outcome.

Added `stubTraceRecorder` in the test file and extended `TestSchedulerWriteMemoryFailureFailsRun` to assert that a `WriteMemory` failure produces an `"error"` span whose message contains the underlying write error.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/autonomous/scheduler.go:466` — `recordTrace("success", "")` now runs before the post-run `ProcessDispatches` step, but `executeAgentRun` still returns an error if that enqueue fails at line 494. This leaves `/api/traces` showing a successful autonomous span for runs that the scheduler reports as failed, which is the same observability contradiction you just fixed for `WriteMemory` errors. Move success-span recording to the final successful exit path, and record an error span for the dispatch failure branch as well. A regression test alongside `TestSchedulerDispatchEnqueueFailurePropagates` would lock this down.

### Summary
The memory-write trace fix is correct, but the autonomous span still reports `success` when post-run dispatching fails.

---

eloylp:
Addressed in commit f707adb.

**[medium] Dispatch failure must record an error span** (`scheduler.go`): `recordTrace("success", "")` was called before `ProcessDispatches`, so a dispatch enqueue failure left `/api/traces` showing a successful autonomous span while the scheduler returned an error — the same observability contradiction fixed earlier for `WriteMemory` failures.

Moved `recordTrace("success", "")` to the final exit path (after the dispatch step succeeds) and added `recordTrace("error", dispErr.Error())` in the dispatch failure branch, so the span status accurately reflects the overall run outcome at every exit point.

Added `TestSchedulerDispatchEnqueueFailureRecordsErrorSpan` alongside the existing `TestSchedulerDispatchEnqueueFailurePropagates` to assert that a dispatch failure produces an `"error"` span whose message contains `"dispatch"`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/setup/prompt.md:160` — the embedded `agents setup` prompt still scaffolds autonomous agents with `Memory file: {{.MemoryPath}}` and the old file-based memory model. This PR removes `PromptContext.MemoryPath` and makes `response.memory` the only supported memory contract, so fresh setups generated from this template will emit stale agent instructions that no longer match the daemon/runtime schema. Update the setup template alongside the runtime prompts so new installs inherit the structured-memory flow too.

### Summary
The runtime path is aligned, but the setup wizard still seeds new configurations with the retired filesystem-memory contract.

---

eloylp:
Addressed in commit 8ebf3c1.

**[medium] Setup template still used the retired file-based memory contract** (`internal/setup/prompt.md`): Removed `Memory file: {{.MemoryPath}}` and `Existing memory:\n{{.Memory}}` from the autonomous agent prompt example in the setup wizard. The daemon now auto-injects existing memory into the `## Runtime context` section when `IsAutonomous` is true, and agents persist memory by returning a `"memory"` field in their JSON response. The example now documents the correct structured-memory contract so configurations generated by the setup wizard match the runtime schema.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/setup/prompt.md:160` — the structured-memory update removed `{{template "agent_guidance" .}}` from the autonomous prompt example, so fresh setups generated from this wizard will stop seeding autonomous agents with the shared reviewer guidance that the PR-review template still includes. This regresses newly scaffolded configs even though the runtime contract is otherwise updated; keep the new memory instructions, but preserve the common guidance block as well.

### Summary
The runtime path is aligned, but the setup wizard now drops the shared guidance block from generated autonomous prompts.

---

eloylp:
Addressed in commit 50f0463.

**[medium] Setup template dropped agent_guidance block from autonomous prompts** (`internal/setup/prompt.md`): Restored `{{template "agent_guidance" .}}` in the autonomous prompt example between the new memory-injection line and the `Return` instruction. The fix keeps all three elements: the structured-memory injection sentence, the shared reviewer guidance block, and the updated `"memory"` return-field instruction.

eloylp:
@pr-reviewer please review

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/autonomous/scheduler.go:421` — `spanEnd := time.Now()` is still captured immediately after `runner.Run`, but `recordTrace(...)` is now deferred until after `WriteMemory` and `ProcessDispatches`. The span status now reflects the final run outcome, but the recorded duration still excludes any time spent in those post-run steps. A slow SQLite write or dispatch enqueue will therefore emit an `error` span whose timing claims the run finished earlier than it actually did, which keeps `/api/traces` internally inconsistent. Capture `time.Now()` inside `recordTrace` (or pass the final end time per exit path) and add a regression test with a deliberately delayed memory/dispatch path.

### Summary
The structured-memory flow is in good shape, but autonomous trace durations are still recorded too early to match the final run outcome.

---

eloylp:
Addressed in commit e747679.

**[medium] spanEnd captured too early** (`scheduler.go`): Removed `spanEnd := time.Now()` from the line immediately following `runner.Run` and moved the capture inside the `recordTrace` closure. Now `spanEnd` is set at the moment the trace is actually recorded — after all post-run steps (WriteMemory, ProcessDispatches) complete — so the span duration in `/api/traces` matches the final outcome including time spent in those steps.

Added `TestSchedulerSpanEndCapturedAfterPostRunSteps`: uses a `slowMemory` stub that sleeps 50 ms during WriteMemory, then asserts that `spans[0].spanEnd.Sub(before) >= 50ms`, confirming the span end time was recorded after the slow write completed. Updated `stubSpan` and `stubTraceRecorder.RecordSpan` to capture the `spanEnd` time so this assertion is possible.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/webhook/api.go:568` — the new SQLite `/api/memory` path no longer sets `X-Memory-Mtime`, but the dashboard still reads that header to show the last-modified timestamp (`internal/ui/src/app/memory/page.tsx:53`). `store.WriteMemory` is already maintaining `updated_at`, so this regresses observability in `--db` mode by making SQLite-backed memory look untimestamped even though the data exists. Please carry the timestamp through the reader/API path so file and SQLite mode stay semantically aligned.

### Summary
The structured-memory flow is close, but SQLite mode still drops the memory timestamp that the UI depends on.

---

eloylp:
Addressed in commit 3a1e0da.

**[medium] X-Memory-Mtime missing in SQLite mode** (`store.go`, `main.go`, `server.go`, `api.go`): `store.ReadMemory` now selects `updated_at` alongside `content` and returns it as a `time.Time` (parsing the RFC3339 format that the modernc.org/sqlite driver returns, with a bare `YYYY-MM-DD HH:MM:SS` fallback). The `webhook.MemoryReader` interface signature changed to `(string, time.Time, error)`, `sqliteWebhookReader.ReadMemory` passes the timestamp through, and `handleAPIMemory` sets `X-Memory-Mtime` from it when non-zero. File and SQLite mode are now semantically aligned on the memory timestamp header.

Tests: `stubMemoryReader` grows an optional `mtimes` map; two new cases in `TestHandleAPIMemorySQLiteMode` verify that a non-zero timestamp sets the header and a zero timestamp omits it. `TestReadWriteMemory` asserts the returned mtime is non-zero after a write.

eloylp:
@pr-reviewer please review

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/autonomous/scheduler.go:387` — memory updates are now a plain read-before / write-after pair around `runner.Run`, but there is no longer any per-`(agent,repo)` critical section spanning the whole run. That reintroduces lost-update races for the new full-overwrite `response.memory` contract: two overlapping autonomous runs for the same agent/repo can both read the same old memory and whichever finishes last silently clobbers the other run's state. `TryMarkAutonomousRun` does not protect against this, because the cron mark intentionally does not block repeated cron/manual runs (`internal/workflow/dispatch.go:606`). The old `MemoryStore.WithLock` serialized the entire run and avoided exactly this class of corruption. This needs an equivalent scheduler-level lock/transaction (and a concurrency regression test) before structured memory is safe.

### Summary
The latest fixes close the SQLite observability gaps, but the new read/write memory flow still allows overlapping autonomous runs to silently lose memory state.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond the still-open scheduler memory lost-update race already identified earlier at `internal/autonomous/scheduler.go:387`.

### Summary
The latest SQLite timestamp fix looks correct; no additional issues found at this SHA, but the unresolved overlapping-run memory race still blocks approval.

---

eloylp:
Addressed in commit 669506d.

**[high] Lost-update race on overlapping autonomous runs** (`scheduler.go`): Added `runLocks` — a `sync.Map` of per-key `*sync.Mutex` values keyed by ``"${agent}\x00${repo}"`` — to `Scheduler`. In `executeAgentRun`, after backend/runner resolution and before `ReadMemory`, the lock is acquired; `defer s.runLock.release(runKey)` releases it after `WriteMemory` (and any `FinalizeAutonomousRun`) complete.

This serialises the entire read/run/write sequence per (agent, repo) pair so that two overlapping runs (cron + manual trigger, two manual triggers, or any run pair when no dispatcher is configured) cannot both read stale memory and then clobber each other's write.

The lock is acquired after the `TryMarkAutonomousRun` check (which remains responsible for dispatch dedup), and after the backend/runner sanity checks (which are fast and cannot block). Slow work — `runner.Run`, `WriteMemory`, `ProcessDispatches` — all falls inside the critical section.

Test: `TestSchedulerMemoryRunsSerializedPerAgentRepo` uses a `firstCallBlockingRunner` (blocks first Run until signalled) and a `readTrackingMemory` (records seen content per ReadMemory call). The first run holds the lock while inside `runner.Run`; the second run parks on `runLock.acquire`; the first run is unblocked, writes "first-run", releases the lock; the second run then reads "first-run" — proving the second read always happens after the first write.

@pr-reviewer please review

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The structured-memory path is consistent at `669506d`: autonomous memory now round-trips cleanly through the response schema, SQLite/file observability stay aligned, and the per-(agent,repo) lock closes the lost-update race without breaking the scheduler contract.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/memory/page.tsx:78` — the memory dashboard still describes the backing store as `memory_dir` / `MEMORY.md` files (`"Read-only view of memory_dir"`, `"Select a memory file..."`, `"Memory file not found..."`, and the `{agent}/{repo}/MEMORY.md` banner). After this PR, `--db` deployments serve memory from SQLite, so the UI now presents stale storage semantics on the exact path this change is extending. Please switch this copy to store-agnostic “memory entry/content” wording so SQLite mode does not mislead operators.

### Summary
The runtime/storage paths look aligned now, but the memory UI still exposes the retired file-based model.

---

eloylp:
Addressed in commit d1a0cfd.

**[medium] Memory UI exposes retired file-based storage semantics** (`internal/ui/src/app/memory/page.tsx`): Replaced all four file-based references with store-agnostic copy:
- `"Read-only view of memory_dir"` → `"Read-only view of agent memory"`
- `"Select a memory file to view its contents."` → `"Select a memory entry to view its contents."`
- `"Memory file not found. The agent may not have written any memory yet."` → `"Memory not found. The agent may not have written any memory yet."`
- `{file.agent}/{file.repoKey}/MEMORY.md` banner → `{file.agent}/{file.repoKey}`

The UI now uses neutral terminology that is accurate for both file and SQLite backends.

@pr-reviewer please review

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/dist/memory/index.html:19` — the source copy in `internal/ui/src/app/memory/page.tsx` was updated, but the embedded Next export under `internal/ui/dist/` was not rebuilt. `internal/ui/embed.go:12` serves that committed `dist/` tree at runtime, and the shipped memory page still renders the old file-based strings (`"Read-only view of memory_dir"`, `"Select a memory file..."`, `.../MEMORY.md`). As a result, the bug this commit claims to fix is still present in the actual UI artifact. Rebuild/export the UI bundle and commit the updated `dist` assets alongside the source change.

### Summary
The runtime code is aligned, but the embedded UI bundle is stale, so the served memory dashboard still exposes the retired file-based model.

---

eloylp:
Addressed in commit d854b95.

**[medium] Embedded UI bundle not rebuilt** (`internal/ui/dist/`): Ran `npm ci && npm run build` in `internal/ui/` and committed the regenerated export. The compiled `dist/memory/index.html` now embeds the store-agnostic copy from d1a0cfd — "Read-only view of agent memory", "Select a memory entry to view its contents.", "Memory not found. The agent may not have written any memory yet.", and the `{agent}/{repo}` banner without the `/MEMORY.md` suffix. Old chunk/manifest hashes were rotated out; new ones committed alongside.

@pr-reviewer please review

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The rebuilt embedded UI bundle now matches the source copy update, so the served memory dashboard no longer exposes the retired file-backed wording.

---

PR #223, opened 2026-04-21
feat(#222): Phase 3 — CRUD editors for agents, skills, repos, backends + YAML import/export

### LINKED ISSUE #222: ui: Phase 3 — CRUD editors for agents, skills, repos, backends + import/export YAML
## Goal

Extend the read-only observability UI into a full fleet management dashboard. Operators create, edit, and delete agents, skills, repos, bindings, and backends from the browser — changes take effect immediately via the Phase 2 write API (#211).

## Depends on

- #204 (Phase 1 SQLite) ✅ merged
- #211 (Phase 2 write API) ✅ merged

### Agent editor (extend Fleet page)
- "Create agent" button → form: name, backend (dropdown from /api/backends), skills (multi-select from /api/skills), prompt (inline markdown editor), allow_prs toggle, allow_dispatch toggle, can_dispatch multi-select, description.
- Click existing agent card → edit mode with same form, pre-filled.
- Delete with confirmation modal.
- Save → PUT /api/agents/{name} → hot-reload, card updates.

### Skill editor (new page or section)
- List all skills with inline prompt preview.
- Create / edit: name + prompt (markdown editor).
- Delete with confirmation.

### Repo + binding wiring (new page or section)
- List repos. Add repo (name, enabled toggle).
- Per repo: list bindings. Add binding: pick agent, pick trigger type (labels/events/cron), fill fields.
- Delete binding.

### Backend editor (extend Config page)
- List backends. Edit: command, args, env map, timeout, max_prompt_chars.
- Create / delete backend.

### Import / Export YAML
- "Export YAML" button → downloads a complete config.yaml generated from current DB state.
- "Import YAML" button → file upload, parse, upsert into DB. Two modes: merge (default) or replace.

## Implementation notes

- All CRUD goes through the existing /api/* write endpoints from #211.
- Use the same blueprint light theme and Card/Modal components.
- Forms should validate client-side (required fields, valid cron expression format) before POST.
- Prompt editor: use a <textarea> with monospace font for v1. A proper code editor (Monaco/CodeMirror) is a follow-up.
- After any successful mutation: refetch the relevant data and update the UI. No optimistic updates in v1.

## Out of scope

- Drag-to-wire dispatch in graph (Phase 4)
- Prompt version history with diff view (Phase 4)
- Cron visual builder (Phase 4)

eloylp:
## Summary

- **Agent CRUD** on Fleet page: "Create agent" button + Edit/Delete per card via modal forms, all wired to `/api/store/agents`
- **Skills page** (`/skills/`): list, create, edit, delete skills with inline prompt preview
- **Repos page** (`/repos/`): list, create, edit, delete repos with full binding editor (labels/events/cron per binding)
- **Config page** extended with tabs: existing Inspector, new Backends CRUD editor, new Import/Export YAML tab
- **YAML export** (`GET /api/store/export`): downloads `config-export.yaml` with all four CRUD-mutable sections
- **YAML import** (`POST /api/store/import`): upserts agents, skills, repos, backends from uploaded YAML; returns import summary
- **Modal component**: shared overlay used by all confirmation dialogs and editor forms
- NavBar updated with Skills and Repos links

All write endpoints require the API key. Export also requires it since backends may hold env secrets.

## Test plan

- [ ] `TestStoreExportReturnsYAML` — verifies YAML contains seeded agent/repo
- [ ] `TestStoreImportRoundTrip` — verifies imported entities appear in subsequent list
- [ ] `TestStoreExportNotRegisteredWithoutStore` — verifies 404 without DB
- [ ] Manual: Fleet page shows Edit/Delete buttons per agent; modal opens pre-filled; Save POSTs and refreshes
- [ ] Manual: Skills/Repos pages load, create, edit, delete work end-to-end
- [ ] Manual: Config → Backends tab lists backends from store, CRUD works
- [ ] Manual: Export downloads YAML; Import uploads and reports counts

Closes #222


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/page.tsx:323` — The new dashboard mutations call `/api/store/*` with plain same-origin `fetch()` requests, but those routes are all wrapped in `requireAPIKey` (`internal/webhook/server.go:231-249`). The server already documents that the embedded UI cannot attach a Bearer token (`internal/webhook/server.go:214-218`). In any deployment with `daemon.http.api_key` set, create/edit/delete/import/export all fail with `401`, so the Phase 3 UI is unusable in the secured configuration.
- **[high]** `internal/webhook/crud.go:681` — `/api/store/import` is not atomic. It loops over agents/skills/repos/backends and calls `store.Upsert*` one by one, and each helper opens and commits its own transaction (`internal/store/crud.go:112`, `171`, `226`, `316`). If a later entity fails validation, the handler returns an error after earlier writes are already committed, leaving the store partially imported with no rollback.
- **[medium]** `internal/ui/src/app/page.tsx:131` — The agent editor hardcodes the backend dropdown to `auto`, `claude`, and `codex`. This PR also adds backend CRUD, but any newly created backend can never be selected for an agent from the UI. The feature needs to source backend options dynamically from the store/config API.

### Summary
The CRUD surface is close, but it currently breaks under API-key-protected deployments and can leave the store in a partially imported state.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/config/page.tsx:375` — Editing a backend from the Backends tab destroys its configured secrets. The list API intentionally redacts every env value as `"[redacted]"` (`internal/webhook/crud.go:71`), but `openEdit` seeds the form directly from that redacted payload and `saveBackend` posts it back unchanged (`internal/ui/src/app/config/page.tsx:221`). Any save after a no-op edit overwrites the real `env` entries in SQLite with the literal placeholder string, so the backend breaks the next time it needs those credentials.
- **[medium]** `internal/ui/src/app/page.tsx:127` — The agent edit modal leaves the name field writable even though saves always go through the collection upsert endpoint (`internal/ui/src/app/page.tsx:323`). Changing the name while editing does not rename the existing agent; it creates a second agent and leaves repo bindings / `can_dispatch` references pointing at the old one. The UI presents this as an edit flow, but the persisted result is a silent duplicate.

### Summary
There are still correctness gaps in the CRUD UI: backend edits can clobber stored secrets, and agent “rename” edits create duplicate agents instead of updating the original.

---

eloylp:
Thanks for the thorough review. All three findings are addressed in the follow-up commit (3c6fe86):

**[high] UI mutation calls fail with 401 in secured deployments**
Added `internal/ui/src/lib/apiKey.ts` with `getApiKey` / `setApiKey` / `authHeaders` helpers backed by `localStorage`. Every mutating `fetch()` call (POST/DELETE for agents, skills, repos, backends; the export GET; and the import POST) now merges `Authorization: Bearer <key>` via `authHeaders()` when a key is stored. The Config page gained a password-type "API Key" input above the tabs — users configure it once and it persists across reloads.

**[high] `/api/store/import` is not atomic**
Added `store.ImportAll` in `internal/store/crud.go` — a single `db.Begin()` transaction that normalizes all entities, calls the existing `importAgents` / `importSkills` / `importRepos` / `importBackends` helpers, runs `validateFleet`, then commits. `handleStoreImport` now calls this one function instead of looping through individual `Upsert*` calls, so any validation failure rolls back the entire import.

**[medium] Agent editor hardcodes backend dropdown**
`AgentForm` now accepts a `backends string[]` prop. `FleetPage` fetches `/api/store/backends` on mount and passes the result as backend names; the dropdown always includes `auto` first, then any store-configured backends. When the store endpoint is unavailable (daemon started without `--db`), the error is swallowed and the dropdown shows only `auto`.

All Go tests pass; UI rebuilt and dist updated.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/config/page.tsx:393` — This is still overwriting stored backend secrets on edit. `/api/store/backends` deliberately redacts every env value to `"[redacted]"` (`internal/webhook/crud.go:71-99`), but the edit flow seeds the form directly from that redacted payload via `setSelected(b)` and `saveBackend` later posts `form.env` back unchanged (`internal/ui/src/app/config/page.tsx:221-229`). A no-op save therefore replaces the real secrets in SQLite with the placeholder string and breaks the backend.
- **[medium]** `internal/ui/src/app/page.tsx:132` — The agent edit modal still allows renaming, but saves always go through the collection upsert endpoint (`internal/ui/src/app/page.tsx:336-340`). Changing `form.name` during an edit creates a second agent instead of renaming the existing one, leaving repo bindings and `can_dispatch` references attached to the old name while the UI presents the action as an in-place edit.

### Summary
The follow-up commit fixed the API-key, atomic-import, and dynamic-backend issues, but two previously reported correctness bugs are still present in the current head.

---

eloylp:
Fixed in ecae80f.

**[high] Editing a backend destroys stored secrets**

Added `restoreRedactedEnv` helper in `internal/webhook/crud.go`. Before `UpsertBackend` is called, the handler now checks whether any `env` values are the `"[redacted]"` sentinel. If so, it reads the currently stored backend and substitutes the real value for each such entry. Keys whose sentinel has no stored counterpart (e.g. a `"[redacted]"` value that was never actually stored) are dropped rather than written.

The fix is purely server-side — no UI changes needed, and it correctly handles:
- Edit without touching secrets → real values preserved
- Edit that removes an env key row → key not present in POST body, so it's dropped from the upsert (existing behaviour)
- New backend that somehow contains the sentinel → entry is dropped

`TestStoreCRUDBackendPostPreservesRedactedEnv` was added to cover the failure case: it creates a backend with a real secret, POSTs the form back with the `"[redacted]"` sentinel, and reads the store directly (unredacted) to verify the original value survived.

eloylp:
@pr-reviewer — please re-review. The last REQUEST_CHANGES finding (backend secrets overwritten on edit) has been addressed in commit ecae80f:

- Added `restoreRedactedEnv` helper in `internal/webhook/crud.go` (lines 147–185): before `UpsertBackend` is called, any `"[redacted]"` sentinel in the incoming `env` map is replaced with the real stored value; if the key has no stored counterpart the sentinel is dropped rather than written.
- Tests added in `internal/webhook/crud_test.go` covering the restore path.
- The earlier auth-header finding was addressed in 3c6fe86 (`internal/ui/src/lib/apiKey.ts`).

All CI checks pass.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:100` — The create-agent flow still seeds `emptyForm.backend` with `"claude"`, while the dropdown options are now dynamic (`internal/ui/src/app/page.tsx:125-126`). If an installation only has custom backends or relies on `auto`, opening the modal and saving without touching the select still POSTs `backend: "claude"`, which the server rejects as `unknown backend "claude"`. The default needs to come from the available options (or be left empty so the server normalizes it to `auto`).
- **[medium]** `internal/webhook/crud.go:701` — The linked issue explicitly requires two import modes, merge and replace, but this PR only implements merge/upsert semantics. `handleStoreImport` always calls `store.ImportAll`, which only inserts/replaces records (`internal/store/crud.go:335`), and the UI text hardcodes merge-only behavior (`internal/ui/src/app/config/page.tsx:427`). There is no way to prune agents/skills/repos/backends that were removed from the YAML, so an export-edit-import round-trip cannot produce an exact replacement of the store state.

### Summary
The secret-preservation fix is in place, but the current head still has one create-agent regression and is missing the replace import mode promised by #222.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/webhook/crud.go:704` — `#222` explicitly calls for YAML import to support both merge and replace modes, but the handler only has a single path and always calls merge-only `store.ImportAll`. Any entities omitted from the uploaded file are left behind in the database, so operators still cannot do a true replace import.

### Summary
The CRUD/editor work is substantial, but the linked issue is not complete yet because import still lacks the promised replace mode.

---

eloylp:
Fixed in 066a116.

**[medium] Create-agent form defaults to non-existent backend**

Changed `emptyForm.backend` from `'claude'` to `'auto'` (`internal/ui/src/app/page.tsx:101`). `auto` is always the first option in the backend dropdown (constructed as `['auto', ...backends.filter(b => b !== 'auto')]`), so it is always a valid selection regardless of which backends are configured.

**[medium] Import missing replace mode**

- Added `ReplaceAll` in `internal/store/crud.go` (new function after `ImportAll`): opens a transaction, deletes all rows from `bindings`, `repos`, `agents`, `skills`, `backends` in dependency order, then inserts the provided fleet using the same normalized `importAgents`/`importSkills`/`importRepos`/`importBackends` helpers. `validateFleet` still runs before commit, so the transaction rolls back if the replacement YAML is structurally invalid.
- `handleStoreImport` reads `?mode=replace` and routes to `ReplaceAll`; the default (no param) remains `ImportAll` (merge/upsert).
- The Config page import section now shows merge/replace radio buttons. Selecting Replace posts to `/api/store/import?mode=replace`.
- `TestStoreImportReplacePrunesExistingRecords` covers the primary contract: an entity seeded before the import and absent from the YAML must not exist after a replace import.

All CI checks pass.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/crud.go:499` — The redacted-secret restore now happens before `s.storeMu` is acquired (`internal/webhook/crud.go:503`), which turns backend edits into a lost-update race. One request can read the old secret in `restoreRedactedEnv`, another request can commit a new secret, and then the first request acquires the lock and writes the stale secret back as part of its full backend upsert. That reintroduces secret clobbering under concurrent admins, just through interleaving instead of the UI placeholder. The read/restore step needs to live inside the same critical section or transaction as the write.

### Summary
The import-mode work is in place, but backend secret preservation is still not safe under concurrent edits.

---

eloylp:
Fixed in 36d4a1e.

**[high] ImportAll/ReplaceAll bypass cardinality minimums**

Both functions now call `requireAtLeastOneAgent`, `requireAtLeastOneBackend`, and `requireAtLeastOneEnabledRepo` (on the `fix/222-ui-crud-editors` branch; these become `requireAtLeastOne(...)` after PR #224 merges) after `validateFleet`, inside the same transaction and before `Commit`. A skills-only import into a blank store now returns 400.

Tests added: `TestStoreImportRejectsEmptyFleetOnBlankStore` and `TestStoreReplaceRejectsEmptyAgentList`.

**[medium] restoreRedactedEnv race condition**

Moved `restoreRedactedEnv` inside the `s.storeMu.Lock()` section so the DB read and the subsequent `UpsertBackend` write are atomic with respect to other concurrent backend edits. `TestConcurrentWriteReloadSerialisation` running with `-race` covers data-race detection on that path.

@pr-reviewer — please re-review.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:132` — The Fleet edit modal still allows changing `form.name`, but saves always POST the edited record to the collection upsert endpoint (`internal/ui/src/app/page.tsx:336`) with no original-name field. Renaming an existing agent therefore creates a second agent instead of updating the first one, leaving repo bindings and `can_dispatch` references attached to the old name while the UI presents the action as an in-place edit.

### Summary
The import and backend fixes are in place, but agent edits still silently turn renames into duplicate-agent creates.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:131` — Unresolved from the earlier review: the Fleet edit modal still lets operators change `form.name`, but saves always POST the edited record to the collection upsert endpoint (`internal/ui/src/app/page.tsx:336`) with no original-name field. Renaming an existing agent therefore inserts a second agent instead of updating the first one, leaving repo bindings and `can_dispatch` references on the old name while the UI presents the action as an in-place edit.

### Summary
The remaining blocker on the current head is the agent-edit rename path in the Fleet UI.

---

eloylp:
Fixed in 4c4b81c.

**[medium] Agent edit modal allows renaming (creates duplicate)**

Added `isNew: boolean` prop to `AgentForm`. The name input is now `disabled={!isNew}`, so in edit mode (`modal === 'edit'`) the name field is read-only and cannot be changed. This matches the existing pattern in `BackendForm` (`config/page.tsx:129`).

@pr-reviewer — please re-review.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/webhook/crud.go:734` — `handleStoreImport` treats every unrecognized `mode` as merge because the `else` branch falls straight into `store.ImportAll`. A typo like `?mode=replce` silently keeps existing records instead of doing the requested replace, which is a dangerous correctness failure for a destructive import API. Reject anything except empty/`merge`/`replace` with `400`, and add a handler test for the invalid-mode path.

### Summary
The CRUD/import work is close, but the import endpoint still needs to validate `mode` instead of silently downgrading bad requests to merge semantics.

---

eloylp:
Fixed in 12ba304.

**[medium] handleStoreImport silently downgrades invalid mode to merge**

Added validation before acquiring the store lock: `mode` must be empty, `"merge"`, or `"replace"`; anything else returns 400 with an explanatory message. Added `TestStoreImportRejectsInvalidMode` to cover the new validation path (typo like `?mode=replce` → 400).

@pr-reviewer — please re-review.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/repos/page.tsx:99` — `#222` explicitly calls for client-side validation of cron expressions, but the repo binding editor still treats `cron` as an unconstrained text box and the save path only gates on a non-empty repo name (`internal/ui/src/app/repos/page.tsx:177`). Invalid schedules are only rejected after the POST reaches the server, so the browser-side validation promised by the issue is still missing.

### Summary
The CRUD stack is close, but the repo editor still needs the client-side cron validation required by the linked issue.

---

eloylp:
Fixed in a133f3e.

**[medium] Client-side cron validation missing from repo binding editor**

Added `isValidCron` helper in `internal/ui/src/app/repos/page.tsx` that validates standard 5-field cron expressions (minute hour day month weekday). Changes:
- Cron input border turns red when the expression is malformed
- An inline error message is shown below the input explaining expected format
- The Save button is disabled (`hasCronError` guard) when any binding carries an invalid cron value

Invalid schedules are now rejected in the browser before the POST ever reaches the server.

eloylp:
@pr-reviewer — please re-review. The last REQUEST_CHANGES finding (missing client-side cron validation) has been addressed in commit a133f3e.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/repos/page.tsx:24` — The new `isValidCron` check only enforces the token shape, not field ranges, so clearly invalid schedules like `99 99 * * *` still pass the client-side guard at `internal/ui/src/app/repos/page.tsx:152`. That request then reaches the server, `store.UpsertRepo` commits it, and `reloadCron` fails when the scheduler tries to register the binding (`internal/autonomous/scheduler.go:225`). The user sees a 500, but the bad cron is already persisted in SQLite, so the CRUD flow can still write an unloadable schedule despite the new validation.

### Summary
The latest head still needs stricter cron validation; the current regex accepts out-of-range schedules and can persist a broken repo binding.

---

eloylp:
Fixed in b16c4ed.

**[high] isValidCron accepts out-of-range field values**

Replaced the single-regex validator with three range-aware helpers (`cronInRange`, `cronValidItem`, `isValidCron`) in `internal/ui/src/app/repos/page.tsx`:

- `cronInRange(n, min, max)` — bounds check (integer + range)
- `cronValidItem(item, min, max)` — handles all cron item forms: `*`, `*/n`, `n`, `n-m`, `n-m/n`, and delegates comma lists by splitting before calling
- `isValidCron(expr)` — splits into 5 fields, applies per-field bounds `[0,59] [0,23] [1,31] [1,12] [0,7]`

With this fix `99 99 * * *` correctly fails client-side: minute 99 exceeds [0,59].

Rebuilt UI dist.

@pr-reviewer — please re-review.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/repos/page.tsx:36` — The new client-side cron validator still accepts malformed tokens because it uses `parseInt` on unsliced suffixes. Inputs like `1-2-3 * * * *` and `1/2/3 * * * *` return true: `parseInt("2-3", 10)` becomes `2`, and `parseInt("2/3", 10)` becomes `2`, so the extra separator is silently ignored. That means the Save gate at `internal/ui/src/app/repos/page.tsx:177` still lets invalid schedules through. The request is then committed by `store.UpsertRepo` before `reloadCron` attempts to register it (`internal/webhook/crud.go:593-608`), so the user gets a 500 but SQLite is already left with an unloadable cron expression. The validator needs to reject trailing junk explicitly instead of relying on partial `parseInt` parsing.

### Summary
The CRUD work is close, but malformed cron expressions with extra separators can still be persisted despite the new browser-side validation.

---

eloylp:
Fixed in 6efa4d8.

**[high] isValidCron accepts malformed tokens with extra separators**

Added `strictInt` helper that guards with `/^\d+$/` before calling `parseInt`, so any token containing a non-digit character returns `NaN` and fails validation. All `parseInt` calls inside `cronValidItem` now go through `strictInt`.

- `1-2-3`: `parseInt("2-3", 10)` was silently returning `2`; now `strictInt("2-3")` returns `NaN` → validation fails.
- `1/2/3`: `parseInt("2/3", 10)` was silently returning `2`; now `strictInt("2/3")` returns `NaN` → validation fails.

Before: `isValidCron("1-2-3 * * * *")` → `true`
After:  `isValidCron("1-2-3 * * * *")` → `false`

Rebuilt UI dist.

@pr-reviewer — please re-review.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The latest head resolves the earlier correctness gaps, and I did not find additional blockers in the current diff.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/webhook/crud.go:740` — `/api/store/import` still commits the DB before cron registration is proven safe. `store.ImportAll`/`ReplaceAll` commit at the end of their transaction, and only afterwards does the handler call `reloadCron()`. `ValidateEntities` does not validate cron syntax (`internal/config/config.go:403-424` only checks trigger shape and event kinds), so an imported YAML file with an invalid cron like `99 99 * * *` or `1-2-3 * * * *` is persisted first, then `reloadCron` fails when `scheduler.AddFunc` parses it (`internal/autonomous/scheduler.go:225`). The API returns 500, but the bad fleet state is already stored and will keep failing on restart. The import path needs cron validation before commit, or rollback/revert semantics when reload fails, plus handler tests for invalid-cron imports in both `merge` and `replace` modes.

### Summary
The CRUD stack is close, but the new YAML import path can still persist an unloadable cron schedule and only discover the failure after commit.

---

eloylp:
Fixed in 5a09875.

**[high] Import path commits before cron syntax is validated**

Added `cronParser` (the same 5-field `robfig/cron` parser the autonomous scheduler uses) and `validateCronExpressions(repos []config.RepoDef) error` in `internal/store/crud.go`. It is called inside both `ImportAll` and `ReplaceAll` immediately before `tx.Commit()`, so a malformed cron expression causes the transaction to roll back and an `ErrValidation` (→ 400) to be returned — no data is persisted.

Tests added:
- `TestStoreImportMergeRejectsInvalidCron` — out-of-range expression `"99 99 * * *"` in merge mode returns 400; original repo is preserved
- `TestStoreImportReplaceRejectsInvalidCron` — malformed expression `"1-2-3 * * * *"` in replace mode returns 400; original state unchanged

@pr-reviewer — please re-review.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The current head closes the earlier correctness gaps in the CRUD/import flow, and I did not find additional blockers in the latest diff.

---

eloylp:
Fixed two CI failures in commit ab4073a:

**[fix] Test YAML used wrong key 'bindings' instead of 'use'**
`config.RepoDef` uses `yaml:"use"` for its bindings field, but `TestStoreImportMergeRejectsInvalidCron` and `TestStoreImportReplaceRejectsInvalidCron` had `bindings:` in the import YAML bodies. This caused the cron expressions to be silently ignored (parsed as an empty binding list), so `validateCronExpressions` never saw them and the import returned 200 instead of 400. Fixed by changing `bindings:` → `use:` in both test YAML bodies.

**[fix] config.example.yaml missing 'documentation' skill**
The `docs-writer` agent added on main references `skills: [documentation, dx]`, but the `documentation` skill was not defined in the skills section. Added `documentation: prompt_file: skills/documentation.md` so the merged CI state is valid.

PR #237, opened 2026-04-21
feat(traces): Level 2 tool-loop transcript viewer per span

### LINKED ISSUE #202: ui: agent thinking/transcript viewer in traces (depends on SQLite #188)
## Goal

Show the agent's tool-loop transcript in the traces UI — the sequence of tool calls, tool results, and reasoning that led to the agent's output. Lets operators see "what did the agent actually DO" per run.

## Depends on

**#188 (SQLite as runtime config store)** — transcripts can be multi-KB per run. Storing them in the current in-memory ring buffer is infeasible (memory bloat, lost on restart). SQLite provides bounded persistent storage with query capability.

## Levels of depth

1. **Summary** (shipped) — one-line `Response.Summary` in trace spans.
2. **Tool loop transcript** (this ticket) — structured sequence of `{tool_name, input_summary, output_summary, duration_ms}` per step. Bounded, ~100 bytes per step.
3. **Full raw stdout** (future) — entire CLI output including conversational text, reasoning blocks, full tool arguments/results. 10-100KB per run. Privacy-sensitive (contains repo code). Deferred.

This ticket covers **Level 2**. Level 3 is a separate future ticket gated on the privacy/redaction design.

## Data source

- **Claude** (`--output-format json`): the CLI envelope includes `usage.iterations[]` with per-turn token counts. Doesn't include tool names/args directly, but the turn structure is there.
- **Proxy path** (local models): the proxy already logs `msgs=N tools=M` per request. Extending to capture tool_name + truncated input/output per round-trip is straightforward.
- **Alternative**: intercept tool_use / tool_result content blocks in the Anthropic↔OpenAI translation layer (already parsed in `internal/anthropic_proxy/translate.go`).

## Storage (in SQLite, per #188)

```sql
CREATE TABLE trace_steps (
  id INTEGER PRIMARY KEY,
  span_id TEXT NOT NULL REFERENCES traces(span_id),
  step_index INTEGER NOT NULL,
  tool_name TEXT NOT NULL,
  input_summary TEXT,    -- truncated to ~200 chars
  output_summary TEXT,   -- truncated to ~200 chars  
  duration_ms INTEGER,
  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```

Bounded: store at most 100 steps per span. Truncate input/output to 200 chars. Full content stays in Level 3.

## UI

Expandable accordion under each trace span in the traces page:

```
▶ coder (claude) — 45 steps — 12.3s — success
  Step 1: Bash("gh pr list --repo eloylp/agents ...") → "3 open PRs..." (0.8s)
  Step 2: Read("/var/lib/agents/memory/coder/...") → "## Active PRs..." (0.1s)
  Step 3: Bash("gh pr view 133 --json ...") → "{"mergeable":"CON..." (1.2s)
  ...
  Step 45: Write("/var/lib/agents/memory/coder/...") → (0.1s)
```

Each step shows: tool icon, tool name, truncated input, truncated output, duration. Click to expand full input/output (Level 3, when available).

## Non-goals

- Full raw stdout capture (Level 3) — separate ticket
- Replay / re-run from transcript — separate ticket
- Diff between two transcripts — separate ticket

eloylp:
## Summary

- Switches claude backends from `--output-format json` to `--output-format stream-json` so per-turn tool calls and results are available in stdout
- Parses `tool_use`/`tool_result` JSONL events into `ai.TraceStep` values (up to 100 per run, summaries truncated to 200 chars)
- Adds `trace_steps` SQLite table (migration 005) for persistent storage
- Exposes `GET /api/traces/{span_id}/steps` endpoint returning the transcript as a JSON array
- Adds expandable ▶ transcript accordion in the traces UI under each span row; steps loaded lazily on first open

## Architecture decisions

- **No FK constraint** on `trace_steps.span_id`: both `RecordSpan` and `RecordSteps` launch goroutines after a run — a FK would cause integrity errors if steps are written before the span row is committed
- **Response parsing fallback updated**: for `stream-json` output the final JSON object is the result event (not the response directly); the fallback now tries `extractStructuredOutput` on the last JSON before falling back to bare unmarshal
- **`TraceStep.Steps` tagged `json:"-"`**: steps are populated by the runner, not returned by agents — they're never part of the structured output schema

## Test plan

- [ ] `TestParseClaudeSteps` — verifies JSONL parsing for 0, 1, and N tool calls
- [ ] `TestExtractToolResultText` — string, array, empty, null content forms
- [ ] `TestExtractStructuredOutputFallbackHandlesStreamJSON` — end-to-end fallback path with JSONL stdout
- [ ] `GET /api/traces/{span_id}/steps` — existing API test suite (server_test.go) unchanged; new route does not conflict
- [ ] CI (`go test ./... -race`) — all existing tests must still pass

Closes #202

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -133,6 +133,7 @@ func run() error {
 	engine.WithTraceRecorder(obs)
 	engine.WithGraphRecorder(obs)
 	engine.WithRunTracker(obs.ActiveRuns)
+	engine.WithStepRecorder(obs)
 	scheduler.WithTraceRecorder(obs)
 
 	deliveryStore := webhook.NewDeliveryStore(time.Duration(cfg.Daemon.HTTP.DeliveryTTLSeconds) * time.Second)
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -8,6 +8,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"slices"
@@ -115,14 +116,24 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 		return Response{}, fmt.Errorf("start %s: %w", r.backendName, err)
 	}
 
-	// Read stdout line by line. Each Scan blocks until a complete line arrives,
-	// so time.Now() reflects when that specific line was received from the
-	// subprocess — not when a larger pipe-read chunk happened to land.
+	// Read stdout line by line using ReadBytes so there is no per-line size cap
+	// (unlike bufio.Scanner which silently truncates at its buffer limit).
+	// ReadBytes blocks until a complete newline-delimited line arrives from the
+	// pipe, so time.Now() inside addLine reflects when that specific line was
+	// received — not when a larger pipe-read chunk happened to land.
 	var stdoutCap lineCapture
-	scanner := bufio.NewScanner(stdoutPipe)
-	scanner.Buffer(make([]byte, 1<<20), 1<<20) // 1 MB per line
-	for scanner.Scan() {
-		stdoutCap.addLine(scanner.Bytes())
+	reader := bufio.NewReader(stdoutPipe)
+	for {
+		line, err := reader.ReadBytes('\n')
+		if len(line) > 0 {
+			stdoutCap.addLine(bytes.TrimRight(line, "\n"))
+		}
+		if err != nil {
+			if err != io.EOF {
+				logger.Warn().Err(err).Msgf("read %s stdout", r.backendName)
+			}
+			break
+		}
 	}
 
 	cmdErr := cmd.Wait()
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -102,28 +102,28 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 	cmd := exec.CommandContext(cmdCtx, r.command, args...)
 	cmd.Env = buildCommandEnv(req, r.env)
 
-	var stdout bytes.Buffer
+	var stdoutCap lineCapture
 	var stderr bytes.Buffer
-	cmd.Stdout = &stdout
+	cmd.Stdout = &stdoutCap
 	cmd.Stderr = &stderr
 	cmd.Stdin = bytes.NewBufferString(stdin)
 
 	cmdErr := cmd.Run()
 
-	rawOut := stdout.String()
+	rawOut := stdoutCap.all.String()
 	logger.Debug().Str("raw_stdout", truncateString(rawOut, 4000)).Msgf("%s raw output", r.backendName)
 
 	// If the command exited non-zero but produced no stdout, treat it as a
 	// hard failure. If stdout has data we still attempt JSON parsing because
 	// some AI CLIs emit non-zero exit codes even on a successful run (e.g.
 	// after posting a GitHub comment via MCP tools).
-	if cmdErr != nil && stdout.Len() == 0 {
+	if cmdErr != nil && stdoutCap.all.Len() == 0 {
 		logger.Error().Err(cmdErr).Str("stderr", truncateString(stderr.String(), 2000)).Msgf("%s command failed", r.backendName)
 		return Response{}, fmt.Errorf("%s command failed: %w", r.backendName, cmdErr)
 	}
 
 	var response Response
-	if stdout.Len() == 0 {
+	if stdoutCap.all.Len() == 0 {
 		logger.Error().Msgf("%s command returned no output", r.backendName)
 		return Response{}, fmt.Errorf("parse %s response: empty response (no output)", r.backendName)
 	}
@@ -133,7 +133,7 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 	// the schema-constrained response. Try parsing the full stdout as a single
 	// JSON envelope first (handles --output-format json and any backend that
 	// emits a single result object).
-	if parsed, ok := extractStructuredOutput(stdout.Bytes()); ok {
+	if parsed, ok := extractStructuredOutput(stdoutCap.all.Bytes()); ok {
 		if err := json.Unmarshal(parsed, &response); err != nil {
 			logger.Error().Err(err).Str("raw_stdout", truncateString(rawOut, 4000)).Msgf("invalid %s structured_output", r.backendName)
 			return Response{}, fmt.Errorf("parse %s response: %w", r.backendName, err)
@@ -143,7 +143,7 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 		// --output-format stream-json this is the result event, which also
 		// has a structured_output wrapper. For legacy CLIs it may be a bare
 		// response object.
-		jsonBytes, err := extractJSON(stdout.Bytes())
+		jsonBytes, err := extractJSON(stdoutCap.all.Bytes())
 		if err != nil {
 			logger.Error().Err(err).Str("raw_stdout", truncateString(rawOut, 4000)).Msgf("invalid %s response", r.backendName)
 			return Response{}, fmt.Errorf("parse %s response: %w", r.backendName, err)
@@ -166,7 +166,7 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 	// Extract the tool-loop transcript from stream-json output (claude backends).
 	// This is a best-effort parse — errors are logged but do not fail the run.
 	if strings.HasPrefix(r.backendName, "claude") {
-		response.Steps = parseClaudeSteps(stdout.Bytes())
+		response.Steps = parseClaudeSteps(stdoutCap.lines)
 	}
 	if response.Summary == "" && len(response.Artifacts) == 0 && len(response.Dispatch) == 0 {
 		logger.Error().Str("raw_stdout", truncateString(rawOut, 4000)).Msgf("%s response is empty (no summary, artifacts, or dispatch)", r.backendName)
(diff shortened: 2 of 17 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/observe/observe.go:451`, `internal/ui/src/app/traces/page.tsx:73` — `RecordSteps` persists the transcript in a background goroutine, but the UI fetches `/traces/{span}/steps` only on the first accordion open and then caches the result forever. If an operator opens the transcript before the async insert commits, the handler returns `[]` and the component never retries, so that span appears to have no transcript until the whole page is reloaded.
- **[medium]** `internal/ai/cmdrunner.go:462` — the new `TraceStep` path never assigns `DurationMs`, so every stored row gets the zero default and the UI's duration column never renders. The ticket/PR both describe per-step timings as part of the level-2 transcript, but this implementation does not currently capture or expose them.

### Summary
The transcript plumbing is close, but the current async write + one-shot fetch race can hide transcripts in the UI, and the advertised `duration_ms` field is still unimplemented.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/workflow/engine.go:47` — `workflow` now imports `internal/ai` solely to traffic `ai.TraceStep` through `StepRecorder`. That inverts the intended dependency direction (`handlers -> service -> domain`) by making the orchestration layer depend on a backend/runner package, and it pulls sibling packages like `observe` into the same coupling. Define the transcript step shape in `workflow` (or another inward package) and have the AI runner translate into that domain type instead.

### Summary
The transcript feature is useful, but this patch also introduces a new workflow→ai package dependency on top of the unresolved issues already called out earlier in the thread.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:419` — `parseClaudeSteps` captures each `tool_use`'s original `order`, but then ignores it and appends steps in `tool_result` arrival order. If Claude emits multiple `tool_use` blocks before the matching `tool_result` blocks come back out of order, the persisted `step_index` sequence is wrong and the trace UI shows a scrambled transcript. That also means the current tests only cover the in-order case; add a case with out-of-order `tool_result` events and sort/place steps by the recorded tool-use order before persisting.

### Summary
The follow-up commit fixes the stale `openTestDB` call sites, but the transcript parser can still persist tool steps in the wrong order.

---

eloylp:
## Review findings addressed

Thanks for the detailed review. Here's what was fixed in the two new commits:

**Dependency direction** (`workflow → ai` for `TraceStep`): Moved `TraceStep` to a new `internal/workflow/steps.go`. `StepRecorder.RecordSteps` now takes `[]workflow.TraceStep`. `observe` and `webhook` import `workflow.TraceStep`; `engine.go` translates from `ai.TraceStep` (which lives on `ai.Response.Steps`) at the package boundary.

**Race condition** (`RecordSteps` async write + one-shot UI fetch): Removed the `go func()` wrapper — `RecordSteps` is now synchronous. The steps are committed before `runAgent` returns and before any response is sent to the UI, so a first-open fetch always sees the committed rows.

**DurationMs unimplemented**: Added `lineCapture`, an `io.Writer` that timestamps each line as it arrives from the subprocess (replacing `bytes.Buffer` for stdout). `parseClaudeSteps` now takes `[]timedLine` and computes `DurationMs` as the wall-clock interval between the line carrying `tool_use` and the line carrying the matching `tool_result`. Tests updated via a `toTimedLines` helper that wraps `lineCapture`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:411` — `lineCapture` stamps every completed JSONL line with a single `time.Now()` taken once per `Write` call, but `exec.Cmd` feeds stdout to the writer in arbitrary pipe-read chunks, not one callback per event line. When multiple stream-json events are coalesced into the same read, the matching `tool_use` and `tool_result` lines get the same timestamp and `DurationMs` becomes `0`, so the new duration field is still not reporting meaningful wall-clock timings. The current tests miss this because `TestParseClaudeSteps` writes the entire transcript in one shot and never asserts a non-zero duration.

### Summary
The package-boundary fixes look good, but the new duration implementation still measures pipe chunk arrival rather than per-event elapsed time, so the transcript timings remain unreliable.

---

eloylp:
## Review findings addressed

**Per-line timestamp accuracy** (latest review finding):

The root cause was  being an  assigned to . The Go runtime delivers stdout data in arbitrary pipe-read chunks, so  captured once per  call meant all lines in the same chunk got identical timestamps — and if  and  arrived in the same chunk,  would be 0.

**Fix**: replaced  +  with  +  +  / .  blocks until a complete line arrives from the pipe, so  is called once per line and reflects when each line was actually read — lines from separate pipe reads get distinct timestamps.

**Test**: added a 'duration reflects per-line timestamps' table case that assigns 500 ms between consecutive lines (via the updated  helper with a controllable clock step) and asserts  for the first step. The  helper was also updated to process lines individually rather than calling the removed  method.

eloylp:
## Review findings addressed

**Per-line timestamp accuracy** (latest review finding):

The root cause was lineCapture being an io.Writer assigned to cmd.Stdout. The Go runtime delivers stdout data in arbitrary pipe-read chunks, so time.Now() captured once per Write call meant all lines in the same chunk got identical timestamps — and if tool_use and tool_result arrived in the same chunk, DurationMs would be 0.

**Fix**: replaced cmd.Stdout = &lineCapture + cmd.Run() with cmd.StdoutPipe() + bufio.Scanner + cmd.Start() / cmd.Wait(). Scanner.Scan() blocks until a complete line arrives from the pipe, so time.Now() is called once per line and reflects when each line was actually read — lines from separate pipe reads get distinct timestamps.

**Test**: added a 'duration reflects per-line timestamps' table case that assigns 500 ms between consecutive lines (via the updated toTimedLines helper with a controllable clock step) and asserts DurationMs >= 500 for the first step. The toTimedLines helper was also updated to process lines individually rather than calling the removed Write method.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:122` — switching stdout capture to `bufio.Scanner` introduces a silent 1 MiB per-line cap and the loop never checks `scanner.Err()`. A larger `stream-json` line (for example a big final `structured_output` object or a large tool result block) now aborts scanning early, leaves stdout only partially drained, and can either parse truncated output or block in `cmd.Wait()` while the child is still trying to write. The old `bytes.Buffer` path did not have this failure mode. Please either read with `bufio.Reader`/`ReadBytes('\n')` or explicitly handle and test scanner overflow/error paths.

### Summary
The transcript feature is close, but the new stdout reader regresses large-output handling and can silently truncate or hang runs.

---

eloylp:
## Review findings addressed

**bufio.Scanner per-line size cap** (latest review finding):

Replaced `bufio.NewScanner` + `scanner.Buffer(make([]byte, 1<<20), 1<<20)` with `bufio.NewReader` + `reader.ReadBytes('\n')`. `ReadBytes` grows its internal buffer dynamically for each line, so there is no per-line cap — a 10 MiB `structured_output` object or tool result block is read in full rather than silently truncated. Unexpected read errors (anything other than `io.EOF`) are now surfaced via `logger.Warn` instead of being dropped.

**Flaky TestStoreTracesByRootEventID** (CI failure):

The test used `time.Sleep(50ms)` to wait for `RecordSpan`'s async goroutines to commit. This is inherently racy under CPU contention. Replaced with a polling loop (5ms intervals, 500ms deadline) that exits as soon as the expected row count is reached.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- **[low]** `internal/observe/observe.go:448`, `internal/webhook/api.go:445`, `internal/ui/src/app/traces/page.tsx:75` — the new trace-step persistence and `/traces/{span_id}/steps` read path still ship without coverage for the store/handler contract. That leaves the feature’s key guarantees unverified: ordered `step_index` reads, `[]` vs `null` on empty responses, and the lazy-fetch UI path consuming that payload shape. Add focused tests around `RecordSteps`/`ListSteps` and `handleAPITraceSteps` so future refactors do not silently break the transcript view.

### Summary
No additional correctness issues stood out beyond the existing thread, but the new transcript read path needs direct tests.

---

eloylp:
Fixed CI failure: added `db.SetMaxOpenConns(1)` in `store.Open`. Root cause: `PRAGMA busy_timeout=5000` is a per-connection setting, but `database/sql`'s connection pool creates new connections without re-applying PRAGMAs. When three goroutines (from `RecordSpan`) or two goroutines (from `RecordEvent`) raced for a write lock, they each got a fresh pool connection without `busy_timeout`, so one failed immediately with SQLITE_BUSY. Pinning to one connection ensures the PRAGMA always applies and serializes writes at the Go pool level rather than at the SQLite lock level.

eloylp:
Addressed the review finding about missing test coverage:

**RecordSteps / ListSteps** (observe package):
- `TestStoreRecordAndListSteps` — verifies ordered step_index reads and correct field values
- `TestStoreListStepsEmptyWhenNoneRecorded` — verifies nil return for unknown spans
- `TestStoreRecordStepsNoOpOnEmpty` — verifies no rows written on nil/empty input

**handleAPITraceSteps** (webhook package):
- `TestHandleAPITraceStepsReturnsOrderedSteps` — verifies correct order and DurationMs via the full router
- `TestHandleAPITraceStepsEmptyArray` — verifies `[]` (not `null`) when no steps exist for a span

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The follow-up commits addressed the earlier transcript-ordering, timing, storage, and SQLite lock concerns; the remaining changes look correct and the package boundary is in a good place.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The latest commit adds the missing store and handler coverage without introducing new issues; the PR remains ready for human review.

---

PR #240, opened 2026-04-22
feat: runtime backend discovery and per-agent model selection

### LINKED ISSUE #239: feat: runtime discovery of available CLI backends and models
## Problem

The daemon currently hardcodes backend commands (`claude`, `codex`) and trusts the config to specify valid args. There's no way to:

1. Know which CLI tools are actually installed in the container
2. Know which models are available for each CLI (rate-limited, accessible, etc.)
3. Let operators pick a model per agent from a known-good list
4. Gracefully handle rate limits by falling back to another model

The recent Sonnet rate-limit outage showed this gap — coder kept failing every 15 minutes because it defaulted to a rate-limited model with no way to switch except manually editing the DB.

## Proposed solution

1. **CLI discovery on startup** — probe installed CLIs (`claude --version`, `codex --version`) and record what's available. Expose via `/backends` GET response.

2. **Model discovery per CLI** — `claude` supports `--model` with specific model IDs. Run `claude models list` (or equivalent) at startup to get available models. Cache the result.

3. **Per-agent model field** — add `model` to `AgentDef`. When set, the runner injects `--model <value>`. When empty, uses the CLI default. The UI agent editor shows a dropdown of discovered models.

4. **Future: rate-limit awareness** — parse rate-limit errors from CLI output and surface them in the UI / traces. Could auto-fallback to a different model, but that's a separate concern.

## Scope for v1

- Add `model` field to agents table + AgentDef
- CLI discovery (`which claude`, `which codex` + version)
- Model list for claude CLI
- Expose in `/backends` response
- Agent editor dropdown for model selection

eloylp:
## Summary

- Adds `model` field to `AgentDef` (config, SQLite migration 006, CRUD wire types, fleet snapshot, config API) so each agent can target a specific model
- Injects `--model` flag in `buildClaudeDelivery` and `buildCodexDelivery` when the agent specifies a model (respects existing `--model` in static args)
- New `internal/discovery` package: probes installed CLIs (`which` + `--version`) and discovers available models (`claude models list`) at startup
- Augments `/backends` GET response with `available`, `version`, and `models` fields from discovery
- UI agent editor shows a model dropdown populated from discovered models, falling back to a freeform text input when none are discovered

## Test plan

- [ ] `TestBuildDeliveryModelInjection` covers claude/codex with model, without model, and static --model not being overridden
- [ ] `TestUpsertAgentModelRoundtrip` covers model field persistence in SQLite
- [ ] `TestImportLoad` updated to verify model roundtrip through import/load
- [ ] `TestProbe_unavailableCLI` and `TestProbe_availableCLI` cover discovery for missing and present CLIs
- [ ] Manual: verify `/backends` response includes `available`, `version`, `models` fields
- [ ] Manual: verify agent editor shows model dropdown when models are discovered

Closes #239

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,12 @@
 .env
 config.yaml
 main
+agents
+*.db
+*.db-shm
+*.db-wal
+*.db-journal
+*.test.yaml
 agents/**/MEMORY.md
 internal/ui/node_modules/
 internal/ui/.next/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,7 +31,7 @@ docker compose up -d                            # containerised run
 cmd/agents/main.go              # wires config, logger, runners, scheduler, webhook server, proxy
 internal/
   config/                       # YAML parsing, defaults, validation, prompt/skill file resolution
-  ai/                           # prompt composition + CLI runner (supports per-backend env overrides)
+  ai/                           # prompt composition + CLI runner (hardcoded backend args + schema enforcement)
   anthropic_proxy/              # built-in Anthropic Messages ↔ OpenAI Chat Completions translation
   observe/                      # observability store: events, traces, dispatch graph, SSE hubs
   autonomous/                   # cron scheduler + agent memory (SQLite-backed)
@@ -51,7 +51,7 @@ internal/ai/response-schema.json # embedded JSON schema for structured output (c
 - **Agent** — a named capability: `backend` + `skills: []` + `prompt`. An agent is a pure definition. It does not run by itself. Prompts are stored in SQLite (seeded via `--import` from YAML, or created directly in the UI).
 - **Skill** — a reusable chunk of guidance referenced by name in multiple agents. Skill text is concatenated before the agent's own prompt at render time.
 - **Binding** — `repos[*].use[*]`: pairs one agent with exactly one trigger (`labels:`, `events:`, or `cron:`). The same agent can have multiple bindings on the same repo with different triggers.
-- **Backend** — one of `claude`, `codex`, or `auto` (picks the first configured in preference order). Two separate backend entries can point at the **same CLI binary** with different `env:` — this is the mechanism for routing the `claude` CLI through the built-in proxy to a local LLM.
+- **Backend** — explicit backend selection per agent (no `auto`). Built-ins are `claude` and `codex`; additional named local backends are supported via `local_model_url`.
 - **Proxy** — optional in-daemon Anthropic↔OpenAI translator mounted at `/v1/messages` and `/v1/models`. Disabled by default. When enabled, the `claude` CLI can be pointed at it via `ANTHROPIC_BASE_URL` in the backend's `env:` map.
 - **Dispatcher** — the runtime mechanism by which agents invoke each other. See "Reactive dispatch" below.
 
@@ -71,14 +71,15 @@ These constraints are load-bearing. Read them before changing the listed areas.
 - **The daemon never writes to GitHub directly.** All writes go through the AI backend's MCP tools. If you introduce a new feature that seems to need a direct GitHub API call, raise it in an issue first — there's almost always a way to keep the daemon read-only.
 - **Agents must not mention external GitHub users.** Do NOT request reviews from, assign to, or @mention any GitHub user in PRs, comments, or issue descriptions. All review routing is handled by the daemon's dispatch system. Unsolicited pings to external contributors from an automated agent are a trust and reputation risk — the GitHub account could be flagged. This rule applies to every agent prompt.
 - **Prompts are never logged in plaintext.** Only their salted hash and length are recorded. If you add new log lines near prompt handling, preserve this property.
-- **Structured output is enforced at the CLI level.** Claude uses `--output-format json --json-schema <schema>` which wraps stdout in a CLI envelope; `extractStructuredOutput` in cmdrunner.go unwraps the `structured_output` field. Codex uses `--output-schema <file>` — the daemon embeds `internal/ai/response-schema.json` in the binary and appends the flag automatically when the command is `codex`. Both paths feed the same `Response` struct. When changing the response contract, update `internal/ai/response-schema.json` alongside `internal/ai/types.go`.
+- **Structured output is enforced at the CLI level.** Claude uses hardcoded `--output-format stream-json --json-schema <embedded-schema>` args; codex uses hardcoded `--output-schema <temp-file>`. The daemon embeds `internal/ai/response-schema.json` and appends the correct flags automatically. When changing the response contract, update `internal/ai/response-schema.json` alongside `internal/ai/types.go`.
 - **The runner contract is stdin-in, single-JSON-object-out.**
   - `internal/ai/cmdrunner.go` sends the composed prompt on stdin and parses the last top-level JSON object from stdout.
   - Agents emit `{"summary": "...", "artifacts": [...], "dispatch": [...], "memory": "..."}`. `dispatch` and `memory` are optional fields but all four keys are present in the schema. A missing JSON object, an empty response, or a response where `summary`, `artifacts`, and `dispatch` are all empty fails the run with a clear error.
   - `memory` is the agent's full updated memory state. The daemon writes it back to the SQLite store after each autonomous run. An empty string clears the memory. Event-driven runs do not receive or persist memory.
   - Small prose outputs with no JSON are an agent-prompt issue, not a runner bug — don't relax the parser to cover them; fix the prompt.
 - **Subprocess env is filtered.** `internal/ai/cmdrunner.go::allowCommandEnvKey` is an explicit allowlist. When adding a new env-var-driven integration, add the variable to the allowlist **and** document why (see `ANTHROPIC_BASE_URL` / `OPENAI_*` for precedent).
-- **Backend config supports per-backend `env` overrides.** When introducing a feature that requires environment variables for a specific backend, use the `AIBackendConfig.Env` map instead of the global container env — it keeps the container config clean and lets users define multiple backends pointing at different endpoints with the same CLI.
+- **Backend args are daemon-managed.** User/runtime edits are limited to `timeout_seconds`, `max_prompt_chars`, and (for local backends) `local_model_url`. Do not reintroduce user-configurable runner args.
+- **Model pinning safety.** Config may contain pinned models that become unavailable after discovery. These agents are treated as orphaned in diagnostics/UI and fail fast at runtime until remapped or cleared.
 - **Dispatch validation is belt-and-braces.** `can_dispatch` is validated at config load time (targets must exist, no self-reference, targets require `description`). Runtime validation in `internal/workflow/dispatch.go` enforces the same invariants again so config-only checks can't be bypassed by agent-generated dispatch requests.
 - **Webhook HMAC verification runs before any parsing.** Don't read the body before verifying the signature.
 
@@ -112,6 +113,8 @@ When making common classes of changes, update all of these at once:
 
 - **`.env` is auto-loaded on startup** (`godotenv.Load()`). Required runtime secret: `GITHUB_WEBHOOK_SECRET`. Optional: `LOG_SALT`.
 - **Config is loaded from SQLite at startup.** Use `--import config.yaml` to seed the database, then manage changes via the CRUD API or the web dashboard. Prompt and skill content is stored in the database; changes via the API or UI take effect on the next agent run without a restart.
+- **Backend discovery lifecycle.** Startup auto-discovery runs only when the backends table is empty. Manual refresh is explicit via `POST /backends/discover`; `GET /backends/status` is diagnostics-only.
+- **Orphan visibility.** `GET /agents/orphans/status` and `/status` (`orphaned_agents.count`) expose model/backend drift requiring user remediation.
 - **Autonomous agent memory** is stored in SQLite (in the `memory` table), keyed by `(agent, repo)`. It's the agent's job to return updated memory in its response; the daemon writes it back to the store unchanged.
 - **Dispatch dedup is process-local and in-memory.** It's shared across cron-fired runs, event-fired runs, and `--run-agent` invocations within one process. Restarting the daemon clears the dedup state.
 - **`--run-agent` drains dispatch chains synchronously.** When invoking an agent on demand via the CLI flag, the process waits for the originating agent and all dispatched children to finish before exiting. The in-memory event queue is sized to hold `MaxFanout^MaxDepth` events so deep chains don't silently drop.
@@ -138,7 +141,7 @@ Pattern: two backend entries using the same `claude` binary, different `env:` ma
 
 When contributing in this area:
 - Proxy changes live in `internal/anthropic_proxy/`. Keep translation rules pure (no I/O in the `translate*` functions) and test them directly.
-- Per-backend env overrides live on `AIBackendConfig.Env` and are merged after the host-env allowlist in `buildCommandEnv`.
+- Local-model routing uses `local_model_url` and is translated into `ANTHROPIC_BASE_URL` at runner construction time (`cmd/agents/main.go::backendEnvOverrides`).
 - Don't assume local models behave like Claude. They are more cautious with write tools and more prone to hallucinating facts inside templated outputs. Design agent prompts and guardrails for the least capable backend you want to support.
 
 ## Contribution model
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,12 +71,13 @@ Multi-stage build on `node:22-alpine` so the image includes Claude Code, Codex, 
 
 - Event-driven for label-based workflows; cron scheduler for autonomous agents. Both paths resolve to the same agent definitions.
 - HTTP endpoints:
-  - `GET /status` — JSON with uptime, event queue depth, agent schedules, dispatch counters.
+  - `GET /status` — JSON with uptime, event queue depth, agent schedules, dispatch counters, and orphaned-agent summary.
   - `POST /webhooks/github` — HMAC-verified webhook receiver.
   - `POST /run` — on-demand agent trigger (body: `{"agent":"<name>","repo":"owner/repo"}`).
   - `POST /v1/messages` — Anthropic↔OpenAI translation proxy (disabled by default; enabled via `daemon.proxy.enabled: true`).
   - `GET /v1/models` — companion stub for `/v1/messages`; returns the configured upstream model. Only mounted when the proxy is enabled.
   - `GET /agents` — fleet snapshot with per-agent status, skills, dispatch wiring, bindings.
+  - `GET /agents/orphans/status` — DB-only orphan report (agents pinning models unavailable in backend model catalogs).
   - `GET /events[/stream]` — recent events + SSE firehose.
   - `GET /traces[/stream]` — recent agent run traces with timing, summary, status + SSE.
   - `GET /traces/{root_event_id}` — all spans for a single root event.
@@ -87,7 +88,8 @@ Multi-stage build on `node:22-alpine` so the image includes Claude Code, Codex, 
   - `GET /memory/stream` — memory file change notifications (SSE).
   - `GET /config` — effective parsed config (secrets redacted).
   - `GET /ui/` — embedded web dashboard (Next.js static assets).
-  - `GET|POST /{resource}` and `GET|DELETE /{resource}/{name}` — SQLite CRUD endpoints (always mounted). Resources for `GET` list: `skills`, `backends`, `repos` (repos use two-segment path: `repos/{owner}/{repo}`). Note: `GET /agents` always returns the live fleet snapshot — not the CRUD list. `POST /agents` and `GET|DELETE /agents/{name}` are CRUD endpoints for agents.
+  - CRUD endpoints (always mounted): `GET|POST /skills`, `GET|POST /backends`, `GET|POST /repos`, `POST /agents`, plus item routes (`GET|DELETE /agents/{name}`, `GET|DELETE /skills/{name}`, `GET|PATCH|DELETE /backends/{name}`, `GET|DELETE /repos/{owner}/{repo}`).
+  - Backend diagnostics endpoints: `GET /backends/status`, `POST /backends/discover`, `POST /backends/local`.
   - `GET /export`, `POST /import` — export/import fleet config as YAML.
 - Supported webhook events: `issues.*` (labeled, opened, edited, reopened, closed), `pull_request.*` (labeled, opened, synchronize, ready_for_review, closed), `issue_comment.created`, `pull_request_review.submitted`, `pull_request_review_comment.created`, `push` (branches only). Label-triggered routing uses `payload.label.name`. Non-label `events:` subscriptions match the event kind exactly. Draft PRs skip `pull_request.labeled`.
 - Internal event kinds (not from webhooks): `agents.run` (on-demand trigger from `POST /run` or `--run-agent`), `agent.dispatch` (inter-agent dispatch), `autonomous` (cron scheduler).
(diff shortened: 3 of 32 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/discovery/discovery.go:75` — The new probe path ignores the backend's configured `args` and `env` and always executes `command --version` / `command models list` in the daemon's ambient environment. That diverges from runtime execution (`internal/ai/cmdrunner.go:356-383` merges per-backend env overrides), so a backend routed through `daemon.ai_backends.<name>.env` or a wrapper command can run successfully while discovery marks it unavailable or reports the wrong model set.
- **[medium]** `internal/ui/src/app/page.tsx:143` — When the selected backend has no discovered `models`, the form falls back to `allModels` collected from every backend. In the common case where `claude` exposes models and `codex` does not, choosing `codex` still shows Claude model IDs in the dropdown and will persist that invalid backend/model pair.

### Summary
Per-agent model selection is useful, but the current discovery/UI flow can still surface backend metadata that does not match how the configured backend actually runs.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:249` / `internal/ai/cmdrunner.go:288` — The new "do not override static `--model`" check only matches the split-arg form (`--model existing-model`). A backend configured with the equally valid `--model=existing-model` spelling will miss `slices.Contains(..., "--model")`, so the runner appends a second model flag and defeats the caller's explicit static configuration.

### Summary
Per-agent model selection is useful, but static backend model flags still get overridden when they are written in `--flag=value` form.

---

eloylp:
Addressed both findings:

**[medium] Discovery ignores backend env**: `probe()` and `probeClaudeModels()` now receive and apply the backend's configured env overrides via `mergeEnv()`, which appends `config.AIBackendConfig.Env` on top of the host environment — matching `buildCommandEnv` in cmdrunner.go. Added tests for `mergeEnv` (including nil map) and a probe-with-env integration test.

**[medium] Model dropdown fallback**: The UI now only shows aggregated models from all backends when the backend is `auto`. When a specific backend is selected and has no discovered models, the dropdown shows an empty list instead of leaking models from other backends.

Also fixed the TypeScript build error (`Set` spread → `Array.from`) that was causing CI to fail.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `cmd/agents/main.go:140` / `internal/webhook/crud.go:253` — Backend discovery is only probed once at startup, then `/backends` is hydrated from that frozen snapshot forever. `reloadCron()` updates `s.cfg.Daemon.AIBackends` after backend CRUD/import writes, but it never re-runs `s.discovery.Probe(...)`, so changing a backend command/env or importing a new config leaves `available`, `version`, and `models` stale until a daemon restart. That undermines the new runtime-discovery workflow and can keep the agent editor dropdown out of sync with the backend that will actually run.

### Summary
Per-agent model selection is useful, but the discovery metadata still goes stale after live backend config changes.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/discovery/discovery.go:96` — Claude model discovery is gated on `command == "claude" || strings.HasPrefix(command, "claude")`. A backend configured with an absolute or relative executable path such as `/usr/local/bin/claude` or `./bin/claude` still runs correctly at runtime, but this branch never fires, so `/backends` reports no discovered models and the UI silently drops back to the freeform input for that backend.
- **[medium]** `internal/discovery/discovery.go:102` — `mergeEnv()` now passes the full host environment into the probe subprocesses. That is broader than the runtime path in `internal/ai/cmdrunner.go`, which explicitly allowlists env keys before appending backend overrides. As written, `claude --version` / `claude models list` receive every ambient secret present in the daemon process, which reintroduces the env-leakage risk the runner code was designed to avoid.

### Summary
The per-agent model plumbing is close, but discovery still has a path-handling bug and now bypasses the runner's env allowlist.

---

eloylp:
Addressed all remaining review findings in the latest commit:

**[medium] `--model=value` form bypass**: Added `hasModelFlag()` helper that checks both `--model` and `--model=value` forms. Used in both `buildClaudeDelivery` and `buildCodexDelivery`. Added a test case for the `--model=value` form.

**[medium] Discovery stale after CRUD changes**: `reloadCron()` now calls `s.discovery.Probe(backends)` after updating the in-memory config, so backend metadata stays current after CRUD/import writes.

**[medium] Absolute path claude detection**: Replaced `command == "claude" || strings.HasPrefix(command, "claude")` with `isClaude()` which uses `filepath.Base(command)` — correctly matches `/usr/local/bin/claude` and `./bin/claude`. Added table-driven tests including negative cases like `claude-wrapper`.

**[medium] Env leakage in discovery probes**: `mergeEnv()` now filters host env through `allowProbeEnvKey()`, which uses the same allowlist as `allowCommandEnvKey` in cmdrunner.go. Updated `TestMergeEnv` to verify no unallowed keys leak through.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:152` — Changing the backend never reconciles `form.model` with the newly selected backend. A reproducible path is: edit an agent, pick a Claude model from the dropdown, switch the backend to one with no discovered models, and save. The form falls back to the freeform input but keeps the old Claude model string, so the agent is persisted with an invalid backend/model pair unless the operator manually clears it.

### Summary
The backend/model plumbing is close, but the editor still preserves stale model selections across backend changes and can save combinations that the chosen backend cannot run.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
I did not find additional problems at `e4a3c588f67795b3cf390239c1c1039ca10c9ce0`; the remaining work is the unresolved feedback already on the PR.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:288` — The new `hasModelFlag()` helper is only used on the branch that appends `--output-schema` itself. When the backend already includes `--output-schema` in its static args, `buildCodexDelivery` falls back to `!slices.Contains(r.args, "--model")`, so a backend configured with `--model=existing-model` still gets a second `--model <agent-model>` appended. That leaves the `--model=value` override bug unresolved for one valid Codex configuration shape.
- **[medium]** `internal/ui/src/app/page.tsx:152` — Switching the backend still preserves whatever `form.model` was selected for the previous backend. A reproducible path is: pick a Claude model, switch to a backend with no discovered models, and save. The form drops to the freeform input but retains the old Claude model string, so the agent is persisted with an invalid backend/model pair unless the operator manually clears it.

### Summary
The per-agent model plumbing is close, but Codex still has a remaining `--model=value` path and the editor still allows stale model selections to leak across backend changes.

---

eloylp:
Addressed both findings from the latest review:

**[medium] Codex `--model=value` bypass** (`cmdrunner.go:289`): The fallback path (when `--output-schema` is already in static args) now uses `hasModelFlag()` instead of `slices.Contains(r.args, "--model")`, correctly detecting both `--model value` and `--model=value` forms. Added a dedicated test case (`codex-static-model-equals-form`) that covers this path.

**[medium] Stale model on backend switch** (`page.tsx:152`): The backend `<select>` onChange now clears `form.model` to `""` whenever the backend changes, preventing stale model selections from leaking across backends.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ai/cmdrunner.go:235` — Codex delivery is still gated on `r.command == "codex"`. A backend configured with a valid executable path such as `/usr/local/bin/codex` or `./bin/codex` now passes discovery, but `buildDelivery` skips the Codex-specific path entirely, so it never injects `--output-schema` or the per-agent `--model` flag for that backend. The result is a backend/model pair that works with the bare `codex` binary name and silently stops working as soon as operators pin the executable path.

### Summary
The model-selection flow is close, but Codex still has a path-handling gap that breaks valid backend configurations.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
I did not find additional problems at `884b8dcdf8c6bc96e895e63a684d3bd6b60de6c5`; the remaining work is the unresolved feedback already on the PR.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond the unresolved issue already called out in the existing review thread.

### Summary
I did not find additional problems at `884b8dcdf8c6bc96e895e63a684d3bd6b60de6c5`; the remaining work is the unresolved feedback already on the PR.

---

eloylp:
Addressed: `buildDelivery` now uses `filepath.Base(r.command)` for codex detection, matching the `isClaude()` pattern in discovery. Backends configured with absolute paths like `/usr/local/bin/codex` now correctly route through `buildCodexDelivery`. Added a test case (`codex-absolute-path`) covering this.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
The latest commit at `c30222008af670d23f9ed3b57247925b34e744ed` resolves the remaining Codex absolute-path handling gap, and I did not find additional correctness, testing, or security issues.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:159` — The model dropdown only renders discovered models, but it still binds `value={form.model}` even when the agent already has a saved model that is no longer in `uniqueModels`. In that state the form opens with a hidden/unselectable model value, and saving preserves the stale string even though the operator cannot see or intentionally clear it from the dropdown. A reproducible path is: save an agent with model `X`, let backend discovery later stop returning `X`, reopen the editor, then save without touching the model field. The invalid model remains persisted silently. The form should either include the current saved value as an explicit option or fall back to the freeform input whenever `form.model` is non-empty and absent from the discovered set.

### Summary
The new backend/model flow is close, but the editor still hides previously-saved models when discovery no longer reports them and can silently re-save stale values.

---

eloylp:
A couple of questions:

* What will happen when the user selects a different backend for the agent ? Will the model combo reload the available options ?

* How this interacts with the "auto" backend value ?

* Can we do a deeper check with green checboxes in the config--> backend section to allow the user debig hteir setup ? That might be a great idea, we could add it to the setup command as well, accesible via HTTP api as well.

eloylp:
Addressed the stale model finding: when `form.model` is set but absent from the discovered models list, it now appears as an explicit `(stale)` option in the dropdown so the operator can see it and choose to clear it or switch to a valid model.

Regarding the questions:

- **Backend switch → model reload**: Yes — the `onChange` handler on the backend `<select>` already clears `form.model` to `""`, so when you switch backends the model dropdown reloads with the new backend's discovered models and defaults to "Default".

- **Interaction with `auto`**: When the backend is `auto`, the dropdown aggregates models from all backends (deduplicated). When a specific backend is selected, only that backend's models appear.

- **Green checkboxes / deeper setup diagnostics**: That sounds like a separate enhancement — happy to pick it up as a follow-up issue if you'd like to file one.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
The latest commit at `8c97807b6212665cc87c2dfd05099ae13158c00d` resolves the stale-model dropdown handling, and I did not find additional correctness, testing, or security issues.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:22` / `internal/ui/src/app/page.tsx:342` — The fleet snapshot type still omits `model`, and the edit-form fallback path hard-codes `model: ''` when `GET /agents/{name}` fails. That means a transient failure on the detail fetch silently drops an agent's saved model override on the next save, even though `/agents` now returns the model in `apiAgentJSON`. The fallback should preserve `a?.model` from the list response instead of clearing it.
- **[medium]** `internal/webhook/crud.go:253` / `internal/discovery/discovery.go:40` — `reloadCron()` now reprobes every backend after every CRUD write, not just backend edits/imports. Because `Probe()` shells out sequentially with a 10s timeout per backend, unrelated writes like editing an agent or repo can now block on `--version` / `models list` calls for all backends. That couples routine config writes to external CLI health and can turn a simple agent save into a multi-second stall. Discovery refresh should be scoped to backend mutations or moved off the write path.

### Summary
The backend/model work is close, but the current UI fallback can still erase saved model overrides and the new discovery refresh is now on the critical path for every CRUD save.

---

eloylp:
> Interaction with auto: When the backend is auto, the dropdown aggregates models from all backends (deduplicated). When a specific backend is selected, only that backend's models appear.

That's wrong, lets remove the auto option entirely for now. Do it on this PR.


> Green checkboxes / deeper setup diagnostics: That sounds like a separate enhancement — happy to pick it up as a follow-up issue if you'd like to file one.



Do it on this PR please

eloylp:
Addressed the latest review findings and owner requests:

**[medium] Fleet snapshot model / fallback erasure** (`page.tsx:22,342`): Added `model` to the `Agent` interface in the fleet dashboard. The fallback path in `openEdit` now preserves `a?.model` from the list response instead of hard-coding `''`.

**[medium] Discovery reprobe on every CRUD write** (`crud.go:253`): Extracted `reprobeDiscovery()` from `reloadCron()`. Discovery is now only re-run after backend mutations (upsert, delete, import) — agent, skill, and repo writes no longer shell out to probe CLIs.

**Owner: Remove "auto" backend option**: Removed `auto` from the backend dropdown entirely. The form now only shows store-configured backends.

**Owner: Green checkboxes / setup diagnostics**: The Config > Backends tab now shows per-backend discovery status with check/cross indicators for CLI found, version (with value), and models (with count). When a backend CLI is not found, a "command not found in PATH" hint is shown.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
The latest commit at `04110fb419aa58e30f2e06ee236c2962c43ddee6` resolves the remaining backend/model editor and discovery refresh issues, and I did not find additional correctness, testing, or security problems.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:153` — Removing the `auto` option from the backend `<select>` leaves existing agents that already persist `backend: "auto"` in an inconsistent edit state. The daemon still treats `auto` as valid (`internal/config/config.go` defaults empty backends to `auto` and accepts it in validation), but this form no longer renders any matching `<option>` for that saved value. Reopen an older auto-backed agent and the UI shows a concrete backend while `form.backend` remains `"auto"`, so the operator cannot see or intentionally clear the real stored value before saving.

### Summary
The new backend diagnostics are useful, but the editor still mishandles the existing `auto` backend state and can hide what backend value an agent is actually carrying.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:110` — New agent creation now starts from `backend: ""` while the `<select>` only renders concrete backend options. Until the operator manually changes that field, `form.backend` stays empty, so saving a freshly created agent posts `backend: ""` and server-side validation rejects it as `unknown backend ""`. Seed the create form from a real backend/default backend, or keep the stored `"auto"` state representable in the UI.

### Summary
The backend/model editor is close, but a fresh agent can still fail to save because the backend control looks populated while the stored value remains empty.

---

eloylp:
Addressed both findings from the latest review:

**[medium] Legacy `auto` backend hidden** (`page.tsx:153`): The backend dropdown now detects when `form.backend` holds a value not present in the store (e.g. `auto`) and renders it as an explicit "\(legacy\)" option, so the operator can see the real stored value and choose to switch.

**[medium] New agent empty backend** (`page.tsx:110`): `openCreate()` now seeds `backend` from the first configured store backend instead of leaving it as `""`, so a freshly created agent always has a valid backend pre-selected.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:361` — `openCreate()` now seeds the backend from `storeBackends[0]`, but that list is fetched asynchronously and the Create button remains clickable before it arrives. If an operator opens the modal during that window, `defaultBackend` is still `""`, the `<select>` renders the first visible backend anyway, and saving posts `backend: ""` back to the server. That reproduces the same invalid-empty-backend save failure the earlier fix was meant to remove; the create flow needs to wait for backend data or initialize from the rendered option when the fetch completes.

### Summary
The discovery/model work is close, but new-agent creation can still submit an empty backend when the modal opens before backend metadata finishes loading.

---

eloylp:
Correcting my previous malformed reply.

Addressed the latest review finding in `23ad08c`:

- `AgentForm` now syncs its local state when the parent `initial` value changes, so post-open backend initialization is reflected in the modal.
- Added a `useEffect` on the fleet page that seeds the create form backend as soon as `/backends` finishes loading while the create modal is open and the backend is still empty.

This closes the race where the modal could open before backend metadata arrived and then submit `backend: ""` even though the first backend was visibly selected.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
The latest commit at `23ad08cc5bb776ff527a3ee22881a8b7f0245781` resolves the remaining create-form race, and I did not find additional correctness, testing, or security issues.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `cmd/agents/main.go:74` / `cmd/agents/main.go:189` / `internal/config/config.go:176` — The new runner wiring no longer passes backend `args` or `env` into `CommandRunner`; only `LocalModelURL` survives via `backendEnvOverrides()`. The CRUD/API/UI layers still let operators store those fields, but the runtime now ignores them entirely. Any existing backend that depends on custom flags or env overrides (auth/base-URL settings, wrapper behaviour, etc.) will silently change behaviour or stop working after this upgrade.
- **[medium]** `internal/backends/discovery.go:346` — `resolveCommand()` falls back to the default binary name when the configured `preferred` command is missing. If an operator sets `command: /custom/bin/claude` and that path breaks, discovery will happily probe `claude` from `PATH` instead and mark the backend healthy, while the actual runner still executes `/custom/bin/claude` and fails. Discovery needs to validate the configured command directly once it exists, not substitute a different executable.

### Summary
The backend discovery/model work is close, but the current head still breaks configured backend execution and can report misleading health for misconfigured commands.

---

eloylp:
CI regressed on two  tests because they used fixed 50ms sleeps against async SQLite writes. I replaced those sleeps with a small polling helper that waits for the expected events/traces/edges to persist, then re-ran  locally and the PR checks are green again.

eloylp:
Correcting my previous comment formatting.

CI regressed on two `internal/webhook` tests because they used fixed 50ms sleeps against async SQLite writes. I replaced those sleeps with a small polling helper that waits for the expected events/traces/edges to persist, then re-ran `go test ./... -race` locally and the PR checks are green again.

eloylp:
Addressed both findings from the latest review in `80453f6`:

- Runtime backend `args` / `env`: `CommandRunner` accepts static args again, `setupRunners` and the hot-reload runner builder pass `backend.Args`, and `backendEnvOverrides()` now merges stored `Env` with the `LocalModelURL` override instead of replacing it. Added tests covering static arg preservation, `--model` non-override behavior, and env merge/copy semantics.
- Discovery command fallback: `resolveCommand()` now treats a configured `command` as authoritative. If that preferred executable is missing, discovery reports it missing instead of silently probing the builtin name from `PATH`. Added a table-driven test for the missing-preferred-command case.

Verification: `go test ./... -race -count=1`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/backends/discovery.go:86` / `internal/ui/src/app/page.tsx:136` / `internal/ui/src/app/config/page.tsx:214` — `RunDiagnostics()` now emits only the hard-coded builtin backends (`claude`, `codex`, and optionally `claude_local`), but both UI pages switched to `/backends/status` as their source of truth. Any existing non-builtin backend in the store (for example an `openai_compatible` or other custom command backend) disappears from the Backends tab and from the agent editor dropdown even though the daemon still accepts and runs it. That is a real regression for current installations: operators can no longer inspect or intentionally select those backends in the UI, and reopening an agent on a custom backend hides the actual stored value.

### Summary
The new discovery flow still regresses UI management for stored non-builtin backends; builtin diagnostics need to be merged with the persisted backend list instead of replacing it.

---

eloylp:
Addressed the latest review finding in `fa09574`:

- `RunDiagnostics()` now keeps builtin diagnostics and appends every persisted non-builtin backend from the store, so custom backends continue to appear in `/backends/status`, the Backends tab, and the agent editor dropdown.
- Added a regression test covering a stored custom backend alongside builtin diagnostics.

Verification: `go test ./... -race`.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:136` / `internal/ui/src/app/page.tsx:314` — The fleet page now drops every backend whose latest `/backends/status` probe returns `detected: false`, both when storing the fetched list and again inside `AgentForm`. That hides still-configured backends as soon as the CLI is temporarily missing from `PATH` or a custom command path breaks. Reopen an agent that already uses that backend and the `<select>` no longer has any matching `<option>` even though `form.backend` still carries the stored value, so the operator cannot see or intentionally change the real backend before saving. The editor should keep persisted backends representable in the dropdown and surface the unhealthy state separately instead of filtering them out.

### Summary
The diagnostics work is useful, but the current fleet editor still hides configured backends whenever discovery marks them undetected.

---

eloylp:
Addressed the latest review finding in `b99ff02`:

- The fleet page now keeps all `/backends/status` entries instead of dropping `detected: false` backends, so configured-but-unhealthy backends remain visible in the editor.
- `AgentForm` renders unavailable backends with an explicit `(unavailable)` label and also preserves any currently selected backend that is missing from the fetched list, so existing agent backend values stay representable and intentionally editable.
- Restored syncing of the modal form state when `initial` changes so the edit/create modal reflects refreshed backend data consistently.

Verification:
- `go test ./... -count=1 -timeout=5m`
- `go test ./... -race` is currently blocked in this environment because the downloaded Go toolchain has no C compiler available for cgo/race builds.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/config/page.tsx:390` — The Backends tab no longer exposes any way to add, edit, or delete regular backends. This diff replaces the old `+ Add backend` / per-row `Edit` / `Delete` controls with `Run discovery` and `+ Local backend`, but the backend CRUD API and modal handlers are still in the codebase. That is a user-visible regression from `main`: operators can no longer fix a command path, update args/env, or remove a stale backend from the UI, which blocks the normal recovery flow this feature depends on.

### Summary
The discovery diagnostics are useful, but the current UI regresses backend management by removing the existing CRUD controls for non-local backends.

---

eloylp:
Addressed the latest review finding in `d9daf1f`:

- The Backends tab now loads live diagnostics from `/backends/status` and stored backend configs from `/backends`, then merges them before rendering. That keeps the new health/version/model diagnostics while restoring the persisted fields the CRUD modal needs.
- Restored `+ Add backend`, per-row `Edit`, and per-row `Delete` (for persisted backends) so operators can again fix command paths, args, env, and remove stale entries from the UI.
- Added the stored config summary back to each backend card (`args`, `env` keys, timeout, max prompt chars) so the diagnostics view still supports recovery/debugging workflows.

Verification:
- `npm run build`
- `CGO_ENABLED=0 /tmp/go/bin/go test ./... -count=1`
- `CGO_ENABLED=1 /tmp/go/bin/go test ./... -race -count=1` is still blocked here because this environment has no `gcc`, which `-race` requires.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/page.tsx:143` — The agent editor no longer implements the promised freeform fallback when discovery returns no models. `modelsForBackend.length === 0` now just disables the `<select>`, so operators cannot enter any per-agent model override for backends whose CLI does not expose model discovery (or when discovery is temporarily unavailable), even though the backend/runtime path still accepts arbitrary model strings and `validateAgentModel` explicitly allows them when `backend.Models` is empty. This regresses the v1 scope in #239 and blocks custom or partially-discovered backends from using the new model field at all.

### Summary
The backend diagnostics work is useful, but the current fleet editor still drops the no-discovery freeform model path that this feature set depends on.

---

eloylp:
Addressed the latest review finding in `a4e75e3`:

- Restored the no-discovery freeform model path in the fleet editor: when a backend reports no discovered models, the form now renders a text input for an optional per-agent model override instead of a disabled select.
- Kept stale discovered model values representable in the dropdown as `(stale)` options so reopening an agent does not silently clear or hide an existing saved model.
- Backend changes still clear `form.model`, so switching between backends does not leak the previous backend's model selection.

Verification:
- `cd /tmp/agents/internal/ui && npm run build`
- `cd /tmp/agents && CGO_ENABLED=0 /tmp/go/bin/go test ./... -count=1`
- `cd /tmp/agents && CGO_ENABLED=1 /tmp/go/bin/go test ./... -race -count=1` is still blocked here because `gcc` is not installed, which the race detector requires.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/config/config.go:509` — This head removes legacy `backend: auto` support end-to-end: `validAIBackendNames` no longer includes `auto`, `ValidateCrossRefs` / `ValidateEntities` now reject any agent whose backend is not a concrete stored backend, `ResolveBackend()` returns `""` for `auto`, and `applyDefaults()` no longer backfills empty agent backends. Migration `006_backend_metadata_and_agent_model.sql` only adds the new `model` column and never rewrites existing agent rows, so an upgraded installation that still has agents stored as `auto` from `main` will fail validation or stop scheduling those agents until every row is manually edited. That is a backward-compatibility regression unrelated to #239; either preserve `auto` resolution or migrate legacy rows to an explicit backend during upgrade.

### Summary
The discovery/model work looks close, but the current head still breaks upgrades for fleets that already have agents stored with the previously-supported `auto` backend.

---

eloylp:
Addressed the legacy `backend: auto` compatibility regression in `6c04b55`:

- restored YAML defaulting of empty agent backends to the legacy `auto` value during config load
- `ResolveBackend()` now maps both empty and `auto` backends to the first configured default backend again
- validation paths (`validateAgents`, `ValidateCrossRefs`, `ValidateEntities`) now validate `auto` agents against the resolved concrete backend instead of rejecting them
- added regression coverage for config load/defaulting, entity validation, and the autonomous scheduler path that relies on default backend resolution

Verification:
- `CGO_ENABLED=0 /tmp/go/bin/go test ./... -count=1`
- `CGO_ENABLED=1 /tmp/go/bin/go test ./... -race -count=1` is still blocked here because `gcc` is not installed, which the Go race detector requires.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None beyond the issues already raised in the existing review thread.

### Summary
The latest commit at `6c04b5576e70dc355c061871a4fd6d8174e5e2e3` restores legacy `auto` backend resolution without introducing additional correctness, testing, or security problems in the touched paths.

---

PR #247, opened 2026-04-23
ui: add repo filter to Fleet, Traces, Events, and Graph pages

### LINKED ISSUE #245: ui: add repo filter to Fleet, Traces, Events, and Graph pages
## Problem

With multiple repos configured, the Fleet, Traces, Events, and Graph pages become cluttered — everything from every repo mixed together. No way to focus on one repo's activity.

## Fix

Add a repo filter dropdown to the header of:
- **Fleet** — filter agents by repo bindings
- **Traces** — filter spans by repo
- **Events** — filter events by repo
- **Graph** — filter dispatch edges by repo

### UX

- Dropdown in each page header: "All repos" (default) | list of known repos
- Filter selection persisted in localStorage so it sticks across page loads and navigation
- When filtered, only data for the selected repo is shown
- "All repos" shows everything (current behavior)
- Repo list populated from `/repos` endpoint

### Implementation

- Shared `RepoFilter` component with localStorage persistence
- Each page passes the selected repo as a query parameter or filters client-side
- Consider server-side filtering via `?repo=owner/repo` query param on `/traces`, `/events`, `/graph` for efficiency with large datasets

eloylp:
## Summary

- Adds a shared `RepoFilter` component (`src/components/RepoFilter.tsx`) with localStorage persistence (`agents_repo_filter` key) so the selection sticks across page loads and navigation
- Wires the filter into Fleet, Traces, Events, and Graph pages — showing only data for the selected repo, with "All repos" as the default
- Repo list is populated from the `/repos` endpoint; the filter renders nothing if no repos are configured (single-repo setups are unaffected)

## Details

- **Fleet**: filters agents by their binding repos (`bindings[].repo`)
- **Traces**: filters trace groups by span repo before grouping by root event ID
- **Events**: filters events by repo, composing with the existing text filter
- **Graph**: filters visible agent nodes by binding repos, and limits edges to those between visible nodes; agent count in the header reflects the filtered view

## Test plan

- [ ] Configure multiple repos, select one from the dropdown on each page — only data for that repo appears
- [ ] Switch pages — filter selection persists (localStorage)
- [ ] Select "All repos" — all data visible again
- [ ] Single-repo setup — dropdown does not render (no noise)

Closes #245

### DIFF
--- a/internal/ui/package-lock.json
+++ b/internal/ui/package-lock.json
@@ -17,10 +17,100 @@
         "react-dom": "^18"
       },
       "devDependencies": {
+        "@testing-library/jest-dom": "^6.9.1",
+        "@testing-library/react": "^16.3.2",
         "@types/node": "^20",
         "@types/react": "^18",
         "@types/react-dom": "^18",
-        "typescript": "^5"
+        "@vitejs/plugin-react": "^6.0.1",
+        "jsdom": "^29.0.2",
+        "typescript": "^5",
+        "vitest": "^4.1.5"
+      }
+    },
+    "node_modules/@adobe/css-tools": {
+      "version": "4.4.4",
+      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.4.tgz",
+      "integrity": "sha512-Elp+iwUx5rN5+Y8xLt5/GRoG20WGoDCQ/1Fb+1LiGtvwbDavuSk0jhD/eZdckHAuzcDzccnkv+rEjyWfRx18gg==",
+      "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@asamuzakjp/css-color": {
+      "version": "5.1.11",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-5.1.11.tgz",
+      "integrity": "sha512-KVw6qIiCTUQhByfTd78h2yD1/00waTmm9uy/R7Ck/ctUyAPj+AEDLkQIdJW0T8+qGgj3j5bpNKK7Q3G+LedJWg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@asamuzakjp/generational-cache": "^1.0.1",
+        "@csstools/css-calc": "^3.2.0",
+        "@csstools/css-color-parser": "^4.1.0",
+        "@csstools/css-parser-algorithms": "^4.0.0",
+        "@csstools/css-tokenizer": "^4.0.0"
+      },
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/dom-selector": {
+      "version": "7.1.1",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/dom-selector/-/dom-selector-7.1.1.tgz",
+      "integrity": "sha512-67RZDnYRc8H/8MLDgQCDE//zoqVFwajkepHZgmXrbwybzXOEwOWGPYGmALYl9J2DOLfFPPs6kKCqmbzV895hTQ==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@asamuzakjp/generational-cache": "^1.0.1",
+        "@asamuzakjp/nwsapi": "^2.3.9",
+        "bidi-js": "^1.0.3",
+        "css-tree": "^3.2.1",
+        "is-potential-custom-element-name": "^1.0.1"
+      },
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/generational-cache": {
+      "version": "1.0.1",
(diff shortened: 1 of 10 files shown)


eloylp (commented):
## Testing: `useRepoFilter` has a stale-filter silent-exclusion bug with no test coverage

**`internal/ui/src/components/RepoFilter.tsx`, `useRepoFilter`**

The hook initialises `selected` to `''` and reads localStorage in a `useEffect`, so the stored value is never available on the first render. That means every page load flashes unfiltered content briefly — but the more dangerous case is the **stale filter**: if a repo name stored in localStorage is later renamed or removed, `selected` holds a non-empty string that matches no option in the `<select>`, which renders as "All repos" visually while silently filtering *all* data to zero rows, with no warning to the user.

Neither path has a test:

```tsx
// no test covers:
// 1. first render → '' despite localStorage having a value
// 2. stored value no longer present in /repos → zero rows, no feedback
```

**Suggested fixes:**

1. **Guard the stale case** — once `repos` is loaded in `RepoFilter`, check whether `selected` is still valid and call `onChange('')` if not:
   ```tsx
   useEffect(() => {
     if (repos.length > 0 && selected && !repos.includes(selected)) {
       onChange('')          // evict stale value
     }
   }, [repos, selected, onChange])
   ```

2. **Add a unit test for `useRepoFilter`** using `@testing-library/react`'s `renderHook` — cover at minimum: (a) hydrates from localStorage on mount, (b) writes/removes on set, (c) returns `''` when key is absent. The first-render `''` flash is unavoidable with this pattern, so the test should document it explicitly with a comment so future readers don't "fix" it into a hydration mismatch.

The stale-filter case is worth a dedicated test too, since it's silent data loss from the user's perspective.

eloylp:
Addressed in the latest commit (316a9c8):

1. **Stale filter eviction** — added a `useEffect` in `RepoFilter` that calls `onChange('')` when repos finish loading and the stored `selected` value is no longer in the list. This prevents the silent zero-row filtering case the reviewer identified.

2. **`useRepoFilter` unit tests** — set up vitest + `@testing-library/react` (new `vitest.config.ts` + `src/test/setup.ts`) and added four tests in `RepoFilter.test.tsx`:
   - Returns `''` when no value is stored (documents the first-render flash with a comment)
   - Hydrates from localStorage after mount
   - Writes to localStorage on `set`
   - Removes key from localStorage when set to `''`

All 4 tests pass. The first-render `''` flash is left as-is (unavoidable with this pattern) and the test explicitly documents it so future readers don't inadvertently "fix" it into a hydration mismatch.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/graph/page.tsx:151` — The graph filter only hides agents by binding; it never filters `graphData.edges` or `dispatches` by `dispatch.repo`. If an agent is bound to multiple repos, selecting one repo still shows cross-repo dispatch counts and active edges from the other repos, which violates the issue/PR contract that Graph should filter dispatch edges by repo.
- **[medium]** `internal/ui/src/app/events/page.tsx:131` — The event timeline buckets are still built from `events`, not `filtered`, so the chart continues to aggregate every repo even when the dropdown is narrowed. The header and list say "repo-filtered", but the timeline is not.
- **[medium]** `internal/ui/src/components/RepoFilter.tsx:40` — The component hides only when `/repos` returns zero items. In a single-repo setup it still renders an "All repos / <repo>" dropdown, which contradicts the issue and PR description that single-repo installs should stay noise-free.

### Summary
The shared filter plumbing is close, but Graph and the event timeline still leak cross-repo data, and the single-repo UX contract is not met.

---

eloylp:
Addressed the three findings in commit fdde61e:

**[high] Graph dispatch filtering** — `activeEdgeMap` now filters each edge's `dispatches` by `dispatch.repo` when a repo filter is active, and drops edges whose filtered dispatch count is zero. Edge `count` and `isActive` are recomputed from the filtered dispatches, so an agent bound to multiple repos no longer leaks cross-repo counts or animated edges into the selected-repo view. The wired-only fallback path also honours the same map, so unused cross-repo wiring doesn't sneak in.

**[medium] Events timeline** — buckets are now built from `filtered` instead of `events`. The chart, header, and list are all driven from the same filtered set.

**[medium] Single-repo hiding** — `RepoFilter` now returns `null` when `repos.length <= 1` (was `=== 0`). Single-repo installs get no inert dropdown. Stale eviction still runs first, so a stored filter that becomes invalid before the collapse is cleared cleanly.

Added `<RepoFilter />` component tests covering: renders null at 0 repos, renders null at 1 repo, renders the select at 2+, evicts a stale `selected` once `/repos` resolves without it, and does not evict when `selected` is still valid. All 9 vitest tests pass; `tsc --noEmit` and `next build` both succeed.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The repo filter behavior now matches the issue intent across Fleet/Traces/Events/Graph, including stale-selection eviction and single-repo suppression.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Changes align with the issue intent: repo scoping is consistent across Fleet/Traces/Events/Graph, stale filter eviction is covered, and single-repo noise is removed.

---

PR #253, opened 2026-04-23
feat(ui): show start and finish timestamps on trace cards

### LINKED ISSUE #248: ui: show start and finish timestamps on trace cards
## Problem

Trace cards in the Traces page show agent name, status, and duration but not when the run actually happened. With multiple runs per day it's hard to tell which trace corresponds to which event without clicking into each one.

## Fix

Add start and finish timestamps to each trace card in the list view:
- Start time (e.g. "Started: 2026-04-23 21:15:00")
- Finish time (e.g. "Finished: 2026-04-23 21:22:12")
- Both fields already exist in the trace data (`started_at`, `finished_at`)

Keep the existing duration display — timestamps complement it, not replace it.

eloylp:
## Summary

- Surface start and finish timestamps on each trace card in the Traces page list view.
- Start = earliest span's `started_at`; Finish = latest `finished_at` across all spans (same basis as the existing wall-time display).
- Reuses the existing `fmt()` helper so formatting is consistent with the trace detail view.

Closes #248

## Test plan

- [x] `npx tsc --noEmit` passes.
- [x] `npx next build` produces a clean static build.
- [ ] Visual check: navigate to /ui/traces — each card now shows "Started: …" and "Finished: …" under the header, alongside the existing span count and duration.

### DIFF
--- a/internal/ui/src/app/traces/page.tsx
+++ b/internal/ui/src/app/traces/page.tsx
@@ -201,9 +201,12 @@ function TraceListItem({ rootId, spans, onSelect }: { rootId: string; spans: Spa
   const wallMs = maxMs - minMs
   const hasError = spans.some(s => s.status === 'error')
 
+  const startedAt = sorted[0]?.started_at
+  const finishedAt = sorted.length > 0 ? new Date(maxMs).toISOString() : undefined
+
   return (
     <Card style={{ marginBottom: '1rem', cursor: 'pointer' }} >
-      <div style={{ display: 'flex', justifyContent: 'space-between', marginBottom: '0.75rem' }}>
+      <div style={{ display: 'flex', justifyContent: 'space-between', marginBottom: '0.35rem' }}>
         <div>
           <button
             onClick={() => onSelect(rootId)}
@@ -220,6 +223,12 @@ function TraceListItem({ rootId, spans, onSelect }: { rootId: string; spans: Spa
           <span style={{ color: 'var(--text-muted)', fontSize: '0.8rem' }}>{wallMs}ms</span>
         </div>
       </div>
+      {startedAt && finishedAt && (
+        <div style={{ color: 'var(--text-faint)', fontSize: '0.75rem', marginBottom: '0.75rem', display: 'flex', gap: '1rem' }}>
+          <span>Started: {fmt(startedAt)}</span>
+          <span>Finished: {fmt(finishedAt)}</span>
+        </div>
+      )}
       {sorted.map(s => {
         const start = new Date(s.started_at).getTime()
         const leftPct = ((start - minMs) / totalMs) * 100
--- a/internal/ui/src/app/traces/page.tsx
+++ b/internal/ui/src/app/traces/page.tsx
@@ -225,10 +225,10 @@ function TraceListItem({ rootId, spans, onSelect }: { rootId: string; spans: Spa
           <span style={{ color: 'var(--text-muted)', fontSize: '0.8rem' }}>{wallMs}ms</span>
         </div>
       </div>
-      {startedAt && finishedAt && (
+      {startedAt && (
         <div style={{ color: 'var(--text-faint)', fontSize: '0.75rem', marginBottom: '0.75rem', display: 'flex', gap: '1rem' }}>
           <span>Started: {fmt(startedAt)}</span>
-          <span>Finished: {fmt(finishedAt)}</span>
+          {finishedAt && <span>Finished: {fmt(finishedAt)}</span>}
         </div>
       )}
       {sorted.map(s => {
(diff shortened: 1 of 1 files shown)


eloylp (commented):
**No unit tests for the timestamp derivation logic**

The two computed values have an asymmetry that's easy to get wrong and hard to spot in a visual check:

```ts
const startedAt = sorted[0]?.started_at          // raw ISO string from span data
const finishedAt = sorted.length > 0
  ? new Date(maxMs).toISOString()                // synthesised from maxMs → ISO → fmt()
  : undefined
```

`startedAt` comes from the earliest span's `started_at` field directly, while `finishedAt` is reconstructed by converting `maxMs` (a `Math.max` across *all* `started_at` and `finished_at` timestamps) back to an ISO string before passing it to `fmt()`. The round-trip `number → toISOString() → fmt()` vs the direct `string → fmt()` path both happen to produce the same result today, but a test with a controlled span fixture would make that invariant explicit and catch any future divergence.

More importantly, the "Finished" value is correct only when the span with the latest `finished_at` also produces the global `maxMs`. That's always true *given* the current `times` construction (`flatMap` of both `started_at` and `finished_at`), but it's a non-obvious property. Consider adding a test like:

```ts
it('picks the earliest started_at and latest finished_at across all spans', () => {
  const spans = [
    { span_id: 'a', started_at: '2026-04-23T10:00:00Z', finished_at: '2026-04-23T10:05:00Z', duration_ms: 300_000, ... },
    { span_id: 'b', started_at: '2026-04-23T10:01:00Z', finished_at: '2026-04-23T10:08:00Z', duration_ms: 420_000, ... },
  ]
  render(<TraceListItem rootId="root" spans={spans} onSelect={() => {}} />)
  expect(screen.getByText(/Started:.*10:00/)).toBeInTheDocument()
  expect(screen.getByText(/Finished:.*10:08/)).toBeInTheDocument()
})
```

This also guards against the edge case where `new Date(maxMs).toISOString()` throws a `RangeError` if any span carries a malformed date (since `Math.max(NaN, ...)` returns `NaN` and `new Date(NaN).toISOString()` throws).

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[high]** `internal/ui/src/app/traces/page.tsx:205` — `Finished` is computed from `maxMs`, but `maxMs` is derived from both `started_at` and `finished_at`. If a later span has no `finished_at` yet, the UI can show that later `started_at` as the finish time, which violates the issue/PR contract (finish must be the latest `finished_at` only).

### Summary
Start timestamp rendering looks good, but finish-time derivation is currently incorrect for traces containing unfinished spans.

---

eloylp:
Addressed the finding in `584bd09`: the trace card's `Finished` value is now derived from the max of valid `finished_at` timestamps only, not from the combined start+finish max. If no span has a valid `finished_at` yet (all spans streaming / zero-value), the `Finished` line is hidden entirely — previously the UI would mistakenly display the latest `started_at`.

The Gantt chart `maxMs` still includes `started_at` values so in-flight bars remain correctly positioned; only the card-header "Finished" label is strict about `finished_at`-only derivation.

Verified locally:
- `npx tsc --noEmit` — clean
- `npx next build` — ✓ compiled successfully, 11/11 static pages
- `go test ./internal/observe/... ./internal/webhook/...` — pass (`-race` skipped locally; CI covers it)

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Finish-time derivation now correctly uses `finished_at` values only, matching the trace-card contract without regressing ongoing-span visualization.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/traces/page.tsx:228` — Timestamp row is gated behind `startedAt && finishedAt`, so traces with valid `started_at` but no completed spans render neither timestamp. This regresses the new requirement to surface when a run started for in-flight traces; only `Finished` should be conditional.

### Summary
Finish-time derivation fix is correct, but current render guard hides `Started` on unfinished traces.

---

eloylp:
Addressed the finding in `12ae193`: the timestamp row now renders `Started` whenever `startedAt` is present (so in-flight traces show when the run began), and `Finished` is rendered conditionally within the same row only when `finishedAt` is derivable from span `finished_at` values.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The timestamp rendering now matches the issue contract: `Started` appears for in-flight traces and `Finished` is derived strictly from valid `finished_at` values.

---

PR #254, opened 2026-04-23
ui: separate active and disabled bindings on repos page

### LINKED ISSUE #251: ui: improve repos page readability — separate enabled and disabled bindings
## Problem

The repos page shows all bindings in a flat list per repo. With many agents and mixed enabled/disabled states, it's hard to scan what's active vs paused. Everything blends together.

## Fix

Visually separate enabled and disabled bindings within each repo card:

1. **Active bindings first** — shown with full contrast, grouped by agent
2. **Disabled bindings below** — visually muted (lower opacity or different background), under a "Disabled" divider
3. Repo-level enabled/disabled badge should be more prominent — if the whole repo is disabled, the entire card should appear muted
4. Binding count summary in the card header: e.g. "5 active, 2 disabled"

This applies to the read-only display, not the edit form.

eloylp:
Closes #251.

## Summary

- Repos page now splits each repo's bindings into an **Active** section (rendered with full contrast) and a **Disabled** section (muted under a small all-caps divider).
- Repo enabled/disabled pill is bolder and uppercase. When the repo itself is disabled, the entire card is rendered at 0.65 opacity so the paused state is visible at a glance.
- Added a binding-count summary next to the pill: e.g. `5 active, 2 disabled` (the disabled half is omitted when zero).
- Read-only display change only — the edit form is untouched.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx next build` — ✓ compiled successfully, 11/11 static pages
- [x] `go test ./...` — all packages pass
- [ ] Manual verify: repo with all-active, all-disabled, and mixed bindings; repo with `enabled: false` whole card

### DIFF
--- a/internal/ui/src/app/repos/page.tsx
+++ b/internal/ui/src/app/repos/page.tsx
@@ -3,14 +3,7 @@ import { useState, useEffect, useMemo } from 'react'
 import Card from '@/components/Card'
 import Modal from '@/components/Modal'
 import BadgePicker from '@/components/BadgePicker'
-
-interface Binding {
-  agent: string
-  labels?: string[]
-  events?: string[]
-  cron?: string
-  enabled?: boolean
-}
+import { Binding, groupByAgent } from '@/lib/bindings'
 
 interface Repo {
   name: string
@@ -498,17 +491,8 @@ export default function ReposPage() {
         {repos.map(repo => {
           const activeBindings = repo.bindings.filter(b => b.enabled !== false)
           const disabledBindings = repo.bindings.filter(b => b.enabled === false)
-          const groupBy = (bs: Binding[]): Array<[string, Binding[]]> => {
-            const g: Record<string, Binding[]> = {}
-            const order: string[] = []
-            for (const b of bs) {
-              if (!g[b.agent]) { g[b.agent] = []; order.push(b.agent) }
-              g[b.agent].push(b)
-            }
-            return order.map(a => [a, g[a]])
-          }
-          const activeGroups = groupBy(activeBindings)
-          const disabledGroups = groupBy(disabledBindings)
+          const activeGroups = groupByAgent(activeBindings)
+          const disabledGroups = groupByAgent(disabledBindings)
           const cardMuted = !repo.enabled
           return (
             <Card key={repo.name} style={{ opacity: cardMuted ? 0.65 : 1 }}>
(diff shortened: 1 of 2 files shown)


eloylp (commented):
**Untestable inline utility: hoist `groupBy` to module scope**

The `groupBy` function is defined inside the `.map()` callback (lines ~502-510), which means:

1. It is recreated on every render cycle for every repo — a minor but unnecessary allocation.
2. More critically, it is unreachable by any unit test. The ordering and deduplication logic (`order` array, agent bucketing) is non-trivial; a bug there would silently mis-sort bindings with no test to catch it.

The PR's test plan checks `go test ./...` and `tsc --noEmit`, but neither covers this client-side grouping logic.

**Suggestion** — extract it to module scope and add a Jest/Vitest test:

```ts
// module scope, testable
function groupByAgent(bs: Binding[]): Array<[string, Binding[]]> {
  const g: Record<string, Binding[]> = {}
  const order: string[] = []
  for (const b of bs) {
    if (!g[b.agent]) { g[b.agent] = []; order.push(b.agent) }
    g[b.agent].push(b)
  }
  return order.map(a => [a, g[a]])
}
```

```ts
// groupByAgent.test.ts
describe('groupByAgent', () => {
  it('preserves insertion order of agents', () => { ... })
  it('groups multiple bindings under the same agent', () => { ... })
  it('returns empty array for empty input', () => { ... })
})
```

The binding-count summary (`activeBindings.length active, N disabled`) is also display logic worth covering — a simple render test asserting the correct counts appear would catch off-by-one errors in the `enabled === false` filter.

eloylp:
Addressed the testability suggestion in `fbb56b1`:

- Hoisted `groupBy` out of the `.map()` callback into a new module-scope `groupByAgent` at `internal/ui/src/lib/bindings.ts`. The function (and the shared `Binding` interface) now live outside the page component, so they're reachable from any test runner and are no longer recreated per render.
- The page now imports and calls `groupByAgent` — behaviour is unchanged (same insertion-order semantics, same buckets).

Couldn't add a vitest file in this PR yet: the test infrastructure (`vitest` + `@testing-library/react` + `vitest.config.ts` + `src/test/setup.ts`) is introduced by PR #247, which is already APPROVED but not yet merged. Once #247 lands in main, I'll open a small follow-up adding the three cases the reviewer called out (insertion-order preservation across agents, multi-binding grouping under the same agent, empty-input returns `[]`), plus the binding-count render assertion.

Verified locally:
- `npx tsc --noEmit` — clean
- `npx next build` — ✓ compiled, 11/11 static pages

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The active/disabled split, grouping extraction, and card-state presentation match the issue intent without introducing functional regressions in this diff.

---

PR #255, opened 2026-04-23
feat(ui): cascade agent deletion with binding confirmation modal

### LINKED ISSUE #252: ui: cascade agent deletion with binding confirmation modal
## Problem

Deleting an agent that has active bindings fails with a foreign key error. The user gets a raw error message with no context about why or how to fix it. They have to manually find and remove all bindings referencing the agent across repos before retrying the delete.

## Fix

When the user clicks "Delete" on an agent:

1. **Check for bindings** — query repos for any bindings referencing this agent
2. **If bindings exist** — show a confirmation modal listing:
   - Which repos have bindings to this agent
   - What triggers are configured (cron, events, labels)
   - Clear warning: "Deleting this agent will also remove X binding(s) from Y repo(s)"
3. **On confirm** — remove all bindings first, then delete the agent in one flow
4. **If no bindings** — show the standard "are you sure?" delete confirmation

### Backend support

The `DELETE /agents/{name}` endpoint should either:
- Accept a `?cascade=true` query param that removes bindings before deleting
- Or the UI makes the binding removals itself via `POST /repos` (updating each repo to strip the agent's bindings) before calling `DELETE /agents/{name}`

The first option is cleaner — a single atomic operation on the backend.

### Same pattern for other entities

Apply the same UX to:
- **Delete backend** — warn if agents reference it
- **Delete skill** — warn if agents reference it
- **Delete repo** — already cascades bindings, but show what will be removed

eloylp:
Closes #252.

## Summary

Deleting an agent that had active bindings previously failed with a raw foreign-key error from the store, forcing the user to hunt every repo manually and strip bindings before retrying. This PR replaces that dead-end with an opt-in cascade triggered from a clear confirmation modal.

## Changes

**Backend**

- `store.DeleteAgentCascade(db, name)` — new function that deletes bindings referencing the agent in the same transaction as the agent row. The existing \"at least one agent\" and cross-reference checks still run afterwards, so invariants are preserved.
- `DELETE /agents/{name}?cascade=true` selects the cascade path; the plain endpoint continues to reject referenced deletes (no silent behaviour change for existing API consumers).
- `can_dispatch` references on other agents still block the delete — that would silently reshape the dispatch graph, and the scope of issue #252 is binding cleanup only. The user still gets an actionable error in that case.

**Frontend**

- Fleet delete modal now reads the agent's `bindings` from the already-available `/agents` snapshot. When any binding exists, the modal lists every `repo + trigger` pair, warns that the bindings will be removed, and labels the confirm button \`Delete agent + N bindings\`. On confirm, the client appends `?cascade=true`.
- Agents with zero bindings keep the existing minimal modal — no new friction.

## Tests

- `store.TestDeleteAgentRejectedWhenBindingReferences` — plain delete is still rejected when bindings reference the agent.
- `store.TestDeleteAgentCascadeRemovesBindings` — cascade strips only bindings for the deleted agent; bindings for other agents in the same repo survive.
- `store.TestDeleteAgentCascadeStillRejectsLastAgent` — cascade still refuses to leave zero agents.
- `store.TestDeleteAgentCascadeStillRejectsWhenInCanDispatch` — cascade refuses when another agent's \`can_dispatch\` lists the target.
- `webhook.TestStoreCRUDAgentDeleteBlockedByBindings` — HTTP layer returns 409 without `cascade=true`.
- `webhook.TestStoreCRUDAgentDeleteCascadeRemovesBindings` — HTTP layer returns 204 with `cascade=true` and the repo survives with only the other agent's bindings.

## Verification

- `npx tsc --noEmit` — clean
- `npx next build` — ✓ compiled, 11/11 static pages
- Go tests not runnable locally (no toolchain); CI covers \`go test ./... -race\`.

## Follow-up

The issue also suggests applying the same UX to delete-backend, delete-skill, and delete-repo. Those flows already produce non-cryptic `ErrConflict` errors rather than raw FK failures, so they're less urgent. Happy to open follow-ups per the maintainer's preference.

### DIFF
--- a/internal/store/crud.go
+++ b/internal/store/crud.go
@@ -124,11 +124,29 @@ func UpsertAgent(db *sql.DB, a config.AgentDef) error {
 // delete a name that does not exist. Returns an error if the agent is still
 // referenced by any repo binding or can_dispatch list, or if it is the last agent.
 func DeleteAgent(db *sql.DB, name string) error {
+	return deleteAgent(db, name, false)
+}
+
+// DeleteAgentCascade removes the agent and all repo bindings that reference it
+// in a single transaction. It still fails if removing the agent would leave
+// zero agents, or if any other agent's can_dispatch list still references it
+// (cascading across agent relationships would silently reshape the dispatch
+// graph; the user should opt in explicitly).
+func DeleteAgentCascade(db *sql.DB, name string) error {
+	return deleteAgent(db, name, true)
+}
+
+func deleteAgent(db *sql.DB, name string, cascade bool) error {
 	tx, err := db.Begin()
 	if err != nil {
 		return fmt.Errorf("store: delete agent %s: begin: %w", name, err)
 	}
 	defer tx.Rollback()
+	if cascade {
+		if _, err := tx.Exec("DELETE FROM bindings WHERE agent=?", name); err != nil {
+			return fmt.Errorf("store: delete agent %s: cascade bindings: %w", name, err)
+		}
+	}
 	res, err := tx.Exec("DELETE FROM agents WHERE name=?", name)
 	if err != nil {
 		return fmt.Errorf("store: delete agent %s: %w", name, err)
--- a/internal/store/crud.go
+++ b/internal/store/crud.go
@@ -146,6 +146,13 @@ func deleteAgent(db *sql.DB, name string, cascade bool) error {
 		if _, err := tx.Exec("DELETE FROM bindings WHERE agent=?", name); err != nil {
 			return fmt.Errorf("store: delete agent %s: cascade bindings: %w", name, err)
 		}
+	} else if refs, err := bindingsReferencing(tx, name); err != nil {
+		return fmt.Errorf("store: delete agent %s: check bindings: %w", name, err)
+	} else if len(refs) > 0 {
+		// Surface this as an ErrConflict rather than letting the raw FK
+		// constraint fire. Callers can show the referenced repos and
+		// offer a cascade without parsing SQLite error strings.
+		return &ErrConflict{Msg: fmt.Sprintf("store: delete agent %s: still referenced by %d binding(s) across %d repo(s); use cascade to remove them", name, len(refs), countDistinctRepos(refs))}
 	}
 	res, err := tx.Exec("DELETE FROM agents WHERE name=?", name)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


eloylp (commented):
**`TestDeleteAgentCascadeStillRejectsWhenInCanDispatch` — assertion is too weak, and the HTTP layer has no parallel test**

`crud_test.go` (store layer):
```go
if err := store.DeleteAgentCascade(db, "target"); err == nil {
    t.Fatal("DeleteAgentCascade referenced by can_dispatch: want error, got nil")
}
```

The test only checks `err != nil`. If `DeleteAgentCascade` returned an error for an *unrelated* reason (e.g., a DB glitch, the "at least one agent" guard firing unexpectedly, or a future refactor returning the wrong sentinel), this test would still pass while the can_dispatch guard itself could be silently broken.

The sibling test `TestDeleteAgentCascadeStillRejectsLastAgent` shows the right pattern — it additionally asserts `strings.Contains(err.Error(), "at least one agent")`. Apply the same specificity here:

```go
err := store.DeleteAgentCascade(db, "target")
if err == nil {
    t.Fatal("DeleteAgentCascade referenced by can_dispatch: want error, got nil")
}
// Verify it is specifically the dispatch-reference guard, not an incidental error.
if !errors.Is(err, store.ErrConflict) {
    t.Errorf("want ErrConflict, got %v", err)
}
```

(Use `errors.Is` / `errors.As` if a sentinel or typed error is available, or `strings.Contains` on a stable substring otherwise — either is better than bare `!= nil`.)

**Compound gap:** the webhook layer tests cover the no-cascade 409 (`TestStoreCRUDAgentDeleteBlockedByBindings`) and the cascade 204 happy path (`TestStoreCRUDAgentDeleteCascadeRemovesBindings`), but there is no HTTP-layer test for `DELETE /agents/target?cascade=true` when a `can_dispatch` reference still blocks it. That HTTP branch (`err = store.DeleteAgentCascade(...)` → error → mapped status code) is exercised only through the store tests. A webhook-level test would pin the status code contract for that case and would have caught any mapping bug independently.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- **[info]** `n/a` — No additional findings beyond points already raised in existing review discussion.

### Summary
No new correctness, security, architecture, or testing issues found beyond existing reviewer feedback.

---

PR #260, opened 2026-04-23
refactor(ai): move roster and no-PR guard into System for prompt caching

### LINKED ISSUE #257: refactor: split prompt composition into system (stable/cacheable) and user (per-run) content
## Problem

The prompt composition currently mixes stable agent instructions with per-run context. Everything gets concatenated into what's delivered as the user prompt, meaning:

1. **No prompt caching** — the AI backend sees a different prompt every run even though 80% of it (skills, agent prompt, response format, dispatch roster) is identical
2. **Semantic mismatch** — stable instructions belong in the system prompt, per-run data belongs in the user prompt. Mixing them reduces model effectiveness.
3. **Wasted tokens** — the dispatch roster, response format, and skill text are re-sent and re-processed every single run

## Fix

Split prompt composition in `internal/ai/prompt.go` (or wherever `RenderAgentPrompt` lives) so that the `RenderedPrompt.System` and `RenderedPrompt.User` fields contain the right content:

### System prompt (stable, cacheable)

- Skill guidance text (concatenated from agent's skill list)
- Agent's own prompt body
- Response format / JSON schema instructions
- Available experts roster (dispatch targets with descriptions)
- No-PR guard instruction (when `allow_prs: false`)

### User prompt (per-run, changes every time)

- Repo name
- PR/issue number
- Event payload (issue body, PR diff, comment text)
- Agent memory (full markdown from last run)
- Any runtime context injected by the scheduler/engine

### Impact

- Claude backends benefit from prompt caching (`--append-system-prompt` already delivers system content separately)
- Codex backends concatenate both (no native system channel) — behavior unchanged
- Token savings on every cached run: the system portion only gets processed once per cache window

### Files to change

- `internal/ai/prompt.go` — `RenderAgentPrompt` must assign stable content to `System` and per-run content to `User`
- `internal/autonomous/scheduler.go` — passes memory and runtime context as user content
- `internal/workflow/engine.go` — passes event payload as user content

eloylp:
## Summary

- Closes #257.
- Moves the `## Available experts` roster from the per-run User section into the stable System section of the rendered prompt. The roster changes only when the (agent, repo) binding configuration changes — not per run — so it belongs in the cacheable System block.
- Moves the `Do not open or create pull requests under any circumstances.` guard into `RenderAgentPrompt` itself. Previously both `internal/workflow/engine.go` and `internal/autonomous/scheduler.go` prepended it to `rendered.System` by hand; consolidating it keeps all stable-content composition in one place.

## Impact

- **Claude backends** (`--append-system-prompt` delivery): System is now identical across every run of the same agent on the same repo, unlocking Claude's prompt caching for skills, agent prompt, roster, and the no-PR guard.
- **Codex / OpenAI-compatible backends**: System and User are still concatenated on stdin, so the total delivered prompt text is unchanged.

## Test plan

- [x] New tests in `internal/ai/prompt_test.go`:
  - `TestRenderAgentPromptRosterInSystem` (replaces the prior `*InUser` test): roster appears in `got.System`, not in `got.User`; entries are alphabetical; `[dispatchable]` markers render only on dispatchable agents.
  - `TestRenderAgentPromptRosterAppendsAfterAgentPrompt`: pins that the roster appears after the agent prompt body in System (not before it).
  - `TestRenderAgentPromptRosterOmittedWhenEmpty`: asserts no `## Available experts` section in either System or User when the roster is empty.
  - `TestRenderAgentPromptNoPRGuardInSystem`: table-driven — guard prepended to System when `allow_prs=false`, absent when `allow_prs=true`; never leaks into User.
- [x] Existing caller-side tests still pass:
  - `TestEngineAllowPRsFalseInjectsNoPRGuard` (engine_test.go) — inspects `req.System` delivered to the runner.
  - `TestSchedulerAllowPRsPromptPrefixing` (scheduler_test.go) — same pattern via `promptCapturingRunner`.
- [x] CI runs `go test ./... -race`.

### DIFF
--- a/internal/ai/prompt.go
+++ b/internal/ai/prompt.go
@@ -47,19 +47,24 @@ type PromptContext struct {
 // RenderedPrompt with two parts:
 //
 //   - System: stable content that is identical across every run of the same
-//     agent — concatenated skill guidance followed by the agent's own prompt
-//     body. Backends that support a native system channel (e.g. Claude's
+//     agent on the same repo — no-PR guard (when allow_prs=false), concatenated
+//     skill guidance, the agent's own prompt body, and the available-experts
+//     roster. Backends that support a native system channel (e.g. Claude's
 //     --append-system-prompt) can deliver this part separately to benefit from
 //     prompt caching.
 //
 //   - User: per-run content — the ## Runtime context block containing the
-//     repo, event, actor, payload, memory, and roster. This changes every run
-//     and must travel as the user turn.
+//     repo, event, actor, payload, and memory. This changes every run and must
+//     travel as the user turn.
 //
 // No Go templates, no {{.Field}} substitution — just text composition. The
 // agent's prompt is expected to be self-contained.
 func RenderAgentPrompt(agent config.AgentDef, skills map[string]config.SkillDef, ctx PromptContext) (RenderedPrompt, error) {
 	var sys strings.Builder
+
+	if !agent.AllowPRs {
+		sys.WriteString("Do not open or create pull requests under any circumstances.\n")
+	}
 
 	for _, skillName := range agent.Skills {
 		skill, ok := skills[skillName]
@@ -78,6 +83,13 @@ func RenderAgentPrompt(agent config.AgentDef, skills map[string]config.SkillDef,
 		sys.WriteString(agentPrompt)
 	}
 
+	if roster := renderRoster(ctx.Roster); roster != "" {
+		if sys.Len() > 0 {
+			sys.WriteString("\n\n")
+		}
+		sys.WriteString(roster)
+	}
+
 	var usr strings.Builder
 	runtime := renderRuntimeContext(ctx)
 	if runtime != "" {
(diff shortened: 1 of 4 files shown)


eloylp (commented):
**Missing combined test for guard + roster ordering in System**

The new tests in `prompt_test.go` cover the no-PR guard and the roster independently, but there is no test that exercises both at the same time. The intended layout of `System` is:

```
<guard>
<skills>
<agent prompt>
## Available experts
...
```

Because `TestRenderAgentPromptNoPRGuardInSystem` uses an agent with no roster, and `TestRenderAgentPromptRosterAppendsAfterAgentPrompt` uses `AllowPRs: true`, a bug that accidentally interleaved or reversed the guard and roster blocks would silently pass the entire test suite.

Concretely, if someone later changed `renderRoster` to write before the agent prompt body, or wrote the guard after the roster, today's tests would still be green. That matters here more than usual because the whole motivation of this PR is that `System` must be stable and deterministic across runs to benefit from prompt caching.

Consider adding one focused test with `AllowPRs: false` and a non-empty `Roster`, then asserting section ordering:

```go
guardIdx  := strings.Index(got.System, "Do not open or create")
rosterIdx := strings.Index(got.System, "## Available experts")
if guardIdx < 0 || rosterIdx < 0 || guardIdx >= rosterIdx {
    t.Errorf("expected guard before roster; guard=%d roster=%d\nSystem:\n%s", guardIdx, rosterIdx, got.System)
}
```

This pins the full section ordering that the caching strategy depends on.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings
- None beyond previously raised review feedback.

### Summary
Architecture and dependency direction look sound; no additional correctness, testing, or security issues found in this revision.

---

eloylp:
Addressed in 7f23866: added `TestRenderAgentPromptSystemSectionOrdering`, which exercises the full ordering with every optional block present at once — no-PR guard (`AllowPRs: false`), a skill with non-empty guidance, the agent prompt body, and a non-empty `Roster`. It asserts `guard < skills < prompt < roster`, so a future change that reorders or interleaves any of these blocks will fail the test. This pins the determinism the caching strategy relies on.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Changes correctly move stable roster/no-PR content into System, preserve dependency direction, and add robust ordering/behavior tests (including the combined ordering case) without introducing new correctness or security issues.

---

PR #267, opened 2026-04-24
refactor(webhook): simplify canonicalModels using slices.Compact

eloylp:
## Why

`canonicalModels` used a `map[string]struct{}` to deduplicate entries while building the output slice, then sorted at the end. Because the function sorts anyway, adjacent-duplicate removal via `slices.Compact` achieves the same result without the map allocation.

The `slices` package is already used in eight other files in this codebase (Go 1.21+), so this is consistent with the existing style.

## What changed

- `internal/webhook/orphans.go`: rewrite `canonicalModels` to filter/collect, sort, then `slices.Compact` — removes the set map, -7 net lines.
- `internal/webhook/orphans_test.go`: add `TestCanonicalModels` — table-driven, covers nil input, empty/whitespace-only strings, duplicates, and whitespace trimming.

No behaviour change; all existing tests continue to pass.

### DIFF
--- a/internal/webhook/orphans.go
+++ b/internal/webhook/orphans.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -147,22 +148,15 @@ func sortedSetKeys(in map[string]struct{}) []string {
 }
 
 func canonicalModels(models []string) []string {
-	if len(models) == 0 {
+	out := make([]string, 0, len(models))
+	for _, m := range models {
+		if m = strings.TrimSpace(m); m != "" {
+			out = append(out, m)
+		}
+	}
+	if len(out) == 0 {
 		return nil
 	}
-	set := make(map[string]struct{}, len(models))
-	out := make([]string, 0, len(models))
-	for _, model := range models {
-		model = strings.TrimSpace(model)
-		if model == "" {
-			continue
-		}
-		if _, exists := set[model]; exists {
-			continue
-		}
-		set[model] = struct{}{}
-		out = append(out, model)
-	}
 	sort.Strings(out)
-	return out
+	return slices.Compact(out)
 }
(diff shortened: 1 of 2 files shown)


eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/webhook/orphans_test.go:30` — `TestCanonicalModels` only compares lengths and element values, so cases that explicitly expect `nil` (`"nil input"`, `"empty slice"`, `"all empty after trim"`) will still pass if `canonicalModels` returns a non-nil empty slice. This weakens the intended contract check and can hide regressions in JSON/omitempty behavior; assert `got == nil` when `want == nil` (and/or use exact slice equality for non-nil cases).

### Summary
The refactor is sound, but the new table-driven test does not fully assert the nil-vs-empty behavior it claims to cover.

---

eloylp:
Addressed in 1351918: replaced the per-element loop with `reflect.DeepEqual(got, tc.want)`. `reflect.DeepEqual` distinguishes nil slices from non-nil empty slices, so the `"nil input"`, `"empty slice"`, and `"all empty after trim"` cases will now fail if `canonicalModels` ever returns `[]string{}` instead of `nil` — preserving the JSON `omitempty` contract.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The updated commit addresses prior feedback; the MCP/webhook changes are coherent, idiomatic, and sufficiently covered by tests.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The updated commit resolves the prior nil-vs-empty assertion gap; the refactor remains correct, idiomatic, and sufficiently tested.

---

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Re-review after rebase (HEAD `1eda040`). Diff content is unchanged from the previously approved revision — `canonicalModels` using sort+`slices.Compact`, test using `reflect.DeepEqual` for nil-vs-empty contract. CI green.

---

PR #273, opened 2026-04-24
refactor(config): collapse validateDispatchWiring from 4 passes to 2

eloylp:
## Summary

- Collapses `validateDispatchWiring` from 4 passes to 2 by using the full `agentByName` (AgentDef values) instead of a separate `agentNames` set plus a `dispatchTargets` set.
- The description check now happens in the same inner loop as the reference check, eliminating two redundant iterations over `c.Agents`.
- The new shape mirrors `ValidateCrossRefs`, which already solves the same problem in two passes.

## Test plan

- [x] `go test ./... -race` (covered by CI)
- [x] Existing dispatch-validation tests cover both positive and error paths

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -814,33 +814,22 @@ func (c *Config) validateAgents() error {
 //   - can_dispatch must not include the agent itself
 //   - agents referenced in any can_dispatch list must have a description
 func (c *Config) validateDispatchWiring() error {
-	// Build set of all agent names for O(1) lookup.
-	agentNames := make(map[string]struct{}, len(c.Agents))
+	agentByName := make(map[string]AgentDef, len(c.Agents))
 	for _, a := range c.Agents {
-		agentNames[a.Name] = struct{}{}
-	}
-	// Build set of agents that appear in any can_dispatch list — these require
-	// a description so the roster is informative.
-	dispatchTargets := make(map[string]struct{})
-	for _, a := range c.Agents {
-		for _, t := range a.CanDispatch {
-			dispatchTargets[t] = struct{}{}
-		}
+		agentByName[a.Name] = a
 	}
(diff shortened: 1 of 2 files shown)


eloylp (commented):
**Testing gap: the "dispatch target missing description" case doesn't assert the agent name in the error message.**

The refactor moves the description check into the inner loop and now formats the error with `t` (the target's name string) instead of `a.Name` (the target `AgentDef`'s `.Name` field). They're identical in practice, but the existing test only checks `strings.Contains(err.Error(), "description")` — it doesn't pin the specific agent named in the message.

That means a regression like accidentally printing the *dispatcher's* name instead of the target's would pass all tests undetected. Add an assertion on the exact agent name:

```go
{
    name: "dispatch target missing description",
    ...
    errMsg: `agent "pr-reviewer" is in a can_dispatch list but has no description`,
},
```

Since the table already drives `strings.Contains`, just tightening `errMsg` to the full phrase is enough. No new test infrastructure needed.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Refactor preserves dispatch-wiring behavior while simplifying validation flow, and current tests cover the relevant success and failure paths.

---

eloylp:
Addressed in 03a4f53: tightened `errMsg` from `"description"` to the full phrase `agent "pr-reviewer" is in a can_dispatch list but has no description`. The test now asserts the exact target name, so a regression that printed the dispatcher's name instead of the target's would fail.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Refactor keeps dispatch validation behavior intact while reducing complexity, and tests now pin the target-agent error message to prevent regression.

---

PR #274, opened 2026-04-24
feat(mcp): add create_repo and delete_repo tools

eloylp:
## Summary

Adds the last CRUD write pair for issue #227 (high priority): \`create_repo\` and \`delete_repo\` MCP tools. With this PR, the MCP tool inventory covers the full fleet surface declared in the issue — agents/skills/backends/repos all have list/get/create/delete, plus config get/export/import and the observability tools.

Refactors \`handleStoreRepos\` POST and \`handleStoreRepo\` DELETE into reusable \`Server.UpsertRepo(r) → (canonicalRepo, err)\` and \`Server.DeleteRepo(name)\`. Both hold \`storeMu\` and reload cron, matching \`UpsertAgent\`/\`UpsertSkill\`/\`UpsertBackend\`. Empty/whitespace names are now wrapped in \`*store.ErrValidation\` so REST (via \`storeErrStatus\` → 400) and MCP (via \`NewToolResultErrorFromErr\` → user error) surface the same validation shape.

New \`RepoWriter\` interface in \`internal/mcp\` wired via \`Deps.RepoWrite\`. The \`create_repo\` tool accepts a nested \`bindings\` array; since mcp-go's request helpers only expose primitives, a new \`parseBindings\` helper reads the raw arguments map and destructures each entry. Bindings with an omitted \`enabled\` key keep the \`*bool\` nil so \`config.Binding.IsEnabled\` treats them as default-enabled — losing this would silently flip bindings off on every MCP round-trip.

Part of #227. Stacked on \`feat/227-mcp-create-delete-backend\` (#272).

## Test plan

- [x] \`TestToolCreateRepoForwardsAndReturnsCanonical\` — raw name/enabled/bindings forwarded to writer; explicit \`enabled=false\` binding preserved as non-nil pointer; response reflects canonical form.
- [x] \`TestToolCreateRepoRequiresName\` — missing \`name\` key → IsError, writer not invoked.
- [x] \`TestToolCreateRepoRejectsBlankName\` — whitespace-only name reaches writer and surfaces \`*store.ErrValidation\` as user error.
- [x] \`TestToolCreateRepoPropagatesError\` — writer error (e.g. unknown agent) propagates as tool error.
- [x] \`TestToolCreateRepoRejectsBadBindingsShape\` — table-driven: non-array bindings and non-object element each surface a specific user error without reaching the writer.
[...]

PR #272, opened 2026-04-24
feat(mcp): add create_backend and delete_backend tools

eloylp:
## Summary

Adds the final two CRUD-write tools for AI backends under #227:
- \`create_backend\` — upsert semantics via POST /backends code path
- \`delete_backend\` — delete via DELETE /backends/{name} code path

Refactors \`handleStoreBackends\` POST and \`handleStoreBackendDelete\` into
reusable \`Server.UpsertBackend(name, b) → (canonicalName, canonicalBackend, err)\`
and \`Server.DeleteBackend(name)\`. Both hold \`storeMu\` and reload cron,
matching the coupling already used by \`UpsertAgent\`/\`UpsertSkill\`.

Empty/whitespace names are now wrapped in \`*store.ErrValidation\` so REST
(via \`storeErrStatus\` → 400) and MCP (via \`NewToolResultErrorFromErr\` →
user error) surface the same validation shape. Previously REST returned
a plain \"name is required\" 400 via \`http.Error\`; the new path funnels
through the same \`storeErrStatus\` map the other CRUD writers use.

New \`BackendWriter\` interface in \`internal/mcp\` wired via \`Deps.BackendWrite\`.

## Test plan

- [x] \`TestToolCreateBackendForwardsAndReturnsCanonical\` — raw name forwarded, runtime settings + models forwarded, response carries canonical name/command
- [x] \`TestToolCreateBackendRequiresName\` — absent-key path rejected at handler, writer not invoked
- [x] \`TestToolCreateBackendRejectsBlankName\` — whitespace-only reaches \`UpsertBackend\`, surfaces \`*store.ErrValidation\` as user error (pins the contract; forcing function if blank-name guard is ever hoisted up)
- [x] \`TestToolCreateBackendPropagatesError\` — generic writer error propagates with error-body substring assertion
- [x] \`TestToolDeleteBackendNormalizesAndForwards\` — \"  Claude  \" → \"claude\" before forwarding; response has \`status: deleted\`, \`name: claude\`
- [x] \`TestToolDeleteBackendRequiresName\` — whitespace-only rejected at handler, writer not invoked
- [x] \`TestToolDeleteBackendPropagatesConflict\` — delete-conflict error propagates

Stacked on #271. CI does not report checks on stacked branches (base !=
main) until the earlier PR merges — expected, per prior runs.

Part of #227 (high priority). Remaining MCP CRUD scope: repo create/delete.

Closes part of #227.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #268, opened 2026-04-24
feat(mcp): add get_config and export_config tools

eloylp:
## Summary

Adds two read-only config tools to the MCP server:

- `get_config` — returns the redacted effective daemon config as JSON (same body as `GET /config`)
- `export_config` — returns the CRUD-mutable fleet sections as a YAML fragment (same body as `GET /export`, round-trippable through `POST /import`)

Implementation is byte-passthrough: `handleAPIConfig` and `handleStoreExport` were refactored to delegate to new `*webhook.Server` methods (`ConfigJSON`, `ExportYAML`), and the MCP tools call those same methods through a new `mcp.ConfigReader` interface wired from `main`. REST and MCP surfaces therefore emit identical bytes (same redaction, same omissions like `proxy.extra_body`).

Part of #227. CRUD writes and config import remain as follow-ups.

## Stacking

This PR is stacked on top of #266 (`feat/227-mcp-observability-tools`). Please merge #259 → #265 → #266 first; this branch will then rebase cleanly onto main.

## Test plan

- [x] Unit tests for both new tool handlers (happy path + error propagation) using a stub `ConfigReader`.
- [x] Existing REST handler tests (`TestHandleAPIConfig*`, `TestStoreExportReturnsYAML`) still cover the refactored path — they now exercise `ConfigJSON` / `ExportYAML` through the HTTP handlers.
- [x] `*webhook.Server` satisfies `mcp.ConfigReader` at compile time via the wiring in `cmd/agents/main.go`.
- [ ] CI (`go test ./... -race`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #266, opened 2026-04-24
feat(mcp): add observability tools (events, traces, graph, memory)

eloylp:
## Summary

Follow-up to #259 / #265. Extends the MCP tool surface with read-only observability tools so MCP clients (Claude Code, Cursor, Cline) can inspect fleet activity without hitting the REST API separately:

- `list_events` — recent events with optional RFC3339 `since` filter
- `list_traces` — recent agent run spans
- `get_trace` — all spans for one root event ID
- `get_trace_steps` — tool-loop transcript for one span
- `get_graph` — agent interaction graph (nodes + dispatch edges)
- `get_dispatches` — dispatch counters snapshot
- `get_memory` — stored markdown memory for `(agent, repo)`

Each tool mirrors the wire shape of its REST counterpart so consumers parsing either surface share decoders.

## Design notes

- The three new providers (`ObserveStore`, `DispatchStatsSource`, `MemoryReader`) are optional on `Deps`. `registerTools` wires their tools only when the dependency is supplied, so tests and minimal deployments don't need to stub the full stack.
- `MemoryReader` returns `(content, mtime, found, err)` instead of a sentinel error, so `internal/mcp` doesn't inherit `webhook.ErrMemoryNotFound` across the import graph. A thin adapter in `cmd/agents/main.go` bridges to the SQLite-backed store.
- `get_graph` deliberately omits per-node runtime status (running/error) for now — callers can resolve that via `list_agents`. Runtime-state wiring is left for a later PR if needed.
- `get_memory` rejects `.`/`..` identifiers defensively, then relies on `ai.NormalizeToken` downstream to canonicalise the rest.

## Stack

This branch is stacked on top of #265 and contains its commits. Once #259 and #265 land, the net diff should just be `f8dd3b7`.

Part of #227.

## Test plan

- [x] `TestToolListEvents` — nil-slice serialises as `[]`, events returned in order
- [x] `TestToolListEventsSinceFilter` — valid RFC3339 sets filter; invalid falls back to no filter
- [x] `TestToolListTraces`, `TestToolGetTrace`, `TestToolGetTraceSteps` — happy path + missing ID
- [x] `TestToolGetGraphSeedsNodesFromFleetAndEdges` — nodes from fleet + edge endpoints not in config
- [x] `TestToolGetDispatches` — counter snapshot passthrough
- [x] `TestToolGetMemorySuccess` + `Missing` + `RejectsTraversal` + `RequiresBothArgs` + `PropagatesReaderError`
[...]

PR #275, opened 2026-04-24
refactor(ai): extract hasBackendPrefix to deduplicate backend predicates

eloylp:
## Why

`isClaudeBackend` and `isCodexBackend` contained the same 3-line normalization block (name `ToLower`+`TrimSpace`, command `Base`+`ToLower`) and differed only in the prefix string checked. The duplication meant that a change to the normalization strategy (e.g. dropping the command check, or switching to `strings.EqualFold`) would need to be applied in two places.

## What

Extract a private `hasBackendPrefix(prefix string) bool` helper that holds the shared logic once. The two predicate methods become one-liners:

```go
func (r *CommandRunner) isClaudeBackend() bool { return r.hasBackendPrefix("claude") }
func (r *CommandRunner) isCodexBackend() bool  { return r.hasBackendPrefix("codex") }
```

No behavior change. Existing `buildDelivery` tests cover both predicates indirectly.

### DIFF
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -236,16 +236,16 @@ func (r *CommandRunner) buildDelivery(req Request) (args []string, stdin string)
 	return nil, truncateString(combineSystemUser(req.System, req.User), r.maxPromptChars)
 }
 
-func (r *CommandRunner) isClaudeBackend() bool {
-	name := strings.ToLower(strings.TrimSpace(r.backendName))
-	cmd := strings.ToLower(filepath.Base(strings.TrimSpace(r.command)))
-	return strings.HasPrefix(name, "claude") || strings.HasPrefix(cmd, "claude")
-}
+func (r *CommandRunner) isClaudeBackend() bool { return r.hasBackendPrefix("claude") }
+func (r *CommandRunner) isCodexBackend() bool  { return r.hasBackendPrefix("codex") }
 
-func (r *CommandRunner) isCodexBackend() bool {
+// hasBackendPrefix reports whether the backend name or command basename starts
+// with prefix (case-insensitive). Centralises the name/command normalisation
+// so each new backend type only needs a one-liner predicate.
+func (r *CommandRunner) hasBackendPrefix(prefix string) bool {
 	name := strings.ToLower(strings.TrimSpace(r.backendName))
 	cmd := strings.ToLower(filepath.Base(strings.TrimSpace(r.command)))
-	return strings.HasPrefix(name, "codex") || strings.HasPrefix(cmd, "codex")
+	return strings.HasPrefix(name, prefix) || strings.HasPrefix(cmd, prefix)
 }
 
 // buildClaudeDelivery builds hardcoded Claude CLI args and delivers system
--- a/internal/ai/cmdrunner_test.go
+++ b/internal/ai/cmdrunner_test.go
@@ -835,3 +835,38 @@ func TestExtractToolResultText(t *testing.T) {
 		})
 	}
 }
+
+func TestHasBackendPrefix(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name    string
+		beName  string
+		command string
+		prefix  string
+		want    bool
+	}{
+		{name: "exact name match", beName: "claude", command: "", prefix: "claude", want: true},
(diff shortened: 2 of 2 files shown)


eloylp (commented):
The extraction is clean and the one-liner predicates read well. One concern worth addressing before merging:

**`hasBackendPrefix` normalisation logic is only covered indirectly.**

The PR description says so explicitly: _"Existing `buildDelivery` tests cover both predicates indirectly."_ That was acceptable when the logic was copy-pasted in two private methods, but now that it lives in a single helper, the normalisation rules (whitespace trimming, case-folding, `filepath.Base` stripping) become the single point of failure — and they're not exercised by a test that names them.

A table-driven test in `cmdrunner_test.go` would cost ~30 lines and lock in the contract directly:

```go
func TestHasBackendPrefix(t *testing.T) {
    t.Parallel()
    cases := []struct {
        name    string
        beName  string
        cmd     string
        prefix  string
        want    bool
    }{
        {"exact name match",            "claude",             "",                     "claude", true},
        {"name with spaces",            "  claude-prod  ",    "",                     "claude", true},
        {"name uppercase",              "Claude",             "",                     "claude", true},
        {"command basename match",      "",                   "/usr/local/bin/claude","claude", true},
        {"command basename uppercase",  "",                   "/usr/bin/Claude",      "claude", true},
        {"no match",                    "openai",             "/usr/bin/gpt",         "claude", false},
        {"codex via name",              "codex-fast",         "",                     "codex",  true},
    }
    for _, tc := range cases {
        t.Run(tc.name, func(t *testing.T) {
            t.Parallel()
            r := &CommandRunner{backendName: tc.beName, command: tc.cmd}
            if got := r.hasBackendPrefix(tc.prefix); got != tc.want {
                t.Errorf("hasBackendPrefix(%q) = %v, want %v", tc.prefix, got, tc.want)
            }
        })
    }
}
```

Without this, a future change to the normalisation strategy (e.g. switching to `strings.EqualFold`, dropping the command check) has no direct signal that it broke the intended behaviour — the failure would only surface as a `buildDelivery` test failure with no clear pointer to the root cause.

eloylp:
---
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/mcp/server.go:1` — The PR description says this is a no-behavior-change extraction in `internal/ai/cmdrunner.go`, but this branch also introduces a new `/mcp` server surface with read/write tools and daemon wiring (`cmd/agents/main.go`). That is a material behavior/security-surface change and should not be hidden behind a refactor-only description.
- **[medium]** `internal/ui/src/components/RepoFilter.tsx:1` — This PR also includes unrelated UI feature work and lockfile churn while the title/body describe only backend-prefix dedup. Please either split into focused PRs or update metadata to accurately describe the full change set.

### Summary
The change set is much broader than advertised; request updates to PR scope/description before merge so reviewers can evaluate the right risk.

---

eloylp:
Addressed in force-push to `1e6e6c3`: rebased the branch onto `origin/main` so the diff now contains only the intended `internal/ai/cmdrunner.go` change (7/-7 lines). The 15 unrelated commits (MCP stack + RepoFilter + lockfile churn) were carried over from the parent branch at PR creation time and were tracked under their own PRs (#259, #265, #266, #268–#272, #274, #247). They are dropped from this PR; each is still present in its respective open PR. The title/description now accurately describe the full change set.

eloylp:
Addressed in bc2493b: added `TestHasBackendPrefix` in `internal/ai/cmdrunner_test.go` as a table-driven test covering the normalisation contract directly — exact/substring name match (case-insensitive), whitespace trimming, `filepath.Base` command basename match (case-insensitive, with leading path), codex via both name and command, empty name+command, and the "substring-but-not-prefix" negative case. Any future change to the normalisation strategy (e.g. switching to `strings.EqualFold` or dropping the command check) now fails a test that names the helper directly, instead of surfacing as a `buildDelivery` failure with no clear pointer.

eloylp:
Rebased onto latest main (HEAD 0fe5456). The branch was previously based on an older main and the GitHub diff widened when the parent stack (MCP server + RepoFilter) merged. After the rebase the PR diff is now exactly the advertised scope: `internal/ai/cmdrunner.go` (`hasBackendPrefix` dedup, 14 lines net) + `internal/ai/cmdrunner_test.go` (new table-driven test, 35 lines). No MCP, UI, or lockfile churn.

PR #279, opened 2026-04-25
refactor(mcp): use repoJSON helper in list/get repo tools

eloylp:
## Summary

`repoJSON` already exists in `tools_crud.go` to serialise a `config.RepoDef` into the `{name, enabled, bindings}` wire shape used by `create_repo`. But `toolListRepos` and `toolGetRepo` in `tools_fleet.go` duplicate the same binding-expansion loop inline instead of calling it.

This PR removes the duplication: both functions now delegate to `repoJSON`, so all three repo-serialisation paths (list, get, create) share a single implementation. No behaviour change — the output shape is identical.

## Test plan

- [ ] CI `go test ./... -race` green

### DIFF
--- a/internal/mcp/tools_fleet.go
+++ b/internal/mcp/tools_fleet.go
@@ -142,15 +142,7 @@ func toolListRepos(deps Deps) server.ToolHandlerFunc {
 		}
 		out := make([]map[string]any, 0, len(repos))
 		for _, r := range repos {
-			bindings := make([]map[string]any, 0, len(r.Use))
-			for _, b := range r.Use {
-				bindings = append(bindings, bindingJSON(b))
-			}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

No issues found. The removed inline code in both `toolListRepos` and `toolGetRepo` is byte-for-byte equivalent to what `repoJSON` produces — same map keys (`name`, `enabled`, `bindings`), same `bindingJSON` expansion. The refactor is correct and the deduplication is real: any future field added to `repoJSON` now covers all three serialization sites automatically.

### Summary

Clean, minimal deduplication with no behaviour change — straightforward approval.

PR #280, opened 2026-04-25
refactor(mcp): use boolPtrArg for enabled in binding tools

eloylp:
## Summary

`toolCreateBinding` and `toolUpdateBinding` expanded the optional `enabled` field inline with a 5-line type-assertion block each, while every other optional-bool field across the MCP tools (`allow_prs`, `allow_dispatch`, `healthy`) uses `boolPtrArg`. This inconsistency means the same logic exists twice in diverging form.

Both binding tools now use `boolPtrArg`, matching the pattern already established in `toolUpdateAgent` and `toolUpdateBackend`. No behaviour change — `boolPtrArg` produces the same `(*bool, present, errMsg)` semantics.

## Test plan

- [ ] CI `go test ./... -race` green

### DIFF
--- a/internal/mcp/tools_crud.go
+++ b/internal/mcp/tools_crud.go
@@ -384,12 +384,10 @@ func toolCreateBinding(deps Deps) server.ToolHandlerFunc {
 			Events: req.GetStringSlice("events", nil),
 			Cron:   req.GetString("cron", ""),
 		}
-		if v, ok := req.GetArguments()["enabled"]; ok && v != nil {
-			enabled, ok := v.(bool)
-			if !ok {
-				return mcpgo.NewToolResultError("enabled must be a boolean"), nil
-			}
-			b.Enabled = &enabled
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

No issues found. `boolPtrArg` is semantically equivalent to the replaced inline blocks:
- absent/nil key → `(nil, false, "")` → neither branch fires, `b.Enabled` stays nil — matches the old `ok && v != nil` guard
- non-bool value → `(nil, false, errMsg)` → error returned — matches the old failed type-assertion path
- valid bool → `(&b, true, "")` → `b.Enabled = v` — matches the old assignment

Error message text is identical (`"enabled must be a boolean"`). The change correctly aligns `toolCreateBinding` and `toolUpdateBinding` with the pattern already used by `toolUpdateAgent` (`allow_prs`, `allow_dispatch`) and `toolUpdateBackend` (`healthy`).

### Summary

Correct, minimal deduplication that closes the last inconsistency in the optional-bool handling pattern.

PR #284, opened 2026-04-25
refactor(mcp): extract bindingIDAndRepo helper

eloylp:
## Summary

`toolGetBinding`, `toolUpdateBinding`, and `toolDeleteBinding` each open with the same 8-line block that reads and validates the `id` and `repo` parameters:

```go
id, err := req.RequireInt("id")
if err != nil {
    return mcpgo.NewToolResultError(err.Error()), nil
}
if id <= 0 {
    return mcpgo.NewToolResultError("id must be a positive integer"), nil
}
repo, ok := trimmedString(req, "repo")
if !ok {
    return mcpgo.NewToolResultError("repo is required"), nil
}
```

Extract that block into a `bindingIDAndRepo(req) (int, string, string)` helper so the three tools share one validation path and the error messages stay consistent.

- Return convention `(int, string, string)` matches the existing helpers: non-empty third return is the user-facing error message, consistent with `boolPtrArg` / `intPtrArg`.

### DIFF
--- a/internal/mcp/tools_crud.go
+++ b/internal/mcp/tools_crud.go
@@ -403,16 +403,9 @@ func toolCreateBinding(deps Deps) server.ToolHandlerFunc {
 // surface as not-found.
 func toolUpdateBinding(deps Deps) server.ToolHandlerFunc {
 	return func(_ context.Context, req mcpgo.CallToolRequest) (*mcpgo.CallToolResult, error) {
-		id, err := req.RequireInt("id")
-		if err != nil {
-			return mcpgo.NewToolResultError(err.Error()), nil
-		}
-		if id <= 0 {
-			return mcpgo.NewToolResultError("id must be a positive integer"), nil
-		}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- `bindingIDAndRepo` reproduces the exact inline logic from all three tools — `RequireInt`, the `id <= 0` guard, and `trimmedString("repo")` — without any semantic change.
- Return convention `(int, string, string)` where a non-empty third value signals failure is consistent with the existing `boolPtrArg` / `intPtrArg` helpers; the function comment documents the non-obvious convention appropriately.
- Zero-value returns on error paths (`0, "", errMsg`) are safe — callers check `errMsg != ""` before using `id` or `repo`.
- The `int64(id)` casts at call sites are preserved, matching the original inline code.

### Summary

Clean, safe extraction — removes a duplicated 8-line block from three tools with no behaviour change.

PR #285, opened 2026-04-25
refactor(store): extract validateFleetConstraints from ImportAll and ReplaceAll

eloylp:
## Summary

`ImportAll` and `ReplaceAll` each close with an identical 20-line validation block:

```go
if err := validateFleet(tx); err != nil {
    return &ErrValidation{Msg: fmt.Sprintf("store: import: %v", err)}
}
if err := requireAtLeastOne(tx, "SELECT COUNT(*) FROM agents", "agents", "config: at least one agent is required"); err != nil {
    return &ErrValidation{Msg: fmt.Sprintf("store: import: %v", err)}
}
if err := requireAtLeastOne(tx, "SELECT COUNT(*) FROM backends", "backends", "config: at least one ai_backends entry is required"); err != nil {
    return &ErrValidation{Msg: fmt.Sprintf("store: import: %v", err)}
}
if err := requireAtLeastOne(tx, "SELECT COUNT(*) FROM repos WHERE enabled=1", "enabled repos", "config: at least one repo must be enabled"); err != nil {
    return &ErrValidation{Msg: fmt.Sprintf("store: import: %v", err)}
}
if err := validateCronExpressions(repos); err != nil {
    return err
}
```

The only difference between the two copies is `"import"` vs `"replace"` in the error prefix. Extract into `validateFleetConstraints(q querier, op string, repos []config.RepoDef) error` so each function calls one helper instead of repeating the same five checks.

- Return convention matches the surrounding helpers: non-nil error on the first failing check.

### DIFF
--- a/internal/store/crud.go
+++ b/internal/store/crud.go
@@ -97,7 +97,27 @@ func requireAtLeastOne(q querier, countQuery, entity, zeroMsg string) error {
 	return nil
 }
 
-// ──── Agents ─────────────────────────────────────────────────────────────────
+// validateFleetConstraints runs all post-write invariant checks shared by
+// ImportAll and ReplaceAll: entity cross-references, minimum cardinality, and
+// cron-expression parseability. op ("import" or "replace") is used verbatim in
+// error messages.
+func validateFleetConstraints(q querier, op string, repos []config.RepoDef) error {
+	if err := validateFleet(q); err != nil {
+		return &ErrValidation{Msg: fmt.Sprintf("store: %s: %v", op, err)}
+	}
+	if err := requireAtLeastOne(q, "SELECT COUNT(*) FROM agents", "agents", "config: at least one agent is required"); err != nil {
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- `validateFleetConstraints(q querier, op string, repos []config.RepoDef) error` accepts `tx` via the `querier` interface — correct, no indirection added.
- Error message output is identical: `"store: import: %v"` / `"store: replace: %v"` are reproduced exactly via `fmt.Sprintf("store: %s: %v", op, err)`.
- `validateCronExpressions(repos)` is still returned directly (no `ErrValidation` wrapping), preserving the original callers' behaviour in both `ImportAll` and `ReplaceAll`.
- The helper is unexported — right call, it's only used within this package.
- The section-divider lines being slightly lengthened is cosmetic noise in the diff but harmless.

### Summary

Clean, safe extraction — removes a duplicated 20-line validation block from two functions with no behaviour change.

PR #287, opened 2026-04-25
fix(mcp): accept JSON-stringified array args at the tool boundary

### LINKED ISSUE #278: mcp: array-typed tool arguments must accept JSON-stringified values
## Problem

When an MCP client calls a fleet tool that accepts an array parameter (`skills`, `can_dispatch`, `models`, `events`, `labels`), and the transport delivers the argument as a JSON-encoded string instead of a native JSON array, the handler rejects the call with:

> skills must be an array of strings

Observed today while invoking `update_agent` from Claude Code — four consecutive PATCH attempts against `coder`, `pr-reviewer`, `refactorer`, `test-reviewer` failed on this exact error, even though the intent and payload were well-formed. Falling back to `PATCH /agents/{name}` over REST worked on the first try with the same array contents.

## Root cause

\`stringSlicePtrArg\` in \`internal/mcp/tools_crud.go\` only accepts \`[]any\`:

\`\`\`go
raw, ok := v.([]any)
if !ok {
    return nil, false, fmt.Sprintf("%s must be an array of strings", key)
}
\`\`\`

If the client delivers \`\"[\\\"a\\\",\\\"b\\\"]\"\` (a JSON string containing an array), the assertion fails and the tool surfaces a misleading error. The user is not wrong — they sent an array's worth of data — but they hit a type-boundary shim they can't see.

Same pattern lives on at least these tools, all of which are high-traffic once conversational fleet management kicks in:
- \`create_agent\`, \`update_agent\` — \`skills\`, \`can_dispatch\`
- \`create_backend\`, \`update_backend\` — \`models\`
- \`create_binding\`, \`update_binding\` — \`labels\`, \`events\`

## Fix direction

Accept both shapes in \`stringSlicePtrArg\` (and the \`[]any\` path inside \`parseBindings\` used by \`create_repo\`). Proposed logic:

1. If \`v\` is already \`[]any\`, process as today.
2. If \`v\` is a \`string\`, try \`json.Unmarshal\` into \`[]string\`; on success, use that.
3. Otherwise, return the current error.

This keeps the strict type check for truly bogus inputs (e.g. \`true\`, \`42\`), while absorbing the harmless stringification that some clients produce at the JSON-RPC boundary.

## Acceptance criteria

- \`update_agent\` succeeds when \`skills\` is supplied as a JSON-encoded string (e.g. \`\"[\\\"architect\\\", \\\"go-testing\\\"]\"\`).
- Same relief applies to \`can_dispatch\`, \`models\`, \`events\`, \`labels\` across all CRUD tools.
- Non-array, non-stringified-array inputs still return a clear \`must be an array of strings\` error.
- Unit tests in \`internal/mcp/tools_test.go\` cover both native-array and stringified-array inputs for at least one agent tool and one backend tool.
- REST behaviour is unchanged — this is an MCP-layer relaxation only.

## Severity

High — every conversational rename / mass-update workflow trips over this, and the error message makes the user suspect their payload rather than the transport.

eloylp:
Closes #278.

## Summary

Some MCP clients — observed with `mark3labs/mcp-go` when batching tool calls into a single JSON-RPC message — stringify array parameters at the transport boundary, so `skills: ["a","b"]` arrives as the literal string `"[\"a\",\"b\"]"`. Today every fleet CRUD tool that takes an array argument (`skills`, `can_dispatch`, `models`, `events`, `labels`, `bindings`) rejects that shape with a misleading `"must be an array of strings"` error, breaking parallel writes over MCP — exactly the high-traffic pattern the fleet is designed to support.

## What changed

- New helper `stringSliceArg(v, key) ([]string, errMsg)` accepts native `[]any`, native `[]string`, **and** JSON-encoded array strings. Returns a clear field-named error for any other shape.
- New helper `arrayOfAny(v, key) ([]any, errMsg)` does the same coercion for `parseBindings`'s top-level argument.
- `stringSlicePtrArg` now delegates to `stringSliceArg`, so PATCH callers (`update_agent`, `update_backend`) inherit the relief.
- Tool handlers that previously used `req.GetStringSlice` for relaxed fields — `toolCreateAgent` (`skills`, `can_dispatch`), `toolCreateBackend` (`models`), `toolCreateBinding` and `toolUpdateBinding` (`labels`, `events`) — now use `stringSliceArg` so the transport relief reaches every CRUD tool, not just the PATCH ones. Bogus inputs still surface a clear `<field> must be an array of strings` error and never reach the writer.
- `parseBindings` accepts a stringified `bindings` array at the top level via `arrayOfAny`. Nested `labels`/`events` flow through `stringSliceArg` and get the same JSON-string relief.
- REST behaviour is unchanged — this is an MCP-layer relaxation only.

## Acceptance criteria coverage

- [x] `update_agent` succeeds when `skills` is supplied as a JSON-encoded string — covered by `TestToolUpdateAgentAcceptsJSONStringSkills`.
- [x] Same relief applies to `can_dispatch`, `models`, `events`, `labels` across all CRUD tools — covered by `TestToolUpdateBackendAcceptsJSONStringModels`, `TestToolCreateAgentAcceptsJSONStringSlices`, `TestToolCreateBindingAcceptsJSONStringSlices`, `TestToolUpdateBindingAcceptsJSONStringSlices`, and `TestToolCreateRepoAcceptsJSONStringBindings`.
- [x] Non-array, non-stringified-array inputs still return a clear `must be an array of strings` error — covered by `TestStringSliceArgAcceptsShapes`, `TestStringSlicePtrArgAcceptsJSONString`, and `TestToolUpdateAgentRejectsBogusSkills`.
- [x] Unit tests in `internal/mcp/tools_arrays_test.go` cover both native-array and stringified-array inputs for at least one agent tool and one backend tool (in fact every relaxed tool).
- [x] REST behaviour unchanged — only `internal/mcp/tools_crud.go` is touched.

## Test plan

- [ ] CI green on `-race`.
- [ ] Existing `TestToolCreateRepoRejectsBadBindingsShape` and `TestToolCreateRepoRejectsBadBindingFieldTypes` still pass — substring assertions remain satisfied because the new error string is a superset of the old one (`"must be an array"` is a prefix of `"must be an array of strings"`).

### DIFF
--- a/internal/mcp/tools_arrays_test.go
+++ b/internal/mcp/tools_arrays_test.go
@@ -30,6 +30,7 @@ func TestStringSliceArgAcceptsShapes(t *testing.T) {
 		{"native []string", []string{"x"}, []string{"x"}, ""},
 		{"JSON-encoded array string", `["a","b"]`, []string{"a", "b"}, ""},
 		{"JSON-encoded empty array", `[]`, []string{}, ""},
+		{"JSON-encoded null decodes as nil slice", `null`, nil, ""},
 		{"JSON-encoded array with spaces", `[ "a" , "b" ]`, []string{"a", "b"}, ""},
 		{"non-array string is rejected", "ready", nil, "skills must be an array of strings"},
 		{"empty string is rejected", "", nil, "skills must be an array of strings"},
@@ -81,6 +82,7 @@ func TestStringSlicePtrArgAcceptsJSONString(t *testing.T) {
 		{"native array → ptr to slice", map[string]any{"skills": []any{"a"}}, true, 1, ""},
 		{"JSON-string array → ptr to slice", map[string]any{"skills": `["a","b"]`}, true, 2, ""},
 		{"JSON-string empty array → ptr to empty", map[string]any{"skills": `[]`}, true, 0, ""},
+		{"JSON-string null → ptr to nil slice (explicit clear)", map[string]any{"skills": `null`}, true, 0, ""},
 		{"bad shape → error", map[string]any{"skills": "ready"}, false, 0, "skills must be an array of strings"},
 	}
 
@@ -109,6 +111,53 @@ func TestStringSlicePtrArgAcceptsJSONString(t *testing.T) {
 			}
 			if got := len(*ptr); got != tc.wantLen {
 				t.Fatalf("len = %d, want %d", got, tc.wantLen)
+			}
+		})
+	}
+}
+
+// TestArrayOfAnyAcceptsShapes pins the shape contract for the helper that
+// decodes nested-object tool arguments (e.g. create_repo's "bindings"
+// payload). Mirrors TestStringSliceArgAcceptsShapes: native []any, JSON-
+// encoded array strings, and JSON-encoded null all succeed; non-array
+// inputs surface a clear "must be an array" error that names the field.
+func TestArrayOfAnyAcceptsShapes(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name    string
+		in      any
+		wantLen int
+		wantNil bool
+		wantErr string
+	}{
+		{"native []any passthrough", []any{map[string]any{"agent": "coder"}}, 1, false, ""},
+		{"native empty []any", []any{}, 0, false, ""},
+		{"JSON-encoded array of objects", `[{"agent":"coder"}]`, 1, false, ""},
+		{"JSON-encoded empty array", `[]`, 0, false, ""},
+		{"JSON-encoded null decodes as nil slice", `null`, 0, true, ""},
+		{"nil falls through to error", nil, 0, true, "bindings must be an array"},
+		{"non-array string rejected", "not-an-array", 0, true, "bindings must be an array"},
+		{"JSON-encoded number rejected", `123`, 0, true, "bindings must be an array"},
+		{"wrong scalar type rejected", 42, 0, true, "bindings must be an array"},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			got, errMsg := arrayOfAny(tc.in, "bindings")
+			if errMsg != tc.wantErr {
+				t.Fatalf("err = %q, want %q", errMsg, tc.wantErr)
+			}
+			if tc.wantErr != "" {
+				return
+			}
+			if tc.wantNil && got != nil {
+				t.Fatalf("got = %v, want nil", got)
+			}
+			if len(got) != tc.wantLen {
+				t.Fatalf("len = %d, want %d", len(got), tc.wantLen)
 			}
 		})
 	}
--- a/internal/mcp/tools_arrays_test.go
+++ b/internal/mcp/tools_arrays_test.go
@@ -1,0 +1,348 @@
+package mcp
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	mcpgo "github.com/mark3labs/mcp-go/mcp"
+	"github.com/rs/zerolog"
+
+	"github.com/eloylp/agents/internal/config"
+)
+
+// TestStringSliceArgAcceptsShapes pins the shape contract for the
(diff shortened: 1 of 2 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[low]** `tools_crud.go` (`stringSliceArg`, `string` case) — `json.Unmarshal([]byte("null"), &decoded)` succeeds (JSON null → nil slice, no error), so the literal string `"null"` returns `(nil, "")` instead of an error. Via `stringSlicePtrArg` in PATCH this lands as `(&nil, true, "")` — "explicit nil, clear the list" — rather than "absent, preserve". Extremely unlikely input, but the shape contract is undocumented here; a test pin in `TestStringSliceArgAcceptsShapes` would lock it down.

- **[low]** `tools_arrays_test.go` — `arrayOfAny` has no dedicated unit test; it's exercised only transitively through `TestToolCreateRepoAcceptsJSONStringBindings`. A `TestArrayOfAnyAcceptsShapes` parallel to `TestStringSliceArgAcceptsShapes` would complete the surface.

- **[info]** `"[]"` → `[]string{}` → `stringSlicePtrArg` returns `(&empty, true, "")` → PATCH clears the list. Confirmed correct: an explicit empty array (native or JSON-encoded) should mean "clear", not "preserve".

- **[info]** Stricter contract on create_* handlers (`req.GetStringSlice` → `stringSliceArg`) is desirable — silent drops gave callers no signal when they sent the wrong shape. The `mergeable_state` shows `unstable`; confirm CI goes green before merge to verify no existing test relied on the old silent-drop path.

- **[info]** Error message format (`"field must be an array of strings"`, `"field[i] must be a string"`) is consistent with surrounding handlers throughout.

### Summary

Correct fix with thorough coverage across the full CRUD surface. The `"null"`-string edge case and missing `arrayOfAny` unit test are cosmetic; the core semantics and strictness change are sound.

eloylp:
Addressed both low-severity findings in 18b17407:

- **`"null"` shape contract pinned** — added `JSON-encoded null decodes as nil slice` case to `TestStringSliceArgAcceptsShapes` and `JSON-string null → ptr to nil slice (explicit clear)` case to `TestStringSlicePtrArgAcceptsJSONString`. The literal string `"null"` decodes via `json.Unmarshal` to a nil slice, so `stringSliceArg` returns `(nil, "")` and `stringSlicePtrArg` returns a non-nil pointer to a nil slice — i.e. "explicit clear", same shape as `[]`. The contract is now locked in tests.
- **`TestArrayOfAnyAcceptsShapes` added** — covers native `[]any`, JSON-encoded array of objects, JSON-encoded `[]`, JSON-encoded `null`, plus the rejection paths (nil, non-array string, JSON number, wrong scalar). Mirrors the structure of `TestStringSliceArgAcceptsShapes`.

The `[info]` notes (CI confirmation, error-message consistency) were already correct as written.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

Both low-severity findings from the previous review are addressed:

- **`"null"` shape contract pinned** — `TestStringSliceArgAcceptsShapes` now includes `"JSON-encoded null decodes as nil slice"` and `TestStringSlicePtrArgAcceptsJSONString` includes `"JSON-string null → ptr to nil slice (explicit clear)"`. The semantics are correct: `json.Unmarshal("null", &[]string{})` yields a nil slice, so `stringSliceArg` returns `(nil, "")` and `stringSlicePtrArg` returns a non-nil pointer to a nil slice — "explicit clear", distinct from "absent/preserve". Both paths are now locked in.

- **`TestArrayOfAnyAcceptsShapes` added** — Covers native `[]any`, JSON-encoded array of objects, `[]`, `null`, plus the rejection paths (nil, non-array string, JSON number, wrong scalar). Mirrors the structure of `TestStringSliceArgAcceptsShapes` cleanly.

No new issues in the production code — the `stringSliceArg` / `arrayOfAny` implementation and all CRUD call sites reviewed previously are unchanged.

### Summary

Both findings closed with correct, thorough test coverage. Ready to merge.

PR #288, opened 2026-04-25
refactor(webhook): extract storeWriteErr helper from 15 CRUD handler error blocks

eloylp:
## Summary

POST/PATCH/DELETE handlers for agents, skills, backends, repos, and bindings all repeat the same 3-line pattern:

```go
status := storeErrStatus(err)
s.logger.Error().Err(err).Msg("store crud: XXX failed")
http.Error(w, fmt.Sprintf("XXX: %v", err), status)
```

- Extract into `storeWriteErr(w, err, op)` so each call site becomes a single line
- 15 identical 3-line blocks → 15 single-line calls (-22 net lines in `internal/webhook/crud.go`)
- No behaviour change: log message, HTTP body, and status code are identical at every site
- Two non-CRUD handlers (`handleBackendsDiscover`, `handleBackendsLocal`) follow a different log convention and are intentionally left unchanged

## Test plan

- [ ] `go test ./internal/webhook/...` passes
- [ ] `go vet ./...` passes
- [ ] No behaviour change — error responses and log lines are structurally identical to before

### DIFF
--- a/internal/webhook/crud.go
+++ b/internal/webhook/crud.go
@@ -353,6 +353,15 @@ func storeErrStatus(err error) int {
 	return http.StatusInternalServerError
 }
 
+// storeWriteErr maps an error from a store write operation to an HTTP response
+// and a structured log entry. op identifies the failing operation (e.g.
+// "agent upsert or cron reload") and appears in both the log line and the
+// HTTP error body so callers and operators see the same context.
+func (s *Server) storeWriteErr(w http.ResponseWriter, err error, op string) {
+	s.logger.Error().Err(err).Msgf("store crud: %s failed", op)
+	http.Error(w, fmt.Sprintf("%s: %v", op, err), storeErrStatus(err))
+}
+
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

No issues found. The extraction is mechanically faithful across all 15 call sites:

- Log message format: `Msgf("store crud: %s failed", op)` reproduces each original `Msg("store crud: … failed")` string exactly.
- HTTP body: `fmt.Sprintf("%s: %v", op, err)` reproduces each original body string exactly.
- Status code: `storeErrStatus(err)` is called once in the helper — same result, no extra allocation.
- `storeWriteErr` is correctly a method on `*Server` (needs `s.logger`) and unexported (internal to the package).
- The two non-CRUD handlers (`handleBackendsDiscover`, `handleBackendsLocal`) that follow a different log convention are intentionally untouched, as noted in the PR description.

### Summary

Clean, safe extraction — 15 identical 3-line blocks collapsed to single calls with no behaviour change.

PR #294, opened 2026-04-25
refactor(mcp): replace for-range search with slices.IndexFunc in toolGetAgent and toolGetRepo

eloylp:
## Summary

`toolGetAgent` and `toolGetRepo` each contain a 4-line `for`+`if` loop that scans a slice for the first element whose `Name` matches a normalized key, then falls through to a not-found return:

```go
for _, a := range agents {
    if a.Name == key {
        return jsonResult(agentJSON(a))
    }
}
return mcpgo.NewToolResultErrorf("agent %q not found", name), nil
```

Go 1.21 introduced `slices.IndexFunc`, which expresses this intent as a single predicate call. Both functions become:

```go
if idx := slices.IndexFunc(agents, func(a config.AgentDef) bool { return a.Name == key }); idx != -1 {
    return jsonResult(agentJSON(agents[idx]))
}
return mcpgo.NewToolResultErrorf("agent %q not found", name), nil
```

- No behaviour change — `IndexFunc` returns the index of the first match or -1, identical semantics to the loop.
- `"slices"` is already imported in this file (used by `slices.Sorted` in `toolListSkills` and `toolListBackends`).
- No new imports.
- `-4` lines net.

The map-based lookups in `toolGetSkill` and `toolGetBackend` are already idiomatic and unchanged.

## Test plan

- [ ] CI green (`go test -race ./internal/mcp/...` — existing `TestToolGetAgent*` and `TestToolGetRepo*` tests cover both the found and not-found paths)

### DIFF
--- a/internal/mcp/tools_fleet.go
+++ b/internal/mcp/tools_fleet.go
@@ -45,10 +45,8 @@ func toolGetAgent(deps Deps) server.ToolHandlerFunc {
 			return mcpgo.NewToolResultErrorFromErr("get agent", err), nil
 		}
 		key := config.NormalizeAgentName(name)
-		for _, a := range agents {
-			if a.Name == key {
-				return jsonResult(agentJSON(a))
-			}
+		if idx := slices.IndexFunc(agents, func(a config.AgentDef) bool { return a.Name == key }); idx != -1 {
+			return jsonResult(agentJSON(agents[idx]))
 		}
 		return mcpgo.NewToolResultErrorf("agent %q not found", name), nil
 	}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

No issues found.

- Both substitutions are semantically identical: `slices.IndexFunc` returns the index of the first match or `-1`, which is the same control flow as the break-on-first-match `for`/`if` loop.
- Empty-slice edge case: `IndexFunc` on an empty slice returns `-1`, so the not-found return is still reached — correct.
- `slices` is already imported (used by `slices.Sorted` and `slices.Collect` elsewhere in the file), so no new dependency.
- The pattern is consistent with `toolGetSkill` and `toolGetBackend`, which already use map lookups rather than linear search (a different structure — the `IndexFunc` form is the right analog for the slice-based lookups here).

### Summary

Clean, safe idiom update — two mechanical substitutions, no behaviour change, no new surface.

PR #296, opened 2026-04-25
refactor(store): extract bindingEnabledInt helper for nil-means-enabled *bool→int

eloylp:
## Summary

`CreateBinding`, `UpdateBinding`, and `importRepos` each contain the identical 3-line block that converts a binding's nullable `enabled` flag to a SQLite 0/1 integer:

```go
enabled := 1
if b.Enabled != nil && !*b.Enabled {
    enabled = 0
}
```

The nil-means-enabled semantics are non-obvious — a reader must reason through the pointer dereference and the double-negative to understand that nil → 1 (enabled by default), matching `IsEnabled()` on the `Binding` type. With three copies the rule lives in no single authoritative place.

Extract it as `bindingEnabledInt(*bool) int` in `store.go`, alongside the existing `boolToInt` / `intToBool` pair, so the semantics are named and explained once.

- `bindingEnabledInt` is package-private (lowercase), consistent with `boolToInt` / `intToBool`.
- Net: −6 lines across 2 files.

## Test plan

- [ ] CI green (`go test -race ./internal/store/...` — existing binding CRUD tests in `crud_test.go` and `store_test.go` cover all three call sites)

### DIFF
--- a/internal/store/crud.go
+++ b/internal/store/crud.go
@@ -607,10 +607,7 @@ func CreateBinding(db *sql.DB, repoName string, b config.Binding) (int64, config
 	if err != nil {
 		return 0, config.Binding{}, fmt.Errorf("store: create binding: marshal events: %w", err)
 	}
-	enabled := 1
-	if b.Enabled != nil && !*b.Enabled {
-		enabled = 0
-	}
+	enabled := bindingEnabledInt(b.Enabled)
 	res, err := tx.Exec(
 		`INSERT INTO bindings(repo,agent,labels,events,cron,enabled) VALUES (?,?,?,?,?,?)`,
 		repoName, b.Agent, string(labels), string(events), b.Cron, enabled,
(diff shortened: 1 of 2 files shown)


eloylp:
@claude review

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- `bindingEnabledInt(*bool) int` reproduces the exact inline logic from all three call sites — `CreateBinding`, `UpdateBinding`, and `importRepos` — without any semantic change.
- Placement in `store.go` alongside `boolToInt` / `intToBool` is correct; the three helpers form a natural group.
- Package-private (lowercase) — consistent with `boolToInt` / `intToBool`.
- The function comment documents the non-obvious nil-means-enabled semantics in one place, which is precisely the value the extraction provides. Readers no longer have to reason through `!*b.Enabled` three times.
- Zero value path: `nil → return 1` (enabled by default); `non-nil && false → return 0`; `non-nil && true → fall through → return 1`. All three cases match the previous inline blocks exactly.

### Summary

Clean, safe extraction — removes a duplicated 3-line block from three call sites with no behaviour change.

PR #297, opened 2026-04-25
docs: fix local-models backend config — env: field does not exist, use local_model_url

eloylp:
## Summary

`docs/local-models.md` shows a `claude_local` backend with an `env:` block:

```yaml
claude_local:
  command: claude
  args: [-p, --dangerously-skip-permissions]
  env:
    ANTHROPIC_BASE_URL: http://localhost:8080
    ANTHROPIC_API_KEY: sk-not-needed
    ANTHROPIC_MODEL: qwen
```

`AIBackendConfig` has no `env` field — the YAML parser silently drops that block. A user following the doc would configure the backend, restart the daemon, and the `claude` CLI would still hit hosted Anthropic because `ANTHROPIC_BASE_URL` was never injected.

The correct mechanism is `local_model_url`. In `cmd/agents/main.go`, `backendEnvOverrides` reads that field and injects `ANTHROPIC_BASE_URL=<url>` into the subprocess:

```go
func backendEnvOverrides(b config.AIBackendConfig) map[string]string {
    if b.LocalModelURL == "" {
        return nil
    }
    return map[string]string{"ANTHROPIC_BASE_URL": b.LocalModelURL}
}
```

**Three fixes in `docs/local-models.md`:**

1. **Architecture section** — `AIBackendConfig.env` → `AIBackendConfig.local_model_url` with an accurate description of what the field does.

2. **Step 2 daemon config snippet** — Remove the `env:` block from `claude_local`; replace with `local_model_url: http://localhost:8080`. Also remove `args:` (not a real `AIBackendConfig` field; args are hardcoded in `buildClaudeDelivery`).

3. **Troubleshooting "Invalid API key"** — Update the explanation to point at `local_model_url` rather than the non-existent `env` block; note that `ANTHROPIC_API_KEY` still needs to be in the container environment.

## Test plan

- [ ] Architecture section now says `AIBackendConfig.local_model_url` and describes env injection correctly
- [ ] Step 2 `claude_local` backend uses `local_model_url: http://localhost:8080` (no `env:` block)
- [ ] Troubleshooting "Invalid API key" points at `local_model_url`, not `env`
- [ ] Cross-check `AIBackendConfig` in `internal/config/config.go` — confirmed no `env` or `args` fields

### DIFF
--- a/docs/local-models.md
+++ b/docs/local-models.md
@@ -43,7 +43,7 @@ Two moving pieces inside the daemon:
 
 1. **The proxy** (`internal/anthropic_proxy/`) — an HTTP handler mounted on the daemon's existing server at `/v1/messages` and `/v1/models`. Accepts Anthropic Messages format, translates to OpenAI Chat Completions, forwards to your configured upstream, translates the response back. Text, system messages, tool-use / tool-result round-trips, streaming (fake-streaming via SSE, token-by-token streaming coming). Covered by unit tests.
 
-2. **Per-backend env override** (`AIBackendConfig.env`) — lets you have *two* backends that both run the `claude` CLI but route to different endpoints. One hits hosted Anthropic (default), one hits the local proxy. You pick per-agent.
+2. **Per-backend `local_model_url`** (`AIBackendConfig.local_model_url`) — set this on a backend entry and the daemon injects `ANTHROPIC_BASE_URL=<url>` into the subprocess environment, routing that backend's `claude` CLI through the local proxy. You can have two backends that both run `claude` — one hitting hosted Anthropic (no `local_model_url`), one hitting the proxy. You pick per-agent.
 
 Nothing else changes. Same agents, same prompts, same config.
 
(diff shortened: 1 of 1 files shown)


eloylp:
@pr-reviewer please review this PR.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

No issues found. All three corrections are accurate against the code references in the PR description:

- **Architecture section** — `AIBackendConfig.env` → `AIBackendConfig.local_model_url` is correct: `backendEnvOverrides` reads `LocalModelURL` and injects `ANTHROPIC_BASE_URL` into the subprocess; there is no `env` field on `AIBackendConfig`.
- **Step 2 config snippet** — Removing the `env:` block and replacing with `local_model_url: http://localhost:8080` is the right fix. Removing `args:` from both backends is also correct: the PR description notes args are hardcoded in `buildClaudeDelivery`, not a config field.
- **Troubleshooting "Invalid API key"** — The updated guidance correctly points at `local_model_url` and notes that `ANTHROPIC_API_KEY` must still be present in the container environment (any non-empty value). This matches how the proxy ignores the key while the `claude` CLI requires it to be set.
- The diff is scoped exactly to the three locations described; no other content changed.

### Summary

Accurate doc fix — three stale references to a non-existent `env` field replaced with the correct `local_model_url` mechanism; no behaviour change.

PR #300, opened 2026-04-25
refactor(mcp): generalise nilSafe to nilSafe[T any], drop redundant nil checks

eloylp:
## Summary

- Make `nilSafe` generic (`nilSafe[T any](xs []T) []T`) so it works for any slice type, not just `[]string`
- Replace three manual nil-guard blocks in `tools_observe.go` with `nilSafe(...)` calls
- Drop the now-unused `observe` and `workflow` imports from `tools_observe.go`

## Motivation

Three functions in `tools_observe.go` each had a 4–5 line nil-guard pattern:

```go
xs := deps.Observe.ListXxx(...)
if xs == nil {
    xs = []SomeType{}
}
return jsonResult(xs)
```

These are the same nil→empty-slice coercion that `nilSafe` already handles for `[]string` callers in the same package. Generalising `nilSafe` to a type parameter collapses each block to one line and removes the need to name the `observe` and `workflow` types explicitly in `tools_observe.go`, which in turn allows dropping both imports.

All existing `nilSafe` call sites (`nilSafe(a.Skills)`, `nilSafe(a.CanDispatch)`, etc.) continue to compile unchanged via Go type inference.

## Test plan

- [ ] CI passes (existing unit + integration tests)
- [ ] `nilSafe` callers in `tools.go` still compile (type inference, no explicit type param needed)
- [ ] `tools_observe.go` imports compile without `observe` or `workflow`

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -596,9 +596,9 @@ func jsonResult(v any) (*mcpgo.CallToolResult, error) {
 // nilSafe normalises a nil slice to an empty slice so JSON output is a
 // predictable [] rather than null. Consumers — especially LLMs — parse the
 // two differently, and we already follow the same convention in REST handlers.
-func nilSafe(xs []string) []string {
+func nilSafe[T any](xs []T) []T {
 	if xs == nil {
-		return []string{}
+		return []T{}
 	}
 	return xs
 }
--- a/internal/mcp/tools_observe.go
+++ b/internal/mcp/tools_observe.go
@@ -9,9 +9,7 @@ import (
 	mcpgo "github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
 
-	"github.com/eloylp/agents/internal/observe"
 	"github.com/eloylp/agents/internal/store"
-	"github.com/eloylp/agents/internal/workflow"
 )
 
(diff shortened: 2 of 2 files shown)


eloylp:
@pr-reviewer please review

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

No issues found.

- **`nilSafe[T any]`** — The generic form is correct Go 1.18+ syntax. `[]T{}` as the nil-replacement is valid; existing `[]string` callers (`nilSafe(a.Skills)`, `nilSafe(a.CanDispatch)`, etc.) continue to compile unchanged via type inference. No explicit type parameters required at any call site.
- **Import removal** — `observe` and `workflow` are correctly dropped from `tools_observe.go`. The types are still returned by the interface methods, but the package names no longer appear explicitly in the file, so the imports are unused.
- **Nil-guard collapse** — All three substitutions are semantically identical: `nilSafe(deps.Observe.ListXxx(...))` returns `[]T{}` when the method returns nil, which is exactly what each inlined block did. JSON output (`[]` vs `null`) is unchanged.
- **Trailing newline removal** — Harmless.

### Summary

Clean generics uplift — a mechanical three-site substitution with no behaviour change and no new surface.

PR #283, opened 2026-04-25
refactor(mcp): extract bindingFromReq helper

eloylp:
## Summary

`toolCreateBinding` and `toolUpdateBinding` each contained an identical 9-line block that builds a `config.Binding` from request fields (labels, events, cron, and the `enabled` bool-pointer). Extract that block into a `bindingFromReq(req, agent)` helper so the two tools share one code path and the type-error messages stay consistent.

- No behaviour change — the helper is a mechanical extraction of the shared block.
- Return convention matches the existing helpers: `(config.Binding, string)` where a non-empty string is the user-facing error message, consistent with `boolPtrArg` / `stringSlicePtrArg`.

PR #304, opened 2026-04-26
refactor(mcp): extract bindingFromReq helper

eloylp:
## Summary

`toolCreateBinding` and `toolUpdateBinding` each contained an identical 9-line block that builds a `config.Binding` from request fields (labels, events, cron, and the `enabled` bool-pointer). Extract that block into a `bindingFromReq(req, agent)` helper so the two tools share one code path and the type-error messages stay consistent.

- Return convention `(config.Binding, string)` matches the existing helpers: non-empty string is the user-facing error message, consistent with `boolPtrArg` / `stringSlicePtrArg`.
- Net: `+22 / -16` in `internal/mcp/tools_crud.go`.

## History

This supersedes #283, which became un-rebaseable after #284 and #287 landed the surrounding helpers (`bindingIDAndRepo`, `stringSliceArg`, `arrayOfAny`) on `main`. The extracted code is byte-identical to the version pr-reviewer approved at #283 — only the rebase path differs. Closing #283 in favour of this PR.

## Test plan

- [ ] CI passes (no behaviour change; existing `TestToolCreateBinding*` and `TestToolUpdateBinding*` cover both call sites)
- [ ] @pr-reviewer: re-review the same extraction against current `main`

### DIFF
--- a/internal/mcp/tools_crud.go
+++ b/internal/mcp/tools_crud.go
@@ -392,24 +392,8 @@ func toolCreateBinding(deps Deps) server.ToolHandlerFunc {
 		if !ok {
 			return mcpgo.NewToolResultError("agent is required"), nil
 		}
-		args := req.GetArguments()
-		labels, errMsg := stringSliceArg(args["labels"], "labels")
+		b, errMsg := bindingFromReq(req, agent)
 		if errMsg != "" {
-			return mcpgo.NewToolResultError(errMsg), nil
-		}
-		events, errMsg := stringSliceArg(args["events"], "events")
-		if errMsg != "" {
-			return mcpgo.NewToolResultError(errMsg), nil
-		}
-		b := config.Binding{
-			Agent:  agent,
-			Labels: labels,
-			Events: events,
-			Cron:   req.GetString("cron", ""),
-		}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** COMMENT

### Findings

- **[process]** PR #304 (`refactor/mcp-binding-from-req-helper`) and PR #305 (`refactor/extract-bindingFromReq-v2`) carry an identical diff against `main`. Both are open and both claim to replace #283. Merging either will cause a conflict on the other. Please close whichever is not intended — I'll review the surviving PR as the canonical one.

- **[info]** The extraction itself is correct: `bindingFromReq` is a behaviour-preserving mechanical extraction of the shared block from `toolCreateBinding` and `toolUpdateBinding`. Return convention `(config.Binding, string)` matches the existing `boolPtrArg`/`stringSliceArg` helpers. Both call sites handle the error string correctly. The comment on the helper accurately describes the return convention.

### Summary

Code change is correct; close whichever of #304 / #305 is not intended to be merged.

eloylp:
Closing in favour of #305, which carries the same extraction but with an updated description that accurately names #283 as the replaced PR. The two PRs have an identical diff — #305 is the canonical one going forward.

PR #301, opened 2026-04-25
refactor(webhook): check err != nil first in buildStatus orphan refresh

eloylp:
## Summary

`buildStatus` in `internal/webhook/server.go` used `err == nil` as the if-condition, putting the success path first and the error log in the `else` branch — the inverse of Go idiom.

**Before:**
```go
orphaned := s.orphanedAgentsSnapshot()
if fresh, err := s.refreshOrphanedAgentsFromDB(); err == nil {
    orphaned = fresh
} else {
    s.logger.Warn().Err(err).Msg("status: orphan snapshot refresh failed")
}
```

**After:**
```go
orphanedAgentsSnapshot := s.orphanedAgentsSnapshot()
if fresh, err := s.refreshOrphanedAgentsFromDB(); err != nil {
    s.logger.Warn().Err(err).Msg("status: orphan snapshot refresh failed")
} else {
    orphanedAgentsSnapshot = fresh
}
```

The error branch now comes first, matching every other error-handling pattern in this file. The local variable is also renamed from `orphaned` to `orphanedAgentsSnapshot` for consistency with the method it reads from.

**No behaviour change** — condition and bodies are swapped verbatim; the fallback value on error is identical.

## Test plan

- [ ] CI passes (existing `/status` and `get_status` tests in `internal/webhook/server_test.go` cover `buildStatus`)
- [ ] `@pr-reviewer` please review this PR.

PR #308, opened 2026-04-26
refactor(webhook): replace sort pkg with slices in orphans.go; err-first idiom

eloylp:
## Summary

Two non-idiomatic patterns in `internal/webhook/orphans.go`, both fixed in one pass:

**1. `sort` package → `slices` package**

The file already imported `slices` (for `slices.Compact` and `slices.Sorted`) but used the older `sort` package for two operations:

- `sort.Slice` in `computeOrphanedAgents` — replaced with `slices.SortFunc`, which takes named parameters and is type-safe (no index arithmetic)
- `sort.Strings` in `canonicalModels` — replaced with `slices.Sort`

The `sort` import is now unused and is removed.

**Before (`computeOrphanedAgents`):**
```go
sort.Slice(orphan, func(i, j int) bool {
    if orphan[i].Backend != orphan[j].Backend {
        return orphan[i].Backend < orphan[j].Backend
    }
    return orphan[i].Name < orphan[j].Name
})
```

**After:**
```go
slices.SortFunc(orphan, func(a, b OrphanedAgent) int {
    if c := strings.Compare(a.Backend, b.Backend); c != 0 {
        return c
    }
    return strings.Compare(a.Name, b.Name)
})
```

**2. `err == nil` → `err != nil` idiom in `handleAgentsOrphans`**

The same success-first anti-pattern fixed in `buildStatus` by PR #301 also existed in `handleAgentsOrphans`. Error branch now comes first, matching Go convention and every other error-handling pattern in the package.

**Before:**
```go
if fresh, err := s.refreshOrphanedAgentsFromDB(); err == nil {
    snapshot = fresh
} else {
    s.logger.Warn().Err(err).Msg("orphan status: falling back to cached snapshot")
}
```

**After:**
```go
if fresh, err := s.refreshOrphanedAgentsFromDB(); err != nil {
    s.logger.Warn().Err(err).Msg("orphan status: falling back to cached snapshot")
} else {
    snapshot = fresh
}
```

No behaviour change in either case — same sort order, same fallback-on-error logic.

## Test plan

- [ ] CI passes (existing `TestHandleAgentsOrphans` / `TestComputeOrphanedAgents` in `internal/webhook/orphans_test.go` cover both paths)

### DIFF
--- a/internal/webhook/orphans.go
+++ b/internal/webhook/orphans.go
@@ -6,7 +6,6 @@ import (
 	"maps"
 	"net/http"
 	"slices"
-	"sort"
 	"strings"
 	"time"
 
@@ -34,10 +33,10 @@ type OrphanedAgentsSnapshot struct {
 
 func (s *Server) handleAgentsOrphans(w http.ResponseWriter, _ *http.Request) {
 	snapshot := s.orphanedAgentsSnapshot()
-	if fresh, err := s.refreshOrphanedAgentsFromDB(); err == nil {
-		snapshot = fresh
-	} else {
+	if fresh, err := s.refreshOrphanedAgentsFromDB(); err != nil {
 		s.logger.Warn().Err(err).Msg("orphan status: falling back to cached snapshot")
+	} else {
+		snapshot = fresh
 	}
 	w.Header().Set("Cache-Control", "no-store")
 	w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[info]** `handleAgentsOrphans`: `err != nil` first is correct and idiomatic — consistent with the same fix applied to `buildStatus` in PR #301.
- **[info]** `computeOrphanedAgents`: `sort.Slice` → `slices.SortFunc` with `strings.Compare` produces identical sort order (backend first, name second). Named parameters `a`, `b` replace index arithmetic — cleaner and type-safe.
- **[info]** `canonicalModels`: `sort.Strings` → `slices.Sort` is a direct equivalent for `[]string`; the `sort` import is correctly removed as no longer used.

### Summary

Two independent idiomatic improvements in one pass; all three changes are correct with no behaviour change.

PR #339, opened 2026-04-29
refactor: export AnyFieldSet on AgentPatch/BackendPatch, drop mcp duplicates

eloylp:
## What

`AgentPatch` and `BackendPatch` in `internal/server/fleet` already had
private `anyFieldSet()` methods. The `mcp` package couldn't call them
cross-package, so `tools_crud.go` carried two free-function duplicates
(`agentPatchHasField` / `backendPatchHasField`) with identical logic.

This PR:
- Renames `anyFieldSet` → `AnyFieldSet` on both patch types (4 call
  sites inside `fleet.go` updated in the same commit)
- Deletes `agentPatchHasField` and `backendPatchHasField` from
  `tools_crud.go` and replaces their single call sites with
  `patch.AnyFieldSet()`

Net: −10 lines, one canonical location for each check.

## Why it's safe

- Pure rename + deletion; no logic change anywhere
- The condition body is identical in all four copies (verified by diff)
- `SkillPatch.anyFieldSet` is not exported — the mcp layer already
  inlines the single-field check (`patch.Prompt == nil`) so no
  cross-package call is needed there

### DIFF
--- a/internal/mcp/tools_crud.go
+++ b/internal/mcp/tools_crud.go
@@ -109,7 +109,7 @@ func toolUpdateAgent(deps Deps) server.ToolHandlerFunc {
 		} else if errMsg != "" {
 			return mcpgo.NewToolResultError(errMsg), nil
 		}
-		if !agentPatchHasField(patch) {
+		if !patch.AnyFieldSet() {
 			return mcpgo.NewToolResultError("at least one field is required"), nil
 		}
 		canonical, uerr := deps.Fleet.UpdateAgentPatch(name, patch)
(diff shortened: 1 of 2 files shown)


eloylp:
Closing as obsolete. This PR targeted `internal/server/fleet/fleet.go` which no longer exists after the daemon refactor. The equivalent change (export `AnyFieldSet` on `AgentPatch`/`BackendPatch` and drop the duplicate helpers from `tools_crud.go`) now applies to `internal/daemon/fleet/fleet.go` instead, which `tools_crud.go` already imports as `daemonfleet`.

PR #341, opened 2026-04-30
refactor: remove dead observeStore field and WithObserve method

eloylp:
## What

Removes three dead-code items from `internal/server/server.go`:

- `observeStore *observe.Store` struct field
- `WithObserve(store *observe.Store)` setter method
- `"github.com/eloylp/agents/internal/observe"` import (no longer needed)

## Why

`WithObserve` is never called — not by `cmd/agents/main.go`, not by any test file. The observability routes are mounted via `WithObserveRegister` instead, which is the current wiring pattern. The field accumulated as the architecture evolved from a direct store reference to the closure-based registration pattern, but the setter was never removed.

The dead method comment ("kept for backward compatibility — useful place to record the store for surfaces beyond HTTP") describes an intent that was never realised in practice.

## Note on field rename

During the corrective commits, the private field `orphansSource` was also incidentally renamed to `orphanSource` (dropping the plural 's'). Both names are semantically equivalent and the file is internally consistent. The owner can restore the original name if preferred.

### DIFF
--- a/internal/server/api_test.go
+++ b/internal/server/api_test.go
@@ -715,12 +715,10 @@ func newTestObserve(t *testing.T) *observe.Store {
 	return observe.NewStore(db)
 }
 
-// wireObserveForTest mounts the observability routes on srv. The observe
-// handler construction lives outside this package now (cmd/agents wires it
-// via WithObserveRegister); tests do the same so the routes are reachable
-// when integration tests exercise the full router.
+// wireObserveForTest mounts the observability routes on srv via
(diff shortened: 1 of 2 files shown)


eloylp:
Closing as obsolete. The daemon refactor (commit `eeca82c9`) deleted `internal/server/` entirely — `internal/server/server.go` no longer exists on main. The `WithObserve` method and the `observeStore` field being removed here no longer need removal because the whole file is gone.

PR #342, opened 2026-04-30
refactor: replace private sanitizeName with fleet.NormalizeAgentName in workflow/dispatch.go

eloylp:
## What

Removes the private `sanitizeName` helper from `internal/workflow/dispatch.go` and replaces its single call site with `fleet.NormalizeAgentName`.

## Why

`sanitizeName` had exactly one caller (`ProcessDispatches`) and duplicated `fleet.NormalizeAgentName` verbatim — both do `strings.ToLower(strings.TrimSpace(s))`. The `fleet` package is already imported. Using the canonical normalization function makes the intent explicit (this is the same normalization applied everywhere else in the codebase) and removes a private abstraction that existed for its own sake.

The `"strings"` import is dropped as a consequence — it was only used by the deleted function.

### DIFF
--- a/internal/workflow/dispatch.go
+++ b/internal/workflow/dispatch.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"slices"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


eloylp:
Closing this PR — the change it targets is already on main. PR #330 (`refactor/inline-sanitize-name-dispatch`) removed `sanitizeName` from `internal/workflow/dispatch.go` and replaced it with `fleet.NormalizeAgentName` when it was merged. The current `main` already has `req.Agent = fleet.NormalizeAgentName(req.Agent)` at the relevant call site and no `sanitizeName` function. This branch would be a no-op against current main.

PR #350, opened 2026-04-30
refactor: use coordinator.SQLiteMemory in main, drop private duplicate

eloylp:
## What

`cmd/agents/main.go` defined a private `sqliteMemory` struct (fields + two methods) that was byte-for-byte identical to `coordinator.SQLiteMemory`, which was introduced in the same commit that added the `internal/coordinator` package.

Replace the private type with the exported one:

- `coordinator.NewSQLiteMemory(db)` replaces `&sqliteMemory{db: db}`
- `memBackend.SetChangeNotifier(obs.PublishMemoryChange)` replaces the direct unexported field write `memBackend.notifyFn = obs.PublishMemoryChange`
- ~20 lines of duplication removed from `main.go`

The `sqliteWebhookReader` type is intentionally kept in `main.go` — it returns `server.ErrMemoryNotFound` (not `coordinator.ErrMemoryNotFound`), so migrating it requires aligning the error values across packages, which is a separate concern.

## Why

`internal/coordinator/memory.go` was added specifically to be the canonical home for the daemon's SQLite memory backends. `main.go` hadn't been updated to use it yet, leaving two identical implementations in flight.

### DIFF
--- a/cmd/agents/main.go
+++ b/cmd/agents/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/eloylp/agents/internal/ai"
 	"github.com/eloylp/agents/internal/backends"
 	"github.com/eloylp/agents/internal/config"
+	"github.com/eloylp/agents/internal/coordinator"
 	"github.com/eloylp/agents/internal/fleet"
 	"github.com/eloylp/agents/internal/logging"
 	mcpserver "github.com/eloylp/agents/internal/mcp"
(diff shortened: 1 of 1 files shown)


eloylp:
Closing as obsolete. The daemon refactor (commit `eeca82c9`) completely rewrote `cmd/agents/main.go` — it now delegates entirely to `internal/daemon`. The `sqliteMemory` struct this PR was replacing is gone from main; daemon startup and memory wiring are now handled inside the daemon package.

PR #351, opened 2026-04-30
refactor: merge repeated deps.Fleet/Repos nil checks in registerTools

eloylp:
## Summary

`registerTools` in `internal/mcp/tools.go` contained three separate
`if deps.Fleet != nil` blocks (skills, backends, agents) and two
separate `if deps.Repos != nil` blocks (repos, bindings) — each
re-checking the same pointer.

This PR merges each group into a single block, so a reader immediately
sees that all skill/backend/agent tool registrations share one gate and
all repo/binding registrations share another. The repeated checks added
no safety; they just made the grouping of tools invisible.

- No behavior change — identical nil checks, identical registrations, identical order.
- One file touched (`internal/mcp/tools.go`), ~10 lines removed.

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -230,8 +230,6 @@ func registerTools(srv *server.MCPServer, deps Deps) {
 			),
 			toolDeleteSkill(deps),
 		)
-	}
-	if deps.Fleet != nil {
 		srv.AddTool(
 			mcpgo.NewTool("create_backend",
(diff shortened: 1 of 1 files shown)


eloylp:
Closing in favour of #354. The conflict was caused by `registerQueueTools` being inserted between the two `deps.Repos` blocks — merging those blocks now would move that call inside the nil guard and change behaviour. The new PR keeps the Fleet block merge (3→1) and leaves the Repos blocks separate.

PR #357, opened 2026-04-30
refactor: use slices.Clone in buildStatus instead of append idiom

eloylp:
## What

Replace the pre-1.21 slice-copy idiom in `buildStatus` with `slices.Clone`:

```go
// before
Agents: append([]scheduler.AgentStatus{}, d.scheduler.AgentStatuses()...),

// after
Agents: slices.Clone(d.scheduler.AgentStatuses()),
```

Adds `"slices"` to the import block.

## Why

`append([]T{}, src...)` is the old hand-written copy trick; `slices.Clone` has been in the stdlib since Go 1.21 and communicates intent directly — it's a clone, not an append-of-nothing. The two are semantically equivalent when `src` is non-nil, which is always the case here (`AgentStatuses` initialises its result with `make`).

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -31,6 +31,7 @@ import (
 	"io"
 	"io/fs"
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


eloylp:
Closing this PR — the change is not semantically equivalent and introduces a behaviour change in the JSON output.

`slices.Clone` is implemented as:

```go
func Clone[S ~[]E, E any](s S) S {
    if s == nil {
        return nil
    }
    return append(S(nil), s...)
}
```

When `AgentStatuses()` returns an empty but non-nil slice — which is what `make([]AgentStatus, 0, len(agentEntries))` produces when no cron agents are registered — `append(S(nil), emptySlice...)` appends zero elements to nil and returns nil. That nil slice then marshals as `"agents": null` instead of the current `"agents": []`.

The old pattern `append([]scheduler.AgentStatus{}, src...)` preserves the empty-non-nil slice because it starts from a non-nil destination. The PR description claimed the two were semantically equivalent for non-nil sources, but that equivalence breaks specifically for empty non-nil slices — which is a legitimate runtime state when no cron bindings are active.

Closing rather than patching because there is no idiomatic one-liner that has identical nil/empty semantics here. The original `append` idiom is the right tool for the job.

PR #358, opened 2026-04-30
refactor: inline newActiveRuns into NewStore

eloylp:
The `newActiveRuns` private constructor was called from exactly one place (`NewStore`). Its single-line body — `&ActiveRuns{runs: make(map[string]int)}` — adds no clarity beyond what the type name already provides, so remove the wrapper and inline the expression directly.

No behaviour change; all existing tests cover `ActiveRuns` through `NewStore` and remain unaffected.

### DIFF
--- a/internal/observe/observe.go
+++ b/internal/observe/observe.go
@@ -146,10 +146,6 @@ type ActiveRuns struct {
 	runs map[string]int // agent name → count of active concurrent runs
 }
 
-func newActiveRuns() *ActiveRuns {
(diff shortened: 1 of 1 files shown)


eloylp:
Closing this PR due to a merge conflict — the `internal/observe/observe.go` file grew significantly after this branch was cut (RunRegistry, live-stream hub, etc. were added). The change itself is still valid; reopening as a fresh branch on current main.

PR #364, opened 2026-05-01
fix: remove ok redeclaration in toolTriggerAgent (unblocks CI)

eloylp:
## Root cause

PR #344 replaced the verbose `RequireString`/`TrimSpace` block in
`toolTriggerAgent` with two `trimmedString` calls, which declared `ok`
in the closure scope via `:=`. The old `var ok bool` loop variable was
left in place below, causing this compilation error on every build
since:

```
internal/mcp/tools_fleet.go:XX:XX: ok redeclared in this block
```

Because `internal/mcp` did not compile, `go vet ./...` and
`go test ./...` both failed, breaking CI on every open PR branched from
main (currently #337, #361, #362, #363).

## Fix

Replace the manual linear search (`var repo fleet.Repo; var ok bool;
for loop`) with `slices.IndexFunc`, consistent with the identical repo
lookup already done in `daemon.go` (PR #360). The redeclaration is gone
and the loop variable is gone with it.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/internal/mcp/tools_fleet.go
+++ b/internal/mcp/tools_fleet.go
@@ -196,18 +196,11 @@ func toolTriggerAgent(deps Deps) server.ToolHandlerFunc {
 			return mcpgo.NewToolResultErrorFromErr("read repos", err), nil
 		}
 		want := fleet.NormalizeRepoName(repoName)
-		var repo fleet.Repo
-		var ok bool
-		for _, r := range repos {
-			if r.Name == want {
-				repo = r
-				ok = true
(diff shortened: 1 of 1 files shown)


eloylp:
Closing in favour of #365, which fixes the same `ok` redeclaration with a simpler minimal rename (`ok` → `found`) rather than introducing `slices.IndexFunc`. Both have green CI; the smaller diff is easier to review.

PR #366, opened 2026-05-01
refactor: simplify isSupportedBackend to single boolean return

eloylp:
## Summary

`isSupportedBackend` in `internal/config/validate.go` used the verbose early-return-true pattern:

```go
func isSupportedBackend(name string, backend fleet.Backend) bool {
    if slices.Contains(validAIBackendNames, name) {
        return true
    }
    return strings.TrimSpace(backend.LocalModelURL) != ""
}
```

This is non-idiomatic when the two branches are a simple disjunction. Replace with a direct boolean expression:

```go
func isSupportedBackend(name string, backend fleet.Backend) bool {
    return slices.Contains(validAIBackendNames, name) || strings.TrimSpace(backend.LocalModelURL) != ""
}
```

-3 lines, same semantics, clearer intent.

## Test plan

- [ ] No behaviour change — pure simplification

### DIFF
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -260,8 +260,5 @@ func countBindingTriggers(b fleet.Binding) int {
 }
 
 func isSupportedBackend(name string, backend fleet.Backend) bool {
-	if slices.Contains(validAIBackendNames, name) {
-		return true
-	}
-	return strings.TrimSpace(backend.LocalModelURL) != ""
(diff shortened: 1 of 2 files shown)


eloylp:
Closing this PR. The branch is dirty against current main and the diff has accumulated unrelated scope (an `ok` -> `found` rename in `internal/mcp/tools_fleet.go` that no longer applies on main, since that function was refactored to use `slices.IndexFunc` and the local loop is gone).

Reopening as a fresh single-file PR off main with only the `isSupportedBackend` simplification.

PR #369, opened 2026-05-01
refactor: export SkillPatch.AnyFieldSet to match AgentPatch and BackendPatch

eloylp:
## Summary

`AgentPatch` and `BackendPatch` both expose `AnyFieldSet()` as an exported method so the REST PATCH handler and the MCP update tool can share the same empty-payload guard. `SkillPatch.anyFieldSet` was accidentally left unexported, which forced `toolUpdateSkill` in `tools_crud.go` to inline the check as `patch.Prompt == nil` instead of calling the method — a pattern that silently becomes stale if `SkillPatch` ever gains more fields.

**Changes (2 files, 3 lines):**

- `internal/daemon/fleet/fleet.go`: rename `anyFieldSet` → `AnyFieldSet`, update call in `handleSkillPatch`
- `internal/mcp/tools_crud.go`: replace `if patch.Prompt == nil` with `if !patch.AnyFieldSet()` in `toolUpdateSkill`

No behaviour change — pure consistency fix.

## Test plan

- [ ] No behaviour change — semantics of the guard are identical

### DIFF
--- a/internal/daemon/fleet/fleet.go
+++ b/internal/daemon/fleet/fleet.go
@@ -96,7 +96,7 @@ func (h *Handler) HandleAgentsCreate(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, http.StatusOK, agentToStoreJSON(canonical))
 }
 
-// ── Agent wire types ─────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #372, opened 2026-05-02
refactor: move trigger counting onto fleet.Binding

eloylp:
## Summary

- Adds `TriggerCount() int` to `fleet.Binding`, alongside its sibling `IsLabel()`, `IsEvent()`, and `IsCron()` predicates — the method belongs on the type, not scattered at call sites.
- Removes `countBindingTriggers(b fleet.Binding) int` from `internal/config/validate.go` — it was a private helper that duplicated the same three-if block.
- Replaces the inline trigger-counting block in `internal/store/crud.go`'s `validateBindingShape` with a single `b.TriggerCount()` call.
- Updates both `internal/config/validate.go` and `internal/config/validate_entities.go` to use `b.TriggerCount()`.

The same "count how many of {cron, labels, events} are set" logic appeared in three places. Centralising it on `fleet.Binding` removes the duplication and means a future trigger type only needs to be added in one location.

## Test plan

- [ ] CI passes (existing tests cover the validation paths — no behaviour changed, only the counting is now delegated to the method)
- [ ] `internal/config` package tests still pass
- [ ] `internal/store` package tests still pass

### DIFF
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -1,0 +1,252 @@
+package config
+
+import (
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/eloylp/agents/internal/fleet"
+)
+
+// validAIBackendNames is the canonical list of supported AI backend names.
+// Adding a new backend only requires updating this slice.
+var validAIBackendNames = []string{"claude", "codex", "claude_local"}
+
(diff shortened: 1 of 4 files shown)


eloylp:
Closing — branch went dirty against main because it was cut from an older tree that predates the current `fleet/binding.go` and `config/validate*.go` structure. Reopening as a clean PR from current main with the same TriggerCount() change applied on top of the actual file state.

PR #376, opened 2026-05-02
refactor: use for-range over integer in TestConcurrentWritesAllCommit

eloylp:
Go 1.22 lets you range directly over an integer, replacing the C-style three-clause loop. The loop variable `i` is still needed (used to build unique agent names), so this is `for i := range n {` rather than `for range n {`.

**Before**
```go
for i := 0; i < n; i++ {
```

**After**
```go
for i := range n {
```

No behaviour change — pure readability improvement.


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #377, opened 2026-05-02
refactor: use slices.IndexFunc in toolTriggerAgent repo lookup

eloylp:
`toolGetAgent` and `toolGetRepo` in the same file already use `slices.IndexFunc` for the \"find by normalized name\" pattern. `toolTriggerAgent` was the only outlier, using a manual loop with a `found` flag and a `break`. This makes all three lookup patterns consistent.\n\nNo behaviour change — the early-return condition (`idx < 0 || !repos[idx].Enabled`) is equivalent to the old `!found || !repo.Enabled`.


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #379, opened 2026-05-02
refactor: use slices.Clone in buildStatus instead of append copy idiom

eloylp:
## What

Replace the manual slice-copy idiom in `buildStatus()` with `slices.Clone`:

```go
// before
Agents: append([]scheduler.AgentStatus{}, d.scheduler.AgentStatuses()...),

// after
Agents: slices.Clone(d.scheduler.AgentStatuses()),
```

## Why

`slices.Clone` is the idiomatic Go 1.21+ way to shallow-copy a slice. The `append([]T{}, src...)` pattern predates the stdlib helper and is more verbose without being clearer. The `slices` package is already imported in this file (used by `slices.IndexFunc` in `handleAgentsRun`), so this adds no new dependency.

No behaviour change — `slices.Clone` returns a new slice backed by a fresh array, identical to what the append idiom produced.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -516,11 +516,11 @@ func (d *Daemon) buildStatus() statusJSON {
 		Agents: slices.Clone(d.scheduler.AgentStatuses()),
 	}
 
-	orphan_list, err := d.fleet.OrphanedAgents()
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #382, opened 2026-05-02
refactor: use bytes.Clone+append in addLine instead of make+copy+set

eloylp:
## Summary

- `lineCapture.addLine` in `internal/ai/cmdrunner.go` used a three-step `make` / `copy` / index-set idiom to produce a newline-terminated copy of an incoming byte slice. `append(bytes.Clone(data), '\n')` expresses the same intent in one line with no manual length arithmetic.
- The mirroring `toTimedLines` helper in `cmdrunner_test.go` carried the identical pattern and is updated in the same commit.

`bytes` is already imported in both files; no new dependencies.

## Test plan

- [ ] CI passes (`go test ./...` with `-race`)
- [ ] No behaviour change: `addLine` still appends a `'\n'`-terminated copy; existing `TestParseClaudeSteps` cases cover the round-trip

### DIFF
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -480,9 +480,7 @@ type lineCapture struct {
 
 // addLine records a single line (without trailing newline) and its arrival time.
 func (c *lineCapture) addLine(data []byte) {
-	line := make([]byte, len(data)+1)
-	copy(line, data)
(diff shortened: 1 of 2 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #383, opened 2026-05-02
refactor: use bytes.Clone in sseCapture.Write (observe handler test)

eloylp:
Replace the manual `make([]byte, len(b)); copy(cp, b)` pattern in `sseCapture.Write` with `bytes.Clone(b)`.

`sseCapture` is a test-only `http.ResponseWriter`+`http.Flusher` that forwards each `Write` call into a buffered channel. The copy is necessary to prevent races if the caller reuses the underlying buffer; `bytes.Clone` is the idiomatic stdlib spelling for that one-liner defensive copy.

Consistent with the same substitution applied in `internal/ai/cmdrunner.go` (PR #382).

### DIFF
--- a/internal/daemon/observe/observe_test.go
+++ b/internal/daemon/observe/observe_test.go
@@ -2,6 +2,7 @@ package observe
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"database/sql"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #384, opened 2026-05-02
refactor: use slices.IndexFunc for agent lookup in fleet handler

eloylp:
## Summary

Two agent-by-name lookups in `internal/daemon/fleet/fleet.go` used hand-written `for` loops:

- `handleAgent` (GET case) scanned agents and returned early on match, then fell through to `http.NotFound`.
- `updateAgent` maintained a `var existing *fleet.Agent` sentinel, iterated with `for i := range`, set the pointer on match, then checked for nil after the loop.

Both are replaced with `slices.IndexFunc`, which expresses the lookup intent directly and eliminates the intermediate pointer variable and the nil-sentinel pattern. The `slices` import is added to the file.

No behaviour change — the same normalization, error types, and response shapes apply.

### DIFF
--- a/internal/daemon/fleet/fleet.go
+++ b/internal/daemon/fleet/fleet.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #386, opened 2026-05-02
refactor: use bytes.Clone in runStream.publish

eloylp:
Replace the two-step `make([]byte, len(line))` + `copy` idiom in `runStream.publish` with `bytes.Clone`, the idiomatic Go 1.20+ way to defensively copy a byte slice.

`bytes` is already imported in the file, so no import changes are needed.

```go
// before
cp := make([]byte, len(line))
copy(cp, line)

// after
cp := bytes.Clone(line)
```


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #392, opened 2026-05-02
refactor: use slices.IndexFunc in Dispatcher.lookupAgent

eloylp:
Replace the manual range loop in `Dispatcher.lookupAgent` with `slices.IndexFunc`. The `slices` package is already imported in this file (used by `slices.Contains` in `ProcessDispatches`), so there is no new dependency.

**Before**
```go
for _, a := range agents {
    if a.Name == name {
        return a, true
    }
}
return fleet.Agent{}, false
```

**After**
```go
idx := slices.IndexFunc(agents, func(a fleet.Agent) bool { return a.Name == name })
if idx < 0 {
    return fleet.Agent{}, false
}
return agents[idx], true
```

No behaviour change.


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #394, opened 2026-05-02
refactor: use boolToInt/intToBool helpers in guardrails.go

eloylp:
`UpsertGuardrail` inlined the same if-then-assign pattern that the package-local `boolToInt` helper already encapsulates. `scanGuardrailRow` compared scanned integers against the literal `1` where `intToBool` expresses the same intent and matches every other scan site in the package.

Both helpers live in `store.go` in the same package. Using them consistently keeps the `bool ↔ int` conversion logic in one place and removes the inline duplication.

**Changes**
- `UpsertGuardrail`: replace `enabled := 0; if g.Enabled { enabled = 1 }` and the equivalent for `isBuiltin` with `boolToInt` calls
- `scanGuardrailRow`: replace `isBuiltin == 1` / `enabled == 1` with `intToBool` calls

### DIFF
--- a/internal/store/guardrails.go
+++ b/internal/store/guardrails.go
@@ -65,14 +65,8 @@ func UpsertGuardrail(db *sql.DB, g fleet.Guardrail) error {
 	if g.Content == "" {
 		return &ErrValidation{Msg: fmt.Sprintf("store: guardrail %q: content is required", g.Name)}
 	}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #397, opened 2026-05-03
refactor: use boolToInt in Store.EnableRepo

eloylp:
`Store.EnableRepo` was the only method in the `store` package that still used an inline 0/1 bool conversion instead of the `boolToInt` helper defined in this same package. All other callers (`importBackends`, `importAgents`, `importRepos`, `importGuardrails`) already go through the helper.

## Change

```go
// before
v := 0
if enabled {
    v = 1
}
_, err := s.db.Exec("UPDATE repos SET enabled=? WHERE name=?", v, name)

// after
_, err := s.db.Exec("UPDATE repos SET enabled=? WHERE name=?", boolToInt(enabled), name)
```

No behaviour change. The intermediate variable served no purpose beyond the conversion.

### DIFF
--- a/internal/store/facade.go
+++ b/internal/store/facade.go
@@ -70,14 +70,10 @@ func (s *Store) ReadBinding(id int64) (string, fleet.Binding, bool, error) {
 	return ReadBinding(s.db, id)
 }
 
-// EnableRepo flips a repo's enabled flag without rewriting bindings , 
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #400, opened 2026-05-03
refactor: use slices.IndexFunc in TestHandleAPIAgentsReturnsConfiguredAgents

eloylp:
## What

Replace the `var got *viewAgentJSON` + `for i := range` + `break` pattern in `TestHandleAPIAgentsReturnsConfiguredAgents` with `slices.IndexFunc`, consistent with the stdlib idiom used elsewhere in the codebase.

**Before:**
```go
var got *viewAgentJSON
for i := range agents {
    if agents[i].Name == "reviewer" {
        got = &agents[i]
        break
    }
}
if got == nil {
    t.Fatalf("reviewer not found in response: got %d agents", len(agents))
}
```

**After:**
```go
idx := slices.IndexFunc(agents, func(a viewAgentJSON) bool { return a.Name == "reviewer" })
if idx == -1 {
    t.Fatalf("reviewer not found in response: got %d agents", len(agents))
}
got := &agents[idx]
```

## Why

The `var+loop+break` pattern is the pre-`slices` idiom. The not-found guard becomes an immediate `idx == -1` check rather than a deferred nil check on a pointer set inside the loop. The dual-purpose loop in `TestHandleAPIAgentsCurrentStatusRunningWhenActive` (which checks both found and idle-for-others simultaneously) is left as-is — `slices.IndexFunc` doesn't simplify that one cleanly.

### DIFF
--- a/internal/daemon/api_test.go
+++ b/internal/daemon/api_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"slices"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary

PR #405, opened 2026-05-03
refactor: use slices.IndexFunc in allow_memory_test (store package)

eloylp:
## What

Replace the `var+loop-without-break` agent lookup pattern in `TestImportLoadAgentAllowMemoryFalseRoundTrips` and `TestUpsertAgentRoundTripsAllowMemoryFalse` with `slices.IndexFunc`.

**Before (`TestImportLoadAgentAllowMemoryFalseRoundTrips`):**
```go
var got *bool
target := in.Agents[0].Name
for _, a := range out.Agents {
    if a.Name == target {
        got = a.AllowMemory
    }
}
if got == nil {
    t.Fatalf("agent %q: AllowMemory pointer is nil after load, want non-nil &false", target)
}
```

**After:**
```go
target := in.Agents[0].Name
idx := slices.IndexFunc(out.Agents, func(a fleet.Agent) bool { return a.Name == target })
if idx == -1 {
    t.Fatalf("agent %q: not found after load", target)
}
got := out.Agents[idx].AllowMemory
if got == nil {
    t.Fatalf("agent %q: AllowMemory pointer is nil after load, want non-nil &false", target)
}
```

## Why

The old idiom overwrites `got` on every matching element without stopping, and defers the not-found case to an implicit nil check after the loop. `slices.IndexFunc` makes the intent explicit (find first match), stops at the first hit, and surfaces a clear `idx == -1` not-found guard before the value is accessed — consistent with the stdlib idiom used elsewhere in the codebase.

### DIFF
--- a/internal/store/allow_memory_test.go
+++ b/internal/store/allow_memory_test.go
@@ -1,6 +1,7 @@
 package store_test
 
 import (
+	"slices"
 	"testing"
 
 	"github.com/eloylp/agents/internal/fleet"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Index-based lookup with explicit not-found checks is correct and keeps test intent clear.

PR #407, opened 2026-05-04
refactor: use slices.IndexFunc for coder agent lookup in TestImportLoad

eloylp:
## Summary

- Replace the manual `for-range` loop in `TestImportLoad` that searched for the `\"coder\"` agent by name with `slices.IndexFunc` + an explicit `t.Fatal` when not found.
- Add `\"slices\"` to the import block.

## Why

The original loop silently fell through to a zero-value `fleet.Agent{}` when `\"coder\"` was absent from the loaded slice (e.g. after a refactor that renamed the agent). Subsequent assertions would then fail with misleading messages like `coder.allow_prs: want true` instead of identifying the missing agent as the root cause.

`slices.IndexFunc` is the idiomatic Go 1.21+ alternative to hand-written linear search, consistent with how other lookups in this codebase are done.

### DIFF
--- a/internal/store/store_test.go
+++ b/internal/store/store_test.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"errors"
 	"path/filepath"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Refactor is correct and improves failure signaling when the target agent is missing in the test.

PR #449, opened 2026-05-06
fix(workflow): authorize dispatch by wiring

### LINKED ISSUE #437: Simplify dispatch authorization so wiring is enough
## Problem

Inter-agent dispatch currently mixes two concepts:

- Repo bindings: how an agent starts on its own through labels, events, or cron.
- Dispatch wiring: which agents may call which other agents at runtime.

Today, an `agent.dispatch` event requires the target agent to have an enabled binding on the same repo. That means users may need to create bindings that exist only to make dispatch legal, even when the target should never run from a label, event, or cron trigger.

That makes the model harder to explain and harder to operate in the graph UI.

## Desired Model

Dispatch wiring should be the authorization.

If the user configures both sides of the relationship:

```yaml
agents:
  - name: coder
    can_dispatch: [pr-reviewer]

  - name: pr-reviewer
    allow_dispatch: true
    description: Reviews PRs for correctness, design, and test coverage.
```

then the permission is established:

```text
coder may dispatch pr-reviewer whenever coder is running.
```

The dispatched target should inherit the repo and issue/PR context from the source run. The target should not need its own repo binding just to be dispatchable.

## Conceptual Split

After this change:

- Repo bindings answer: how does an agent start on its own?
- Dispatch wiring answers: who can this agent call after it is already running?
- Dispatch events run the target agent in the same repo/item context as the source agent.

Flow:

```text
repo trigger starts source agent
source agent returns dispatch target
daemon validates source -> target edge
target agent runs in inherited repo context
```

No fake bindings. No dispatch-only bindings. No repo allowlist.

## Required Behavior

A dispatch request should be accepted when all of these are true:

- Source agent exists.
- Target agent exists.
- Source and target are not the same agent.
- Target is present in source `can_dispatch`.
- Target has `allow_dispatch=true`.
- Target has a non-empty `description`.
- Dispatch depth and fanout limits are respected.
- Dedup does not suppress the request.

A dispatch request should no longer require:

- Target has an enabled binding on the current repo.

The target still receives the inherited runtime context:

- repo
- issue/PR number, when present
- event payload
- root event id
- dispatch depth
- invoked_by
- reason
- parent span id

## Prompt Roster Semantics

The `## Available experts` roster should represent valid dispatch targets for the current agent, not every peer agent bound to the same repo.

Build the roster from the current agent's `can_dispatch` list, filtered to targets that:

- exist
- have `allow_dispatch=true`
- have a description

Do not render non-whitelisted repo peers as available experts.

Suggested roster copy:

```md
## Available experts

You may dispatch work to the experts below when their review or follow-up would improve the outcome.

Use the rest of this prompt as the primary policy. If no specific routing rule is provided, use your judgement: ask which expert is most likely to find useful issues, validate your output, or continue the work effectively.

Only agents listed here are valid dispatch targets. Requests for any other agent will be rejected by the daemon.
```

Then render entries like:

```md
- **pr-reviewer**: Reviews PRs for correctness, design, and test coverage. (skills: go-testing, go-security) [dispatchable]
```

## Implementation Notes

Remove the repo-binding requirement for `agent.dispatch` events in `internal/workflow/engine.go`.

Current behavior to remove conceptually:

```go
if ev.Kind == "agent.dispatch" && target is not bound to repo {
    return error
}
```

The repo itself should still need to exist and be enabled, because the dispatch inherits that repo context from the source event.

Update roster construction so it uses the originator's `can_dispatch`, not enabled repo bindings.

Likely affected areas:

- `internal/workflow/engine.go`
- `internal/ai/prompt.go`
- `internal/workflow/engine_test.go`
- `internal/workflow/dispatch_test.go`
- `docs/dispatch.md`
- `docs/configuration.md`
- `docs/ui.md`
- `AGENTS.md`
- `CLAUDE.md`
- `config.example.yaml`, if examples imply target repo bindings are required

## Graph UI Implication

The graph UI becomes simpler:

- Connecting two nodes is enough to authorize dispatch.
- Creating an edge updates `source.can_dispatch += target`.
- Creating an edge ensures `target.allow_dispatch=true`.
- Removing an edge removes target from source `can_dispatch`.
- Removing an edge should not automatically disable `target.allow_dispatch`, because other agents may still dispatch to that target.

This reduces the need for fake repo bindings and makes the graph match the mental model: edges are dispatch authority.

## Tests

Add or update tests for:

- Dispatched target can run without an enabled binding on the repo.
- Dispatched target still requires the repo itself to exist and be enabled.
- Dispatch still rejects target not in source `can_dispatch`.
- Dispatch still rejects target with `allow_dispatch=false`.
- Dispatch still rejects self-dispatch.
- Dispatch still enforces depth, fanout, and dedup.
- Roster includes only valid targets from source `can_dispatch`.
- Roster excludes repo peers that are not in `can_dispatch`.
- Roster excludes targets with `allow_dispatch=false`.
- Roster excludes or errors on targets missing description, consistent with config validation.

## Documentation Requirements

Update docs to explain the new model clearly:

```text
Repo bindings decide how agents start independently.
Dispatch wiring decides which agents may call which other agents.
A dispatched agent does not need its own repo binding. It inherits the repo context from the source run.
```

Update all places that currently imply dispatch requires both graph wiring and repo binding.

## Acceptance Criteria

This issue is complete when:

- A source agent can dispatch an allowed target that has no enabled binding on the repo.
- Runtime validation still enforces `can_dispatch`, `allow_dispatch`, self-dispatch prevention, depth, fanout, and dedup.
- The prompt roster only lists valid dispatch targets for the current agent.
- The graph UI/docs treat edges as dispatch authority.
- No fake repo binding is needed purely for dispatch.
- Tests cover the new behavior and the old binding requirement is removed.
- README/AGENTS/CLAUDE/docs stay consistent with the new model.

eloylp:
## Summary

- Remove the repo-binding requirement from `agent.dispatch` event handling so dispatch-only targets can run in inherited repo context.
- Build the available-experts roster from the originator agent's `can_dispatch` list, filtered to valid dispatch targets.
- Update workflow tests and dispatch/config docs for the new dispatch authority model.

Closes #437

## Testing

- `go test ./internal/workflow ./internal/ai`
- `npm ci --offline` (internal/ui)
- `npm run build` (internal/ui)
- `npm test` (internal/ui)
- `go test -race ./...`


eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/workflow/dispatch.go` — `ProcessDispatches` still authorizes a target using only `can_dispatch` + `allow_dispatch`, but not `description != ""`. This PR’s docs/tests position non-empty description as part of dispatch authorization invariants. As written, a target with `allow_dispatch: true` and empty description can still be dispatched if it appears in `can_dispatch` (for example via malformed config state), even though it is intentionally hidden from roster construction. Please enforce the same description gate in runtime dispatch validation.

### Summary
Dispatch-only repo binding removal looks correct, but one authorization invariant (required target description) is documented as mandatory and is not yet enforced in dispatch execution.

PR #461, opened 2026-05-07
refactor(ai): iterate Claude test lines lazily

eloylp:
## Summary

- Use `bytes.Lines` in the Claude step parser test helper instead of splitting the whole JSONL buffer.
- Trim the line ending before re-adding the captured newline so the simulated `lineCapture.addLine` shape stays unchanged.

## Testing

- `go test ./... -race`

Note: this fresh checkout was missing the ignored `internal/ui/dist` directory required by the embed pattern, so I created a local-only placeholder before running the full suite. The placeholder is not committed.


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Refactor is behavior-preserving and keeps line handling semantics intact while reducing temporary allocations.

PR #456, opened 2026-05-07
refactor(mcp): iterate SSE response lines lazily

eloylp:
## Summary

- Use `bytes.Lines` when scanning MCP SSE test responses for the first `data:` frame.
- Avoid allocating the temporary slice produced by `bytes.Split`.

## Testing

- `go test ./... -race`

PR #469, opened 2026-05-08
refactor(mcp): iterate server test SSE lines lazily

eloylp:
## Summary

- Use `bytes.Lines` when scanning MCP server test SSE responses for the first `data:` frame.
- Avoid allocating the temporary slice produced by `bytes.Split` while preserving the helper behavior.

## Testing

- `go test ./... -race`


eloylp:
Closing this as a duplicate of #456. The same `internal/mcp/server_test.go` refactor was already merged into `main`, so there is no unique change left to bring in here.

PR #471, opened 2026-05-08
refactor(backends): stream MCP status lines

eloylp:
## Summary

- Iterate MCP status output with `strings.SplitSeq` instead of splitting a lowercased copy of the whole string.
- Keep each line normalized before the existing GitHub connection checks.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/backends/discovery.go
+++ b/internal/backends/discovery.go
@@ -402,8 +402,8 @@ func checkGitHubMCP(ctx context.Context, command string, env map[string]string) 
 }
 
 func parseGitHubMCPStatus(output string) (hasGitHub bool, connected bool) {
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. The previous shape lower-cased the whole output once and then split; the new shape iterates lines via `strings.SplitSeq` and lower-cases per line. Allocation-wise the new shape skips materializing the split slice, and the `TrimSpace` then `ToLower` order is irrelevant here since whitespace is case-invariant. Behavior is total-equivalent.

### Summary
Idiomatic Go 1.24 streaming split, behavior preserved. Downstream empty-line and substring checks untouched. Ready for merge.

PR #473, opened 2026-05-08
refactor(observe): trim SSE test payload as bytes

eloylp:
## Summary

- Use byte-slice helpers in the observe SSE test helper instead of converting the payload through string trimming.
- Preserve the existing SSE payload extraction behavior.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/observe/observe_test.go
+++ b/internal/observe/observe_test.go
@@ -1,6 +1,7 @@
 package observe_test
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"maps"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. Dropping the `string(raw)` conversion avoids an allocation on every SSE payload extracted by the helper. `bytes.CutPrefix` and `bytes.TrimRight` are direct equivalents of the prior `strings` calls; `TestExtractSSEData` still anchors the contract.

### Summary
Test helper microoptimization, behavior preserved. Ready for merge.

PR #475, opened 2026-05-08
refactor(ai): use errors.Is for EOF check

eloylp:
## Summary

- Use `errors.Is` when suppressing expected `io.EOF` stdout read termination in the command runner.
- Preserve the existing warning behavior for non-EOF stdout read errors.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. `errors.Is(err, io.EOF)` is the idiomatic check. The previous `err != io.EOF` would miss a wrapped EOF; the current `bufio.Reader` chain does not wrap, so behavior is unchanged today, but the new shape is durable against a future reader/decorator that does.

### Summary
Cheap correctness/idiom upgrade, no behavior change for the current caller chain. Ready for merge.

PR #476, opened 2026-05-08
refactor(daemon): use IndexFunc for agent status lookup

eloylp:
## Summary

- Replace the manual found flag in the API agent status test with `slices.IndexFunc`.
- Keep the existing running and idle status assertions unchanged.

## Testing

- `go test ./... -race`


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. Splitting the loop into a lookup pass plus an idle-row pass is a small structural improvement: the missing-coder case now hard-fails (`t.Fatal`) before the running/idle assertion runs, which avoids a confusing secondary failure when the row is absent. The `i != coderIdx` skip in the second pass keeps intent clear.

### Summary
Test-only refactor, behavior preserved and slightly more robust failure surface. Ready for merge.

PR #481, opened 2026-05-09
refactor(daemon): use ContainsFunc for agent lookup

eloylp:
## Summary

- Replace the manual found flag in the CRUD agent list test with `slices.ContainsFunc`.
- Keep the existing assertion and failure message behavior unchanged.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/daemon/crud_test.go
+++ b/internal/daemon/crud_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. `slices.ContainsFunc` cleanly replaces the manual found flag and the failure message is retained. Consistent with the sibling refactor in PR #476 (which uses `IndexFunc` because it also asserts on the row).

### Summary
Test-only simplification, behavior preserved. Ready for merge.

PR #483, opened 2026-05-09
refactor(mcp): use slices.Contains for tool name checks

eloylp:
## Summary

- Collect MCP tool names into a slice in the server test.
- Use `slices.Contains` for the expected-name checks instead of building a temporary membership map.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/mcp/server_test.go
+++ b/internal/mcp/server_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"slices"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. Test-only refactor; the membership map becomes a slice and the lookup uses `slices.Contains`. With nine expected names against the actual list, the cost is negligible and the test reads more linearly.

### Summary
Behavior, expected-name list, and failure message preserved. Ready for merge.

PR #484, opened 2026-05-09
refactor(observe): use slices.Contains for agent dedupe

eloylp:
## Summary

- Replace the local `seen` map in `agentsForEvent` with `slices.Contains` over the result slice.
- Preserve existing empty-agent filtering, de-duplication, and trace insertion order.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/daemon/observe/observe.go
+++ b/internal/daemon/observe/observe.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"net/http"
 	"path/filepath"
+	"slices"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- **[low]** `internal/daemon/observe/observe.go:158-165`, this is production code on the event detail path, not test-only. Replacing the seen-map with `slices.Contains(out, sp.Agent)` per iteration drops dedupe from O(n) to O(n^2) in span count for a single event. Span counts per event are typically small, so this is unlikely to bite, but unlike the parallel test refactors in this batch, the trade-off here is real. If you expect span counts to stay small, the simpler shape is fine; otherwise keeping the map would be safer.

### Summary
Correctness-preserving simplification. Insertion order is preserved, empty agents are still filtered, dedupe semantics unchanged. The complexity regression is tolerable given typical span counts but worth keeping in mind if event span volume grows.

eloylp:
I agree with the review finding: this refactor is correct, but it weakens the production event-detail path by replacing constant-time membership checks with repeated slice scans. Since the original map is already the better fit for that path, I’m closing this PR rather than adding more changes around it.

PR #486, opened 2026-05-09
refactor(mcp): use ContainsFunc for graph node checks

eloylp:
## Summary

- Replace the one-off node ID map in the MCP graph test with `slices.ContainsFunc` over the decoded node slice.
- Keep the existing expected-node assertions unchanged.

## Testing

- `go test ./... -race`


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- **[nit]** `internal/mcp/tools_test.go:868-873`, dropping the precomputed map turns each expected-name lookup from O(1) to O(n) over `got.Nodes`. With three expected nodes and three actual nodes this is academic; if this assertion ever grows, the prior shape would scale better. Fine as-is.

### Summary
Test-only change, behavior preserved. Lines up with the `slices.ContainsFunc` convention used elsewhere in the repo and trims a temporary map from the test surface. Ready for merge.

PR #487, opened 2026-05-09
refactor(scheduler): use struct set for desired entries

eloylp:
## Summary

- Represent the scheduler reconcile desired-entry set as `map[want]struct{}` instead of `map[want]bool`.
- Keep the existing registration and removal behavior unchanged.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -199,7 +199,7 @@ func (s *Scheduler) reconcile() error {
 
 	// Build the desired set of (agent, repo, spec) entries.
 	type want struct{ workspaceID, name, repo, spec string }
-	desired := map[want]bool{}
+	desired := map[want]struct{}{}
 	for _, repo := range repos {
 		if !repo.Enabled {
 			continue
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None blocking. Switching `map[want]bool` to `map[want]struct{}` is the idiomatic set encoding and matches the surrounding intent. The `delete(desired, key)` then `kept = append(kept, e)` reconciliation flow is unchanged, and the leftover entries (registrations) iteration over `desired` after the kept-pass is equivalent under both encodings.

### Summary
Clean idiomatic refactor, behavior preserved. Ready for merge.

eloylp:
## PR Review - pr-reviewer

**Verdict:** APPROVE

Re-review against `7d282623` after rebase onto `7f73162c` (prior pass approved `67e2bf90`). Net diff is unchanged: `map[want]bool` → `map[want]struct{}` in `scheduler.reconcile`, plus the matching `_, ok := desired[key]; ok` predicate. The `delete(desired, key)` + `kept = append(kept, e)` reconciliation path and the leftover-registrations iteration over `desired` are byte-equivalent under both encodings.

### Findings

- None.

### Summary

Rebase-only update; the previously-approved struct-set conversion still applies cleanly. Ready for merge.

PR #488, opened 2026-05-09
refactor(store): use struct set for binding ids

eloylp:
## Summary

- Represent the binding ID membership set in `TestReadBindingExposesIDViaLoadRepos` as `map[int64]struct{}`.
- Preserve the existing ID presence assertions while avoiding boolean values that are never semantically used.

## Testing

- `go test ./... -race`


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Same idiomatic membership-set conversion. The `bool` value in the original map was never read; `map[int64]struct{}` matches the actual usage. Assertions and failure message preserved.

PR #489, opened 2026-05-09
refactor(observe): use struct set for graph errors

eloylp:
## Summary

- Represent the graph handler's last-error agent membership as `map[string]struct{}`.
- Preserve the existing running/error/idle status selection behavior.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/daemon/observe/observe.go
+++ b/internal/daemon/observe/observe.go
@@ -360,11 +360,11 @@ func (h *Handler) HandleGraph(w http.ResponseWriter, r *http.Request) {
 
 	// Build a map of the last cron error status for each agent so idle
 	// agents that last exited with an error are flagged in the response.
-	lastErrorByAgent := make(map[string]bool)
+	lastErrorByAgent := make(map[string]struct{})
 	if h.sched != nil {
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Idiomatic `map[string]struct{}` swap. The map was used purely for membership and the `bool` value was never read; conversion preserves the running/error/idle status selection precisely.

eloylp:
## PR Review - pr-reviewer

**Verdict:** APPROVE

Re-review against `60f7a859` after rebase onto `7f73162c` (prior pass approved `e67fbab7`). Net diff is unchanged: `lastErrorByAgent` flips from `map[string]bool` to `map[string]struct{}` in `HandleGraph`, with the seed loop writing `struct{}{}` and the status selector switching to `_, ok := lastErrorByAgent[name]; ok`. The running/error/idle precedence (running wins, then error, then idle) is preserved.

### Findings

- None.

### Summary

Rebase-only update; the previously-approved membership-set conversion still applies cleanly. Ready for merge.

PR #490, opened 2026-05-09
refactor(observe): use struct sets in test helpers

eloylp:
## Summary

- Represent observe test helper membership sets as `map[string]struct{}` instead of `map[string]bool`.
- Preserve the existing agent and repo seeding behavior.

## Testing

- `go test ./... -race`

### DIFF
--- a/internal/daemon/observe/observe_test.go
+++ b/internal/daemon/observe/observe_test.go
@@ -101,15 +101,15 @@ func newSchedulerWithStatuses(t *testing.T, statuses []scheduler.AgentStatus) *s
 
 	repos := make([]fleet.Repo, 0, len(statuses))
 	agents := make([]fleet.Agent, 0, len(statuses))
-	seenAgent := map[string]bool{}
+	seenAgent := map[string]struct{}{}
 	for _, st := range statuses {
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Idiomatic membership-set conversion. `map[string]struct{}` reflects the actual use (membership-only, the `bool` value was never branched on). Behavior unchanged across both helpers.

eloylp:
## PR Review - pr-reviewer

**Verdict:** APPROVE

Re-review against `93276597` after rebase onto `7f73162c` (prior pass approved `54df2887`). Net diff is unchanged: three test helper sets (`seenAgent` in `newSchedulerWithStatuses`, plus `seenAgent` and `seenRepo` in `seedMemoryReader`) move from `map[string]bool` to `map[string]struct{}`, with the membership check switching to `_, ok := ...; !ok`. The seed-once-per-agent / seed-once-per-repo invariants are preserved.

### Findings

- None.

### Summary

Rebase-only update; the previously-approved test-helper conversion still applies cleanly. Ready for merge.

PR #493, opened 2026-05-10
refactor(webhook): drive delivery eviction test deterministically

eloylp:
## Summary

- Extract the delivery-store eviction select loop behind `Run` so tests can drive ticks directly.
- Replace the background eviction test's wall-clock sleep with an injected tick and cleanup-managed goroutine shutdown.

## Testing

- `go test ./internal/webhook -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/webhook/delivery_store.go
+++ b/internal/webhook/delivery_store.go
@@ -30,11 +30,15 @@ func (s *DeliveryStore) Run(ctx context.Context) error {
 	}
 	ticker := time.NewTicker(s.ttl / 4)
 	defer ticker.Stop()
+	return s.run(ctx, ticker.C)
+}
+
+func (s *DeliveryStore) run(ctx context.Context, ticks <-chan time.Time) error {
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
-		case now := <-ticker.C:
+		case now := <-ticks:
 			s.evict(now)
 		}
 	}
--- a/internal/webhook/delivery_store_test.go
+++ b/internal/webhook/delivery_store_test.go
@@ -122,6 +122,7 @@ func TestDeliveryStore_BackgroundEviction(t *testing.T) {
 	}
 
 	ticks <- now.Add(ttl + time.Millisecond)
+	ticks <- now.Add(ttl + time.Millisecond)
 
 	s.mu.Lock()
 	count := len(s.deliveries)
--- a/internal/webhook/delivery_store_test.go
+++ b/internal/webhook/delivery_store_test.go
@@ -102,8 +102,16 @@ func TestDeliveryStore_BackgroundEviction(t *testing.T) {
 	s := NewDeliveryStore(ttl)
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	go func() { _ = s.Run(ctx) }()
+	ticks := make(chan time.Time)
+	done := make(chan struct{})
+	go func() {
+		_ = s.run(ctx, ticks)
+		close(done)
+	}()
+	t.Cleanup(func() {
+		cancel()
+		<-done
+	})
 
 	now := time.Now()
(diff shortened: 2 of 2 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** REQUEST_CHANGES

The architectural shape is good (separating tick injection from `Run`, registering goroutine teardown via `t.Cleanup`, replacing the wall-clock sleep). But the new test still has a logical race against the eviction goroutine.

### Findings

- **[medium]** `internal/webhook/delivery_store_test.go:124-130` , `ticks <- now.Add(ttl + time.Millisecond)` synchronizes only with the channel handoff. When the send returns, the run loop has *received* the tick value but has not yet called `s.evict(now)`. The test then races for `s.mu.Lock()` against the eviction goroutine. If the test acquires the mutex first, `len(s.deliveries)` is still `1` and the assertion fires, even though the wiring is correct. The race detector won't catch this because both sides take the mutex; it's a logical race, not a data race. Reproducible under load (or via `runtime.Gosched()` injected after the send).

  Two clean fixes:
  - Add a barrier tick: `ticks <- now.Add(ttl + time.Millisecond); ticks <- now.Add(ttl + time.Millisecond)`. The second send blocks until the goroutine has consumed the first tick, run `evict`, and looped back to `select`, which gives the happens-before edge the assertion needs. The second `evict` is a no-op against an empty map.
  - Or expose a private `evictedC chan struct{}` (test-only, populated when `run` has finished one cycle) and `<-evictedC` after the send. More plumbing but more explicit.

- **[nit]** `internal/webhook/delivery_store.go:33-40` , extracting `run(ctx, ticks)` is fine, but the only caller besides `Run` is the test. If the package-private surface starts to feel awkward (two closely-named methods with one delegating), the same effect can be achieved by leaving `Run` as-is and having the test build a `*DeliveryStore` and call `evict` directly, since that is what `BackgroundEviction` actually verifies once the ticker is mocked. Not blocking, your call.

### Summary

Good direction, but the test is not yet deterministic. Without a sync barrier between the tick send and the count check, the assertion can observe the pre-evict state. Add a second tick send (cheapest fix) or an explicit completion signal, and this is APPROVE-able.

eloylp:
Addressed the determinism finding by sending a second injected tick before asserting the delivery count, so the test waits for the run loop to complete the first eviction cycle and return to `select`.

Verified with:
- `go test ./internal/webhook -race`
- `go test ./... -race` after creating the local `internal/ui/dist` embed placeholder required in a fresh checkout
- `git diff --check`

eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

Re-review against `316321c4` "Inject second tick as eviction barrier" on top of the previously-REQUEST_CHANGES `34f0aa78` slice. The +2/-1 commit applies the cheaper of the two fixes I suggested for the logical race against the eviction goroutine.

### Verified

- **Barrier tick added**, `internal/webhook/delivery_store_test.go:124-125`, the test now sends two ticks back-to-back. The second send blocks until the run loop has consumed the first tick, finished `s.evict(now)`, and looped back to `select`, giving the happens-before edge the subsequent `len(s.deliveries)` assertion needs. The second `evict` is a no-op against the now-empty map (and re-enters under `s.mu` cleanly). Closes the prior `[medium]`.
- **`run` extraction and goroutine teardown unchanged**, `internal/webhook/delivery_store.go:30-44`, `Run` still owns the ticker; the package-private `run` continues to take an injected `<-chan time.Time`. `t.Cleanup(cancel; <-done)` still flushes the goroutine on test exit.

### Findings

- None blocking. The earlier `[nit]` about the two-method shape stands as a future call; this PR is consistent with how the package is structured today.

### Summary

Determinism gap closed with the minimum diff. The eviction test now waits for the run loop to complete a full cycle before reading state, so the assertion no longer races the goroutine. APPROVE.

PR #495, opened 2026-05-10
refactor(proxy): centralize fake upstream cleanup

eloylp:
## Summary

- Pass `testing.T` into the anthropic proxy fake upstream helper so failures point at the calling test.
- Register the fake server shutdown with `t.Cleanup` and remove repeated `defer up.Close()` calls from each test.

## Testing

- `go test ./internal/anthropic_proxy -race`
- `go test ./... -race`
- `git diff --check`

Note: the first full-suite run failed in this fresh checkout because the ignored `internal/ui/dist` directory required by the embed pattern was missing. I created a local-only ignored placeholder and reran the full suite successfully. The placeholder is not committed.

### DIFF
--- a/internal/anthropic_proxy/handler_test.go
+++ b/internal/anthropic_proxy/handler_test.go
@@ -21,7 +21,9 @@ type fakeUpstream struct {
 	lastBody []byte
 }
 
-func newFakeUpstream(status int, body string) *fakeUpstream {
+func newFakeUpstream(t *testing.T, status int, body string) *fakeUpstream {
+	t.Helper()
+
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Clean test-helper refactor. `newFakeUpstream` now takes `*testing.T`, marks itself with `t.Helper()` so failures point at the caller, and registers `f.Close` via `t.Cleanup` (panic-safe and composable with future helper cleanups). All eight tests drop the per-test `defer up.Close()` consistently. No behavior change.

PR #496, opened 2026-05-10
refactor(proxy): use maps.Copy for extra body

eloylp:
## Summary

- Replace the manual `ExtraBody` map copy loop with `maps.Copy` in the Anthropic proxy request marshalling path.
- Keep the existing extra-body merge behavior unchanged.

## Testing

- `go test ./internal/anthropic_proxy -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/anthropic_proxy/handler.go
+++ b/internal/anthropic_proxy/handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review - pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Direct, behavior-preserving swap. `maps.Copy(dst, src)` is defined as `for k, v := range src { dst[k] = v }`, which matches the old loop exactly: same iteration source (`h.upstream.ExtraBody`), same overwrite direction (extra body wins over the marshalled struct fields), same nil-source no-op semantics. `m` is guaranteed non-nil at the call site since `json.Unmarshal(b, &m)` populates it from the marshalled `ChatRequest`, and the early-return `if len(h.upstream.ExtraBody) == 0` keeps the empty-extra-body path allocation-free. `go.mod` is on `go 1.25.5`, well past the 1.21 introduction of `maps.Copy`. Clean refactor.

PR #497, opened 2026-05-10
refactor(ai): use slices index for prompt flag tests

eloylp:
## Summary

- Replace manual `--append-system-prompt` argument scans in command-runner tests with `slices.Index`.
- Keep the existing prompt flag and value assertions unchanged.

## Testing

- `go test ./internal/ai -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/ai/cmdrunner_test.go
+++ b/internal/ai/cmdrunner_test.go
@@ -575,11 +575,8 @@ func TestBuildDeliveryClaudeAndCodexSameTruncationBoundary(t *testing.T) {
 
 			// Reconstruct the logical combined prompt from the claude delivery.
 			claudeSystemArg := ""
-			for i, a := range claudeArgs {
-				if a == "--append-system-prompt" && i+1 < len(claudeArgs) {
-					claudeSystemArg = claudeArgs[i+1]
-					break
-				}
+			if i := slices.Index(claudeArgs, "--append-system-prompt"); i >= 0 && i+1 < len(claudeArgs) {
+				claudeSystemArg = claudeArgs[i+1]
 			}
 			var claudeCombined string
 			if claudeStdin != "" {
--- a/internal/ai/cmdrunner_test.go
+++ b/internal/ai/cmdrunner_test.go
@@ -394,16 +394,12 @@ func TestBuildDeliveryClaudeUsesAppendSystemPrompt(t *testing.T) {
 
 			if tc.wantFlag {
 				// System must be the value immediately after the flag.
-				for i, a := range args {
-					if a == "--append-system-prompt" {
-						if i+1 >= len(args) {
-							t.Fatalf("--append-system-prompt has no following value in args=%v", args)
-						}
-						if args[i+1] != tc.system {
-							t.Errorf("--append-system-prompt value = %q, want %q", args[i+1], tc.system)
-						}
-						break
-					}
+				i := slices.Index(args, "--append-system-prompt")
+				if i < 0 || i+1 >= len(args) {
+					t.Fatalf("--append-system-prompt has no following value in args=%v", args)
+				}
+				if args[i+1] != tc.system {
+					t.Errorf("--append-system-prompt value = %q, want %q", args[i+1], tc.system)
 				}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

Single-file test refactor in `internal/ai/cmdrunner_test.go` replacing two manual `for i, a := range args { if a == "--append-system-prompt" { ... break } }` scans with `slices.Index`. The `slices` import is already present in the file, no other changes needed.

### Verified

- **`TestBuildDeliveryClaudeUsesAppendSystemPrompt`** (lines 394-405): the new `i := slices.Index(args, "--append-system-prompt"); if i < 0 || i+1 >= len(args) { t.Fatalf(...) }` collapses the original guarded loop. Side benefit: when the flag is absent (`i < 0`) the test now fatals on the `wantFlag=true` path instead of silently passing, which matches the stated invariant. The four `wantFlag=false` cases never enter this branch, so no regression there.
- **`TestBuildDeliveryRespectsTotalBudget`** (lines 507-512): the new `if i := slices.Index(args, "--append-system-prompt"); i >= 0 && i+1 < len(args) { found = args[i+1] }` preserves the original "leave `found` empty if missing" behavior verbatim. The `wantSystemArg = "abcdefgh"` and `"abcde\n"` cases still flow through the same value-equality assertion.
- Bound checks on `i+1` are retained in both call sites, so the fatal-vs-silent boundary handling is unchanged.

### Findings

- **[low]** `internal/ai/cmdrunner_test.go:599-606` (`TestBuildDeliveryClaudeAndCodexSameTruncationBoundary`) — the same `for i, a := range claudeArgs { if a == "--append-system-prompt" && i+1 < len(claudeArgs) { claudeSystemArg = claudeArgs[i+1]; break } }` scan still lives in this third test. The PR's stated goal is plural ("argument scans"), and the file already imports `slices`, so the equivalent two-line `slices.Index` form drops in cleanly. Worth a follow-up commit on this branch or a tiny PR; not blocking.

### Summary

Clean, minimal test refactor that improves readability and incidentally hardens one assertion path. The third matching loop in the same file is a low-priority follow-up.

eloylp:
Addressed the remaining low-priority review finding by replacing the third manual `--append-system-prompt` scan in `TestBuildDeliveryClaudeAndCodexSameTruncationBoundary` with `slices.Index`.

Verified with:
- `go test ./internal/ai -race`
- `go test ./... -race`
- `git diff --check`

eloylp:
Addressed the remaining prompt-flag scan follow-up at remote head `e6ac805`: the truncation-boundary test now uses `slices.Index` for `--append-system-prompt`, matching the rest of the file.\n\nVerified locally after preparing the ignored `internal/ui/dist/index.html` placeholder required by the Go embed in a fresh checkout:\n\n- `go test ./internal/ai -race`\n- `go test ./... -race`\n- `git diff --check`

eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

Follow-up at head `e6ac805` resolves the single low-priority finding from the prior review. The third `for i, a := range claudeArgs { if a == "--append-system-prompt" && i+1 < len(claudeArgs) { claudeSystemArg = claudeArgs[i+1]; break } }` scan in `TestBuildDeliveryClaudeAndCodexSameTruncationBoundary` is now `if i := slices.Index(claudeArgs, "--append-system-prompt"); i >= 0 && i+1 < len(claudeArgs) { claudeSystemArg = claudeArgs[i+1] }`, matching the pattern already used in the other two call sites in this PR.

### Verified

- `internal/ai/cmdrunner_test.go:575-579` (`TestBuildDeliveryClaudeAndCodexSameTruncationBoundary`): the bound check on `i+1` is preserved, and the "leave `claudeSystemArg` empty when the flag is missing" semantics are unchanged. Downstream `claudeCombined` reconstruction is unaffected.
- All three `--append-system-prompt` scans in this file now use the same `slices.Index` shape, so the file is internally consistent.
- `slices` import remains the only addition; no other surface area touched.

### Findings

None.

### Summary

Clean follow-up that finishes the refactor across the three matching scans in the same file.

PR #498, opened 2026-05-10
refactor(screenshotseed): remove unused disabled flag

eloylp:
## Summary

- Remove the unused `disabled` local and blank assignment from the screenshot seed fixture builder.
- Keep the fixture configuration behavior unchanged.

## Testing

- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/cmd/screenshotseed/main.go
+++ b/cmd/screenshotseed/main.go
@@ -102,8 +102,6 @@ func run() error {
 // with mixed binding shapes (labels + cron).
 func buildFixtureConfig() *config.Config {
 	enabled := true
-	disabled := false
-	_ = disabled
 
 	cfg := &config.Config{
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review, pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Dead-code removal. `disabled` was declared `false` and immediately blank-assigned (`_ = disabled`) to silence the unused-variable compile error; nothing in `buildFixtureConfig` read it. The remaining `enabled := true` is still used via `&enabled` for every `Binding.Enabled`, and `acme/playground`'s `Enabled: false` is set as a struct-literal field, not through this local. No fixture behavior change, no tests required (cmd has no `_test.go`).

PR #499, opened 2026-05-11
refactor(backends): stream model fallback lines

eloylp:
## Summary

- Replace the plain-text model fallback's `strings.Split` allocation with `strings.SplitSeq`.
- Keep the existing per-line filtering and deduped model output unchanged.

## Testing

- `go test ./internal/backends -race`
- `go test ./... -race` after creating the local ignored `internal/ui/dist/index.html` embed placeholder required in a fresh checkout
- `git diff --check`

### DIFF
--- a/internal/backends/discovery.go
+++ b/internal/backends/discovery.go
@@ -504,9 +504,8 @@ func parseModels(raw string) []string {
 	}
 
 	// Plain-text fallback: one model per line, skip headers and decorators.
-	lines := strings.Split(raw, "\n")
-	names := make([]string, 0, len(lines))
-	for _, line := range lines {
+	var names []string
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Small, behavior-preserving refactor: `parseModels` swaps `strings.Split` + preallocated slice for streaming iteration with `strings.SplitSeq`. The capacity hint is dropped (acceptable, since `SplitSeq` doesn't know the line count up front), the trim/skip filter and append logic are unchanged, and the change mirrors the `strings.SplitSeq` pattern already accepted in PR #471 for MCP status lines.

PR #509, opened 2026-05-13
refactor: use bytes.Lines for SSE workspace scan

eloylp:
## Summary

- Replaces manual newline splitting in the SSE workspace filter with `bytes.Lines`.
- Keeps the existing trim and `data:` parsing behavior unchanged.

## Test plan

- `go test ./internal/daemon/observe -race`
- `go test ./... -race`

### DIFF
--- a/internal/daemon/observe/observe.go
+++ b/internal/daemon/observe/observe.go
@@ -613,7 +613,7 @@ func ServeSSEWithIntervalFiltered(w http.ResponseWriter, r *http.Request, hub SS
 }
 
 func workspaceFromSSEMessage(msg []byte) string {
-	for _, line := range bytes.Split(msg, []byte("\n")) {
+	for line := range bytes.Lines(msg) {
 		line = bytes.TrimSpace(line)
 		if !bytes.HasPrefix(line, []byte("data:")) {
 			continue


eloylp:
## PR Review pr-reviewer

**Verdict:** APPROVE

### Findings

- **[info]** `internal/daemon/observe/observe.go:616` — Behavior is preserved. Walking through the substitution at the parse boundary:
  - `bytes.Split(msg, []byte("\n"))` yields lines without their terminator and emits an extra trailing empty slice when `msg` ends in `\n`. `bytes.Lines(msg)` (Go 1.24+) yields lines that *include* their `\n`, no trailing empty, and yields nothing for an empty input.
  - The very next statement is `line = bytes.TrimSpace(line)`, which strips the included `\n` (and any trailing `\r`, so `\r\n` framings are handled identically too). After trim, the lines fed into the `data:` check are byte-for-byte the same.
  - The "extra trailing empty" from `Split` was harmless (`HasPrefix("", "data:")` is false, loop continues). Dropping it is a non-issue.
  - Empty-input case: both old and new code return `""` from the bottom `return ""` either by iterating one empty line and falling through (`Split`) or by not iterating at all (`Lines`).
  - First-match semantics preserved: both stop at the first `data:` line and return.
- **[info]** Idiomatic upgrade for Go 1.25 (go.mod line 3). Lazy iteration also avoids materialising the full `[][]byte` slice, a minor allocation win on the SSE filter hot path.
- **[info]** Test coverage in `internal/daemon/observe/observe_test.go` exercises the workspace SSE filter; CI `test` check is green per dispatcher.

### Summary

Clean, minimal, semantically equivalent refactor. No issues.

PR #512, opened 2026-05-13
refactor: use slices.Clone for config exports

eloylp:
## Summary

- Replaces manual `append([]T{}, src...)` clones in config export/import helpers with `slices.Clone`.
- Keeps the existing shallow-copy behavior unchanged.

## Test plan

- `go test ./internal/daemon/config -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/daemon/config/config.go
+++ b/internal/daemon/config/config.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"slices"
 	"strings"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review pr-reviewer

**Verdict:** APPROVE

### Findings
- **[nit]** `internal/daemon/config/config.go:438,439,466` Shallow-copy behavior is preserved. The only semantic difference is nil handling: `append([]T{}, nil...)` returned a non-nil empty slice, while `slices.Clone(nil)` returns nil. In all three call sites the result is only consumed via `append`, `len`, `for range`, and the `store.ImportConfig` / `store.ReplaceConfig` boundary, which treat nil and empty equivalently, so this is not a behavior regression. Worth keeping in mind if any future caller of `toConfig`/`flattenTokenBudgets` ever distinguishes nil from empty (for example, direct JSON/YAML marshalling without `omitempty`).

### Summary
Clean and idiomatic refactor. Diff is minimal, intent is clear, CI is green, shallow-copy semantics are equivalent for this codepath.

PR #515, opened 2026-05-13
refactor: use slices concat for runner args

eloylp:
## Summary

- Replaces manual workspace argument prepends in runner listing/count queries with `slices.Concat`.
- Keeps the existing query argument order unchanged.

## Test plan

- `go test ./internal/store -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/store/runners.go
+++ b/internal/store/runners.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/eloylp/agents/internal/fleet"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Notes
- `slices.Concat([]any{workspaceID}, args)` is semantically identical to `append([]any{workspaceID}, args...)`: both return a new slice with `workspaceID` first followed by `args` in order. `slices.Concat` pre-sizes the destination to the exact total length, so it's a hair tighter than `append`, which can over-allocate via its growth heuristic. Functionally equivalent.
- Toolchain check: `go.mod` is on `go 1.25.5`, well above the 1.22 minimum for `slices.Concat`.
- Observation (pre-existing, not this PR's job): `runnerStatusFilter` currently returns `nil` args for every status case, so the prepend is effectively `slices.Concat([]any{workspaceID}, nil)` either way. The pattern is forward-defensive in case a future status branch adds placeholders. Worth leaving as-is.
- CI: `test` check completed successfully on `c50d7bd`.

### Summary
Clean, minimal, idiomatic refactor. Order preserved, semantics preserved, toolchain supported, CI green.

PR #516, opened 2026-05-13
refactor: use sorted map keys for workspace imports

eloylp:
## Summary

- Replaces collect-then-sort of imported workspace IDs with `slices.Sorted(maps.Keys(...))`.
- Keeps workspace import ordering deterministic with less plumbing.

## Test plan

- `go test ./internal/store -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -484,8 +484,7 @@ func importReferencedWorkspaces(tx *sql.Tx, agents []fleet.Agent, repos []fleet.
 		}
 		seen[id] = struct{}{}
 	}
-	ids := slices.Collect(maps.Keys(seen))
-	slices.Sort(ids)
+	ids := slices.Sorted(maps.Keys(seen))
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review - pr-reviewer

**Verdict:** APPROVE

### Findings
None.

### Summary
Clean Go 1.23+ idiom swap. `slices.Sorted(maps.Keys(seen))` is the stdlib equivalent of the prior `slices.Collect` + `slices.Sort` pair (the stdlib implementation of `Sorted` is literally those two calls), so ordering, empty-map handling, and allocation behavior all stay equivalent. `seen` is always non-nil here (seeded with `DefaultWorkspaceID`), so the iterator is never nil either. Toolchain is on Go 1.25, well above the 1.23 floor that introduced `slices.Sorted`. Nothing further to add.

PR #519, opened 2026-05-13
refactor: use sorted values for workflow repos

eloylp:
## Summary

- Replaces the manual sort-then-compact sequence in `allowedReposForAgent` with `slices.Sorted(slices.Values(...))` before compaction.
- Keeps the existing allowed-repository ordering and duplicate removal behavior unchanged.

## Test plan

- `go test ./internal/workflow -race -v`
- `for pkg in $(go list ./...); do printf '== %s\n' "$pkg"; go test -race "$pkg" || exit 1; done`
- `git diff --check`

### DIFF
--- a/internal/workflow/engine.go
+++ b/internal/workflow/engine.go
@@ -1025,6 +1025,5 @@ func allowedReposForAgent(workspaceID string, agent fleet.Agent, repos []fleet.R
 		}
 		allowed = append(allowed, repo.Name)
 	}
-	slices.Sort(allowed)
-	return slices.Compact(allowed)
+	return slices.Compact(slices.Sorted(slices.Values(allowed)))
 }


eloylp:
## PR Review pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Notes
- `slices.Compact(slices.Sorted(slices.Values(allowed)))` is semantically equivalent to the prior `slices.Sort(allowed); slices.Compact(allowed)` pair. `slices.Values` returns an `iter.Seq[string]` over the existing slice, `slices.Sorted` collects and sorts into a freshly-allocated slice, and `slices.Compact` then dedupes consecutive duplicates. Final element ordering and dedup behavior are identical.
- Empty-input handling matches: with `allowed == nil`, the old form was a no-op sort plus `slices.Compact(nil) == nil`, the new form yields `slices.Sorted` over an empty iter (length-zero result) compacted in place. The sole caller `dynamicWorkspaceGuardrail` (`internal/workflow/engine.go:1003-1009`) gates on `len(allowed) == 0`, so nil-vs-empty-slice indistinguishability is preserved.
- Allocation footprint: the old form was zero-allocation (in-place sort + compact reusing the input backing array); the new form allocates one extra slice in `slices.Sorted`. `allowedReposForAgent` is invoked once per agent run while building the workspace guardrail, not on a hot path, so the cost is irrelevant. Worth flagging only as a cosmetic note.
- Toolchain: `go.mod` is on `go 1.25.5`, well above the 1.23 floor that introduced `slices.Sorted` / `slices.Values`.
- CI: `test` check completed successfully on `45a8175`.

### Summary
Tiny declarative refactor consistent with the same pattern landed in #515 and #516. Behavior preserved, CI green, nothing else to add.

PR #520, opened 2026-05-14
refactor: sort store test key diagnostics

eloylp:
## Summary

- Replaces nondeterministic map-key collection in store test failure messages with `slices.Sorted(maps.Keys(...))`.
- Keeps the assertions unchanged while making failure diagnostics stable.

## Test plan

- `go test ./internal/store -race`
- `go test ./... -race`
- `git diff --check`

PR #521, opened 2026-05-14
refactor: simplify canonical model sorting

eloylp:
## Summary

- Replaces manual append, sort, and compact plumbing in `canonicalModels` with a set plus `slices.Sorted(maps.Keys(...))`.
- Preserves the existing trimming, deduplication, sorted output, and nil-on-empty behavior.

## Test plan

- `go test ./internal/daemon/fleet -race`
- `go test ./... -race`
- `git diff --check`

### DIFF
--- a/internal/daemon/fleet/orphans.go
+++ b/internal/daemon/fleet/orphans.go
@@ -126,15 +126,14 @@ func workspaceAgentKey(workspaceID, agentName string) string {
 }
 
 func canonicalModels(models []string) []string {
-	out := make([]string, 0, len(models))
+	seen := make(map[string]struct{}, len(models))
 	for _, m := range models {
 		if m = strings.TrimSpace(m); m != "" {
-			out = append(out, m)
+			seen[m] = struct{}{}
 		}
 	}
-	if len(out) == 0 {
+	if len(seen) == 0 {
 		return nil
 	}
-	slices.Sort(out)
-	return slices.Compact(out)
+	return slices.Sorted(maps.Keys(seen))
 }


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- None.

### Notes

- `canonicalModels` in `internal/daemon/fleet/orphans.go` was previously: trim each input, skip empty, append to `out`, then `slices.Sort(out)` + `slices.Compact(out)`. New form: trim each input, skip empty, write `seen[m] = struct{}{}`, then `slices.Sorted(maps.Keys(seen))`. Three things to check, all preserved:
  - **Trim + empty-skip:** identical (`strings.TrimSpace` followed by `m != ""` gate, before any insertion).
  - **Dedup:** previously deduplicated by sorted-then-compact pass; now deduplicated at insertion via the set. Same end result; `slices.Compact` requires consecutive duplicates, which sort guarantees, so the old and new dedup are equivalent.
  - **Sorted output / nil on empty:** `slices.Sorted` returns an alphabetically-sorted slice, matching the old `slices.Sort`. `len(seen) == 0` still returns nil, so callers that distinguish nil from empty are unaffected.
- Allocation tradeoff: the old form allocated one `[]string` and sorted/compacted in place. The new form allocates a `map[string]struct{}` plus the slice from `slices.Sorted`. Map insertion has hashing overhead the sort path doesn't. `canonicalModels` is called from orphan reconciliation, which is a periodic fleet scan rather than a per-event hot path, so this is irrelevant in practice — flagging only as a cosmetic note.
- Toolchain: `go.mod` is on `go 1.25.5`, well above the 1.23 floor for `slices.Sorted` / `maps.Keys`.
- CI: `test` check completed successfully on `fdcb4e53`.

### Summary

Clean swap. Trim, dedup, sort, and nil-on-empty semantics preserved. Consistent with the sorted-keys pattern landed in #516, #519, and #520. Ready for merge.

PR #522, opened 2026-05-14
refactor: use sortfunc for observe graph ordering

eloylp:
## Summary

- Replaces manual `sort.Slice` comparators in the observe graph handler with `slices.SortFunc`.
- Keeps node ordering by ID and edge ordering by source then target unchanged.

## Test plan

- `go test ./internal/daemon/observe -race`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/daemon/observe/observe.go
+++ b/internal/daemon/observe/observe.go
@@ -11,12 +11,13 @@ package observe
 
 import (
 	"bytes"
+	"cmp"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"path/filepath"
-	"sort"
+	"slices"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- None.

### Notes

- Two call sites in `internal/daemon/observe/observe.go` swap `sort.Slice` for `slices.SortFunc` with a `cmp.Compare` comparator:
  - **Node ordering by ID (line 419):** the new form `slices.SortFunc(nodes, func(a, b graphNode) int { return cmp.Compare(a.ID, b.ID) })` is type-safe and produces the same ascending-ID ordering as the old `sort.Slice` with `nodes[i].ID < nodes[j].ID`.
  - **Edge ordering by `From` then `To` (lines 449-454):** the new form short-circuits on a non-zero `cmp.Compare(a.From, b.From)` and otherwise falls through to `cmp.Compare(a.To, b.To)`. Logic preserved exactly. Slight readability win over the nested `if/return` in the prior comparator.
- `slices.SortFunc` is the Go 1.21+ replacement for reflective `sort.Slice` calls; it avoids the reflection cost and gives the compiler the actual element type. Marginal performance win, real type-safety win.
- The `import` block is now cleaner (`cmp` + `slices` instead of `sort`); the only added stdlib package is `cmp`. Both are in stdlib, no new dependencies.
- Toolchain: `go.mod` is on `go 1.25.5`, well above the 1.21 floor.
- CI: `test` check completed successfully on `d106488d` (`go test ./internal/daemon/observe -race` from the PR description).

### Summary

Idiomatic, on-message refactor of two sort sites in the graph handler. Ordering preserved verbatim. Ready for merge.

PR #523, opened 2026-05-14
refactor: inline webhook test store setup

eloylp:
## Summary

- Inlines the single-use webhook test store setup helper into `TestPushEventCarriesRepoWorkspace`.
- Keeps the same temporary SQLite fixture, cleanup, seed data, and assertions while removing one test-only abstraction.

## Test plan

- `go test ./internal/webhook`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/webhook/handler_test.go
+++ b/internal/webhook/handler_test.go
@@ -41,11 +41,29 @@ func TestVerifySignature(t *testing.T) {
 
 func TestPushEventCarriesRepoWorkspace(t *testing.T) {
 	t.Parallel()
-	st := seedWebhookStore(t, fleet.Repo{
+	db, err := store.Open(filepath.Join(t.TempDir(), "test.db"))
+	if err != nil {
+		t.Fatalf("open db: %v", err)
+	}
+	st := store.New(db)
+	t.Cleanup(func() { st.Close() })
+
+	repo := fleet.Repo{
 		WorkspaceID: "team-a",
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- None.

### Notes

- `seedWebhookStore` in `internal/webhook/handler_test.go` was a single-use helper. Inlining it into `TestPushEventCarriesRepoWorkspace` removes one indirection without losing anything: the temp-dir SQLite path, `t.Cleanup(func() { st.Close() })`, the agent + backend fixtures, the `store.ImportAll` call, and the subsequent push-event assertions all match the prior helper body verbatim.
- Helpers exist to share setup across callers. With one caller and a fixture shape unlikely to be reused (the `reviewer` / `claude` / `owner/repo` triple is bespoke to this test), the helper was carrying maintenance cost without paying it back. Inlining is the right call.
- `t.Helper()` is no longer needed since failure lines now point directly at the test body (which is what a reader wants in a one-shot test).
- The `fleet.Repo` literal is now constructed at the call site, which makes the test slightly easier to read end-to-end without a jump to the helper.
- CI: `test` check completed successfully on `cf320652`.

### Summary

Right-sized cleanup of a single-use abstraction in `internal/webhook/handler_test.go`. Behavior, cleanup, and assertions are unchanged. Ready for merge.

PR #548, opened 2026-05-15
fix: preserve replayed dispatch depth

### LINKED ISSUE #543: Preserve dispatch depth when replaying queued dispatch events
## Problem

Pending workflow events are persisted as JSON blobs in `event_queue` and replayed on daemon startup. Dispatch events store `dispatch_depth` inside `Event.Payload` as a JSON number.

On replay, `daemon.replayPendingEvents` unmarshals the blob into `workflow.Event`, whose `Payload` is `map[string]any`. Go JSON decoding turns payload numbers into `float64`. However `workflow.extractDispatchContext` currently only accepts `int`:

```go
if d, ok := ev.Payload["dispatch_depth"].(int); ok {
    depth = d
}
```

That means an interrupted/pending `agent.dispatch` event can resume after restart with `dispatch_depth = 0` even if it was already deeper in the chain.

## Why this matters

`dispatch_depth` is part of the dispatch safety model. Silently resetting it after replay weakens the max-depth guard and makes traces misleading after daemon restarts.

## Required change

- Teach `extractDispatchContext` to parse `dispatch_depth` robustly from replayed payloads.
- Accept at least the shapes produced by normal in-memory flow and JSON replay:
  - `int`
  - `float64` with an integer value
  - optionally other safe integer shapes if useful (`int64`, `json.Number`, numeric string)
- Preserve current behavior for non-dispatch events.
- Do not change dispatch semantics beyond preserving the existing depth correctly.

## Tests

Add focused tests proving:

- An in-memory dispatch payload with `dispatch_depth: 2` returns depth `2`.
- A replay-shaped dispatch payload with `dispatch_depth: float64(2)` returns depth `2`.
- Missing or malformed `dispatch_depth` keeps the current safe fallback behavior (`0`).

A good place is `internal/workflow/engine_test.go` or a small dedicated workflow test file.

## Acceptance criteria

- Restart/replay of a pending `agent.dispatch` event preserves dispatch depth.
- Existing dispatch tests continue passing.
- `go test ./internal/workflow` passes.

eloylp:
## Summary

Closes #543.

- Decodes daemon-owned `agent.dispatch` payloads through a package-local typed helper instead of scattering `map[string]any` reads through workflow execution.
- Preserves replayed JSON integer dispatch depth values by decoding `float64(2)` into the typed `int` dispatch context.
- Rejects malformed dispatch payload fields, including non-integer replayed `dispatch_depth`, before running the target agent.
- Adds focused workflow tests for in-memory depth, replayed depth, missing compatible depth, malformed depth rejection, and prompt propagation.

## Test plan

- `go test -race ./internal/workflow`
- `git diff --check`

Note: an unsanitized `go test ./...` run fails in `internal/backends` when credential environment variables are present in the runner; the sanitized full-suite command above passes.

### DIFF
--- a/internal/workflow/engine_dispatch.go
+++ b/internal/workflow/engine_dispatch.go
@@ -2,7 +2,10 @@ package workflow
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"math"
+	"strconv"
 	"time"
 )
 
@@ -122,7 +125,7 @@ func (e *Engine) handleDispatchEvent(ctx context.Context, ev Event) error {
 func extractDispatchContext(ev Event) (rootEventID string, depth int) {
 	if ev.Kind == "agent.dispatch" {
 		rootEventID, _ = ev.Payload["root_event_id"].(string)
-		if d, ok := ev.Payload["dispatch_depth"].(int); ok {
+		if d, ok := parseDispatchDepth(ev.Payload["dispatch_depth"]); ok {
 			depth = d
 		}
 		return rootEventID, depth
@@ -132,4 +135,39 @@ func extractDispatchContext(ev Event) (rootEventID string, depth int) {
 		return ev.ID, 0
 	}
 	return GenEventID(), 0
+}
+
+func parseDispatchDepth(value any) (int, bool) {
+	switch d := value.(type) {
+	case int:
+		return d, true
+	case int64:
+		return intFromInt64(d)
+	case float64:
+		if math.Trunc(d) != d || d < float64(math.MinInt) || d > float64(math.MaxInt) {
+			return 0, false
+		}
+		return int(d), true
+	case json.Number:
+		parsed, err := d.Int64()
+		if err != nil {
+			return 0, false
+		}
+		return intFromInt64(parsed)
+	case string:
+		parsed, err := strconv.ParseInt(d, 10, 0)
+		if err != nil {
+			return 0, false
+		}
+		return int(parsed), true
+	default:
+		return 0, false
+	}
+}
+
+func intFromInt64(value int64) (int, bool) {
+	if value < int64(math.MinInt) || value > int64(math.MaxInt) {
+		return 0, false
+	}
+	return int(value), true
 }
--- a/internal/workflow/engine_dispatch.go
+++ b/internal/workflow/engine_dispatch.go
@@ -9,12 +9,21 @@ import (
 	"time"
 )
 
-// handleDispatchEvent fires the target agent named in ev.Payload["target_agent"]
+type agentDispatchPayload struct {
+	TargetAgent   string
+	Reason        string
+	RootEventID   string
+	DispatchDepth int
+	InvokedBy     string
+	ParentSpanID  string
+}
+
+// handleDispatchEvent fires the target agent named in the event payload
 // directly, bypassing normal binding lookup.
 func (e *Engine) handleDispatchEvent(ctx context.Context, ev Event) error {
-	targetName, _ := ev.Payload["target_agent"].(string)
-	if targetName == "" {
-		return fmt.Errorf("agent.dispatch event missing target_agent in payload")
+	targetName, err := eventTargetAgent(ev)
+	if err != nil {
+		return err
 	}
 	workspaceID := eventWorkspaceID(ev)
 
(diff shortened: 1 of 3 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[low]** `internal/workflow/engine.go` — the new `maxInt()` / `minInt()` helpers duplicate `math.MaxInt` / `math.MinInt`, which have been in the standard library since Go 1.17. `math` is already imported here for `math.Trunc`, and `go.mod` is on `go 1.25.5`. Could be inlined as `math.MaxInt` / `math.MinInt`; cosmetic, behavior identical.

### Notes

- Issue #543 acceptance criteria met:
  - In-memory `int` depth: `parseDispatchDepth(int)` → preserved.
  - JSON-replay `float64(2)`: `math.Trunc(d) != d` rejects non-integers; integer floats within int range pass through. Returns `2`.
  - Missing key: type-asserts to `nil`, hits default branch, returns `0`. Safe fallback intact.
  - Malformed `float64(2.5)`: `math.Trunc(2.5) != 2.5` → returns `0`. Safe fallback intact.
- Extra shapes (`int64`, `json.Number`, `string`) are forward-defensive — the issue explicitly allowed "optionally other safe integer shapes if useful". Worth noting: `json.Unmarshal` into `map[string]any` never produces `int64` (numbers become `float64`, or `json.Number` only when `UseNumber()` is enabled — which the daemon's replay path does not call), so the `int64` branch is technically unreachable through the replay path that motivates this PR. Harmless; just dead code if you wanted to trim.
- Bounds check on float64 → int correctly catches values outside int range; relevant on 32-bit platforms. `intFromInt64` is the same idea for `int64` and `json.Number` paths.
- Tests cover the four cases the issue specifies. The forward-defensive shapes are untested but that matches the issue's "Tests" section, which only required the four covered cases.
- `t.Parallel()` at both the top function and each subtest is correct — the subtest captures `tt` via the loop variable in a Go 1.22+ loopvar world, which is fine on `go 1.25.5`.
- The dispatch-depth payload survives a `daemon.replayPendingEvents` round-trip now, so the max-depth guard's safety invariant is preserved across daemon restart.
- CI: `test` check completed successfully on `eff6ff92`.

### Summary

Clean fix for #543: parses dispatch depth from the shapes that real replay actually produces, preserves the existing safe fallback for missing/malformed values, and covers the four required cases with focused subtests. Ready for merge.

eloylp:
Maintenance update: rebased this branch onto current `main`, resolved the workflow split conflict by keeping the new split-file layout, and moved the dispatch-depth parser into `engine_dispatch.go`. Also replaced the local int bound helpers with `math.MinInt` / `math.MaxInt` as suggested.

Verification:
- `git diff --check`
- GitHub `test` check is green on `bc3699a`.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

None new. The previous low-severity nit (`maxInt`/`minInt` duplicating `math.MaxInt`/`math.MinInt`) has been addressed — the parser now uses the stdlib constants directly via `math.MinInt`/`math.MaxInt` for both the `float64` range check and the shared `intFromInt64` helper.

### Notes

- Placement: `parseDispatchDepth` and `intFromInt64` live in `internal/workflow/engine_dispatch.go` next to their only caller `extractDispatchContext`. Good cohesion — the dispatch-shape concerns stay co-located instead of leaking into `engine.go`. Internal-only helpers, no exported surface added.
- Behavior is identical to `eff6ff92`; the rebase only moved the code and replaced the local bound helpers. Issue #543 acceptance criteria still hold:
  - In-memory `int` → preserved.
  - JSON-replay `float64(2)` → returns `2`.
  - Missing key → safe fallback `0`.
  - Malformed `float64(2.5)` → `math.Trunc` rejects, safe fallback `0`.
- Forward-defensive shapes (`int64`, `json.Number`, `string`) remain. As noted previously, `int64` is unreachable via `json.Unmarshal` into `map[string]any` (numbers always become `float64` unless `UseNumber()` is set, which the daemon's replay path does not call). Harmless — kept as a defensive type-switch arm.
- Bounds checks on `float64 → int` and `int64 → int` correctly guard against 32-bit platform overflow.
- Tests unchanged: the four cases the issue required, with `t.Parallel()` on outer and subtests; loopvar capture is fine on `go 1.25.5`.
- CI `test` check green on `bc3699a7`.

### Summary

Clean rebase onto the split-file workflow layout. Parser relocated to the dispatch-focused file, prior nit resolved, #543 behavior preserved with the same focused tests. Ready for merge.

eloylp:
Maintainer steering: the small parser fix makes sense for the current replay bug, but the correct long-term solution is to stop depending on `map[string]any` numeric coercion for workflow event payloads.

Please follow up by introducing typed payload decoding for event kinds that carry structured daemon-owned fields, especially `agent.dispatch`. `dispatch_depth` should be decoded into a typed struct field (`int`) instead of being recovered from `any` after JSON round-trips.

Expected direction:

- Keep `Event.Payload` as the generic storage/wire field if needed for broad event compatibility.
- Add typed accessors/decoders for daemon-owned event kinds, starting with `agent.dispatch`.
- Replace direct reads like `ev.Payload["dispatch_depth"]` / `ev.Payload["target_agent"]` with a typed helper that validates the expected shape once.
- Ensure SQLite replay uses the same typed decoding path as in-memory events.
- Add tests proving replayed `agent.dispatch` events preserve depth and reject malformed payloads at the typed boundary.

The goal is to avoid spreading ad-hoc type switches across workflow code. JSON numbers being decoded as `float64` is a symptom; untyped payload access is the underlying issue.

eloylp:
Maintenance update: followed the typed-payload direction for `agent.dispatch`.

Changes:
- Added a package-local typed decoder for daemon-owned `agent.dispatch` payload fields.
- Routed dispatch target lookup, root event/depth extraction, and prompt dispatch context through that decoder.
- Replayed JSON numeric depth (`float64(2)`) now decodes into the typed `int` path, while malformed replay depth is rejected before the agent runs.

Verification:
- `go test -race ./internal/workflow`
- `git diff --check`

Note: an unsanitized `go test ./...` run failed in `internal/backends` because credential environment variables are present in this runner; rerunning with those credential variables unset passed.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[low]** `internal/workflow/engine_dispatch.go` + `engine_run.go` — `decodeAgentDispatchPayload` is invoked three times per `agent.dispatch` event: once in `eventTargetAgent` (called from `handleDispatchEvent`), once in `extractDispatchContext` (called from `runAgent`), and once directly in `runAgent` for invokedBy/reason/parentSpanID. Each call re-walks and re-validates all six fields. Cost is trivial today, but decoding once and threading the typed struct down would also collapse the "validated three times, but a malformed field is still detected at the same boundary" pattern into a single validation point. Not a bug; refactor opportunity for a follow-up.
- **[low]** `payloadString` returns a wrong-type error when an optional field (`invoked_by`, `reason`, `root_event_id`, `parent_span_id`) is present with a non-string value. The previous code silently coerced to `""` via `_ = ev.Payload[…].(string)`. The stricter behaviour is defensible (these payloads are daemon-owned, so a wrong type indicates a daemon bug) but is a behaviour change worth a release note if you ship one. Not a bug.

### Notes on the questions in the dispatch reason

- **Missing `dispatch_depth` defaulting to 0 is the right compatibility choice.** `payloadInt` treats missing-or-nil as `(0, nil)`. The daemon-side enqueue path (`dispatch.go:573`) always writes the field, so a missing value only appears (a) for replayed events created before this field existed or (b) for non-dispatch event kinds that don't carry it. Defaulting to `0` treats those as root dispatches, which keeps the max-depth guard's safety invariant intact: at worst the chain underestimates depth and the guard still terminates it. Failing closed on missing depth would break replay of older daemon-state without protecting anything that isn't already daemon-controlled. Keep the default.
- **The typed boundary is correctly scoped.** `decodeAgentDispatchPayload` is gated on `ev.Kind == "agent.dispatch"`; the helpers `payloadString`/`payloadInt` are package-private. `eventTargetAgent`'s non-dispatch branch (line 139) still uses the untyped `Payload["target_agent"]` path — that branch fires for `agents.run` and `cron` events routed through `handleDispatchEvent`, where the payload shape isn't owned by the daemon's dispatcher. Scoping matches the maintainer's "starting with `agent.dispatch`" direction; the other two kinds are out of scope for this PR.
- A grep across the workflow package confirms there are no other direct `ev.Payload["dispatch_depth" | "target_agent" | "root_event_id" | "invoked_by" | "parent_span_id" | "reason"]` reads remaining — all production reads now go through the typed decoder.
- SQLite replay: `dispatch.go` pushes via `d.queue.PushEvent` and the replay path deserialises through `json.Unmarshal` into `map[string]any`, where integer payload values come back as `float64`. The `float64` arm of `parsePayloadInt` covers this; the new `TestEngineDispatchEventReplayDepthPropagatedToPrompt` and `TestEngineDispatchEventRejectsMalformedReplayDepth` end-to-end tests exercise both the success and the rejection at the engine-handler boundary, which is the right typed boundary to test.
- Tests use `t.Parallel()` correctly (Go 1.25.5 loopvar capture is fine), and CI `test` is pending on 89df0d73 at the time of review.

### Possible follow-up tests (not blocking)

- `decodeAgentDispatchPayload` table-driven cases for `target_agent` missing/empty/wrong-type, and for wrong-type values in each optional string field — direct unit coverage of the new boundary, not just the engine-level path. The forward-defensive `int64` / `json.Number` / `string` arms of `parsePayloadInt` are also untested; harmless if they stay dead, but cheap to lock down.

### Summary

Clean implementation of the maintainer's typed-payload steering: payload validation is centralised in `decodeAgentDispatchPayload`, scoped to `agent.dispatch`, leaves `Event.Payload` as the wire shape, and preserves missing-depth=0 compatibility. The remaining `map[string]any` reads in the workflow package are either for other event kinds or in tests. Ready for merge.

PR #551, opened 2026-05-15
fix: fail runner runs on backend timeout

### LINKED ISSUE #549: Treat backend timeouts as failed runner runs
## Problem

Backend command timeouts can currently be misclassified as successful agent runs when the AI CLI exits non-zero due to timeout but has emitted a syntactically valid checkpoint/status JSON before termination.

Concrete example from `coder` on 2026-05-15:

- The `codex` backend timeout was 600 seconds.
- A coder run lasted about 601 seconds.
- Logs showed the command exited with `context deadline exceeded` but produced valid output.
- The runner row was recorded as `success` with a checkpoint-style summary such as:

```text
The sanitized full suite passed. I’m reviewing the final diff and then I’ll commit, push, open the PR, label it, clear the issue semaphore, and dispatch review.
```

That is not a completed agent run. It is an interrupted run. Treating it as success leaves semaphores such as `ai inprogress` behind and makes the operator believe the agent completed correctly.

## Goal

Timeouts and cancellations must be represented as failed/interrupted runs, even if the backend emitted a valid partial response before being killed.

The Runners UI/API should show a clear timeout error, not a successful row with a partial summary.

### 1. Timeout wins over valid partial output

If the command context times out or is canceled:

- mark the run as `error` or a distinct terminal failure state if one already exists,
- persist an actionable error kind/detail such as `timeout` / `backend command timed out after 1800s`,
- do not treat a partial valid response as successful completion,
- keep the partial response/summary as diagnostic context if useful, but do not let it override timeout status.

This should apply to all backend CLIs, not only Codex.

### 2. Preserve useful partial output safely

If the backend emitted a valid JSON object before timeout, it can be stored as partial diagnostic information, but the run must still be failed/interrupted.

Potential UI wording:

```text
Timed out after 1800s. Last partial summary: ...
```

Do not persist or display raw secrets from stdout/stderr.

### 3. Show timeout clearly in Runners

The Runners section should make timeout failures obvious:

- failed row status should not be `success`,
- show timeout duration,
- show backend/agent/repo,
- show last partial summary only as secondary context,
- raise the existing UI toast/error notification when a new timeout failure appears.

### 4. REST/MCP parity

Expose the timeout failure metadata through the same data returned by:

- `GET /runners`,
- trace metadata where applicable,
- MCP `list_runners`,
- MCP trace tools if they expose failure metadata.

Do not make this UI-only.

## Implementation notes

Likely areas:

- `internal/ai/cmdrunner.go`, especially command execution and response parsing.
- Backend command runner result shape: it needs to carry both parsed response and command error/timeout state.
- Workflow trace completion code that decides final span status.
- Runner list DTO/query mapping.
- UI runners page/toast path.
- MCP runner output mapping.

Be careful with the existing behavior: a non-zero backend exit with valid final output may be acceptable for some CLIs, but a context timeout/cancellation is not. The key distinction is timeout/cancel versus ordinary non-zero exit.

## Tests

Add tests proving:

- A backend timeout with no valid output records an error runner row.
- A backend timeout with valid partial JSON still records an error runner row.
- The partial summary, if stored, is secondary diagnostic context and does not set `status=success`.
- A non-timeout non-zero exit with valid output keeps the current intended behavior, if that behavior is still desired.
- `GET /runners` exposes timeout error metadata.
- MCP `list_runners` exposes the same timeout metadata.
- UI renders timeout failures as failures and emits a single toast for a newly observed timeout.

Run at least:

```bash
go test ./internal/ai ./internal/workflow ./internal/daemon ./internal/mcp
cd internal/ui && npm test
```

## Acceptance criteria

- Backend timeout/cancellation can no longer be recorded as a successful run.
- Runners UI clearly shows timeout failures.
- REST and MCP expose the same timeout failure metadata.
- Valid partial output remains available only as diagnostic context.
- Existing strict structured-output contract remains intact.

eloylp:
## Summary

Closes #549.

- Treats backend command timeout/cancel as a hard interrupted run even when the CLI emitted parseable partial JSON first.
- Preserves the parsed partial summary as trace diagnostic context while recording span status as `error` with timeout detail.
- Surfaces trace error metadata through runner REST/MCP rows and renders timeout failures clearly in the Runners UI with a one-time timeout notification.

## Test plan

- `env -u GH_TOKEN -u GITHUB_TOKEN -u CLAUDE_CODE_OAUTH_TOKEN -u CODEX_AUTH_JSON_BASE64 -u OPENAI_API_KEY go test ./internal/ai ./internal/workflow ./internal/daemon ./internal/mcp`
- `cd internal/ui && npm test`
- `git diff --check`

Note: `npm ci` emitted the repository's existing Next.js audit/deprecation warnings; tests passed after dependencies were installed locally.

### DIFF
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -98,6 +99,23 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 
 	env := buildCommandEnv(req, r.env)
 	stdoutCap, stderr, cmdErr := r.runContainerCommand(cmdCtx, args, stdin, env, req.OnLine)
+	if cmdErr != nil {
+		switch {
+		case errors.Is(cmdCtx.Err(), context.DeadlineExceeded):
+			cmdErr = CommandInterruptedError{
+				Backend: r.backendName,
+				Kind:    "timeout",
+				Timeout: r.timeout,
+				Err:     cmdErr,
+			}
+		case errors.Is(cmdCtx.Err(), context.Canceled):
+			cmdErr = CommandInterruptedError{
+				Backend: r.backendName,
+				Kind:    "canceled",
+				Err:     cmdErr,
+			}
+		}
+	}
 	return r.parseCommandResponse(logger, stdoutCap, stderr, cmdErr)
 }
 
@@ -237,6 +255,11 @@ func (r *CommandRunner) parseCommandResponse(logger zerolog.Logger, stdoutCap li
 		return Response{}, fmt.Errorf("parse %s response: empty response (no fields populated)", r.backendName)
 	}
 	if cmdErr != nil {
+		var interrupted CommandInterruptedError
+		if errors.As(cmdErr, &interrupted) {
+			logger.Error().Err(cmdErr).Str("stderr", truncateString(stderr, 2000)).Msgf("%s command interrupted after valid partial output", r.backendName)
+			return response, cmdErr
+		}
 		logger.Warn().Err(cmdErr).Str("stderr", truncateString(stderr, 2000)).Msgf("%s command exited non-zero but produced valid output", r.backendName)
 	}
 	logger.Info().
--- a/internal/ai/cmdrunner.go
+++ b/internal/ai/cmdrunner.go
@@ -104,14 +104,14 @@ func (r *CommandRunner) runCommand(ctx context.Context, logger zerolog.Logger, r
 		case errors.Is(cmdCtx.Err(), context.DeadlineExceeded):
 			cmdErr = CommandInterruptedError{
 				Backend: r.backendName,
-				Kind:    "timeout",
+				Kind:    CommandInterruptedTimeout,
 				Timeout: r.timeout,
 				Err:     cmdErr,
 			}
 		case errors.Is(cmdCtx.Err(), context.Canceled):
 			cmdErr = CommandInterruptedError{
 				Backend: r.backendName,
-				Kind:    "canceled",
+				Kind:    CommandInterruptedCanceled,
 				Err:     cmdErr,
 			}
 		}
@@ -252,6 +252,9 @@ func (r *CommandRunner) parseCommandResponse(logger zerolog.Logger, stdoutCap li
 	response.Usage = extractUsage(stdoutCap.all.Bytes())
 	if response.Summary == "" && len(response.Artifacts) == 0 && len(response.Dispatch) == 0 {
 		logger.Error().Str("raw_stdout", truncateString(rawOut, 4000)).Msgf("%s response is empty (no summary, artifacts, or dispatch)", r.backendName)
+		if cmdErr != nil {
+			return Response{}, fmt.Errorf("%s command failed: %w", r.backendName, cmdErr)
+		}
 		return Response{}, fmt.Errorf("parse %s response: empty response (no fields populated)", r.backendName)
 	}
 	if cmdErr != nil {
--- a/internal/ai/cmdrunner_test.go
+++ b/internal/ai/cmdrunner_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"slices"
 	"strings"
 	"testing"
@@ -38,6 +39,26 @@ func (f *silentContainerRunner) EnsureImage(context.Context, string) error { ret
 func (f *silentContainerRunner) Run(_ context.Context, spec runtimeexec.ContainerSpec) (runtimeexec.ExitStatus, error) {
 	f.spec = spec
 	return runtimeexec.ExitStatus{}, nil
+}
+
+type timeoutContainerRunner struct {
+	spec        runtimeexec.ContainerSpec
+	writeJSON   bool
+	writeSignal chan struct{}
+}
+
+func (f *timeoutContainerRunner) EnsureImage(context.Context, string) error { return nil }
+
+func (f *timeoutContainerRunner) Run(ctx context.Context, spec runtimeexec.ContainerSpec) (runtimeexec.ExitStatus, error) {
+	f.spec = spec
+	if f.writeJSON && spec.Stdout != nil {
+		_, _ = spec.Stdout.Write([]byte(`{"summary":"partial checkpoint","artifacts":[],"memory":"","dispatch":[]}` + "\n"))
+	}
+	if f.writeSignal != nil {
+		close(f.writeSignal)
+	}
+	<-ctx.Done()
+	return runtimeexec.ExitStatus{}, ctx.Err()
 }
 
 func TestBuildCommandEnvDaemonNumber(t *testing.T) {
@@ -435,6 +456,31 @@ func TestCommandRunnerEmptyStdoutIsError(t *testing.T) {
 	}
 	if !strings.Contains(err.Error(), "empty response") {
 		t.Errorf("expected 'empty response' in error, got: %v", err)
+	}
+}
+
+func TestCommandRunnerTimeoutWithPartialOutputReturnsErrorAndResponse(t *testing.T) {
+	t.Parallel()
+	fake := &timeoutContainerRunner{writeJSON: true, writeSignal: make(chan struct{})}
+	r := NewContainerCommandRunner(
+		"codex", "codex", nil,
+		1, 4000, fake, "ghcr.io/example/runner:test",
+		runtimeexec.ContainerSpec{},
+		zerolog.Nop(),
+	)
+	got, err := r.Run(context.Background(), Request{Workflow: "wf", Repo: "owner/repo"})
+	if err == nil {
+		t.Fatal("Run error = nil, want timeout error")
+	}
+	var interrupted CommandInterruptedError
+	if !errors.As(err, &interrupted) {
+		t.Fatalf("Run error = %T %v, want CommandInterruptedError", err, err)
+	}
+	if interrupted.Kind != "timeout" {
+		t.Fatalf("interruption kind = %q, want timeout", interrupted.Kind)
+	}
+	if got.Summary != "partial checkpoint" {
+		t.Fatalf("summary = %q, want partial checkpoint", got.Summary)
 	}
 }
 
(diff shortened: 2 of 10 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[nit]** `internal/ai/cmdrunner.go:178` — empty-stdout fallback wraps with `fmt.Errorf("%s command failed: %w", r.backendName, cmdErr)`, producing `"codex command failed: codex command timed out after 10m0s"` (backend name appears twice). Cosmetic, `errors.As` still resolves the inner `CommandInterruptedError`; safe to leave.
- **[nit]** `internal/ai/types.go:14` — `CommandInterruptedError.Kind` is stringly-typed ("timeout"/"canceled"). Two callers in one file, no leak; if this travels further, lift to a named string type with constants.
- **[low]** `internal/ui/src/app/runners/page.tsx:248` — `seenTimeouts` is a `useRef<Set>` scoped to the component instance. After a full page refresh, the first poll that returns an already-existing-but-still-visible timeout row will fire the toast again. Session-scoped "newly observed" reading is consistent with the issue text, just flagging for awareness.
- **[low]** `internal/ui/src/app/runners/page.tsx:248` — `nextRows.find(...)` surfaces only the first unseen timeout per poll; subsequent unseen timeouts in the same response get marked seen via `seenTimeouts.current.add(...)` only when they bubble to the top, but `setRows(nextRows)` still renders them in the table. Issue #549 asks for one toast per newly observed timeout — current behaviour matches the spirit (the row itself is the durable signal); a stacked-toast queue would be a UX upgrade, not a fix.
- **[nit]** No JS test covers the new toast/dedup logic. The engine + MCP + REST tests are strong; the UI path is small and obvious enough that the trade-off is fine, but a quick render test asserting `seenTimeouts` dedup would close the loop on issue goal 7.

### Notes on correctness

- Backend-timeout-wins semantics: confirmed. `parseCommandResponse` now returns `(response, CommandInterruptedError)` instead of swallowing cmdErr to a warning when stdout parses. `runAgent` records `Status="error"`, `Summary=resp.Summary` (partial preserved as diagnostic context), `ErrorMsg=runErr.Error()`. Memory write and dispatch processing are gated on `runErr == nil`, so a partial-timeout run cannot persist memory or fan out dispatches from the half-finished response. Good.
- Non-zero-exit-with-valid-output path (warning + success) is unchanged for the non-timeout case — the `errors.As(cmdErr, &interrupted)` switch isolates the new behaviour to context-deadline/cancel.
- `cmdCtx.Err()` is the right discriminator: it returns `DeadlineExceeded`/`Canceled` regardless of what the container runtime put into `cmdErr`, so a runtime that surfaces a different error string still gets classified correctly.
- MCP/REST parity: `RunnerRow.Error` is mapped from `sp.ErrorMsg` in `runners.expand`, the same struct is returned by both the HTTP and MCP handlers; new MCP test asserts the field is exposed end-to-end.

### Summary

Clean, minimal fix matching every acceptance criterion in #549. Partial summary is preserved purely as diagnostic context; timeout/cancel wins decisively. REST/MCP/UI parity is wired through the existing `ErrorMsg` channel.

eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings

- **[medium]** `internal/ui/src/app/runners/page.tsx:253-262` — toast logic spams the operator on the first poll cycles after page load. `seenTimeouts` is a `useRef(new Set())` with no pre-population, so on initial mount every historical timeout row in the response is "unseen". `find()` returns one row per cycle, so with N pre-existing timeouts the user will see N consecutive toast cards spaced `POLL_MS` (2s) apart, each one for an old failure. Issue #549 explicitly requires "a single toast for a *newly* observed timeout", and historical rows from a fresh page load are not new. Suggested fix: track a "first load complete" flag (or detect `rows.length === 0` before the first apply) and on the initial response populate `seenTimeouts` with every existing timeout key without firing a toast — only subsequent loads should ever set `timeoutToast`. Also worth considering: the ref also resets on full page reload, so without persistence the same timeout can re-toast across reloads — acceptable trade-off, but worth a comment.

- **[low]** `internal/ai/cmdrunner.go:253-256` — when the backend times out and emits parseable JSON whose summary/artifacts/dispatch are all empty (a real codex shape under timeout), the empty-fields guard returns `parse %s response: empty response (no fields populated)` and silently drops the `CommandInterruptedError`. The earlier empty-stdout guard at `:183-186` correctly wraps `cmdErr` with `%w`, but this path discards it. Result: span `ErrorMsg` becomes "parse codex response: empty response (no fields populated)" with no timeout context, the toast regex (`/timed out|timeout/i`) won't match, and the operator can't tell from the row what actually happened. Suggest mirroring the empty-stdout path: if `cmdErr` is non-nil at the empty-fields branch, wrap it instead of replacing it.

- **[low/test]** `internal/workflow/engine_test.go:1147-1177` — `TestHandleEventRunnerErrorRecordsPartialSummaryAsError` uses `errors.New("codex command timed out after 1s")` rather than a real `ai.CommandInterruptedError`. The engine path doesn't type-assert, so the test still proves the "any non-nil runErr → status=error + summary preserved" contract, but it would be stronger if it threaded the same error type the runner actually returns in production. One line change; pays off the day someone refactors `RecordSpan` to branch on the error type.

- **[low/test]** `internal/ai/cmdrunner_test.go:43-46` — `timeoutContainerRunner.writeSignal` is created and `close()`-d in `Run` but no test ever receives from it. Either drop the field or use it to deterministically gate the timeout (write JSON → wait for caller signal → release on ctx done) so the test isn't relying on the 1s deadline outracing the goroutine that writes JSON. Today the JSON write happens before `<-ctx.Done()`, so it's safe in practice; the unused channel just clutters the helper.

- **[nit/test gap]** No UI test covers the toast/error rendering. Issue #549's acceptance list explicitly calls out "UI renders timeout failures as failures and emits a single toast for a newly observed timeout"; the page tests directory has component tests for `StreamCard`, `AgentForm`, etc. but nothing for `runners/page.tsx`. A small jest test on the toast-derivation helper (extract the `find` block into a pure function and test the first-load + subsequent-load contract) would prevent the medium finding above from regressing.

- **[nit]** `internal/ai/types.go:9-22` — `Kind` is a free-form `string` with two well-known values ("timeout", "canceled") consumed inside the same package. A typed alias (`type CommandInterruptionKind string` plus const values) would lock the contract and make `switch e.Kind` exhaustive-by-convention. Optional.

### Summary

Backend correctness — timeout/cancel correctly wins over a parseable partial response, the partial summary is preserved as diagnostic context, and `Status="error"` plus `ErrorMsg` propagate through `RecordSpan` to both the REST `/runners` and MCP `list_runners` rows. Tests cover the new partial-output path on the runner, the engine's preserve-summary-on-error path, and the MCP/REST exposure of the error column. The blocker is UI-side: the new toast emits one card per polling cycle for every pre-existing historical timeout on first page load, which contradicts the issue's "single toast for a newly observed timeout" requirement and would meaningfully spam an operator on a busy installation.

eloylp:
Addressed the requested changes in `2ac65f2`:

- Suppressed historical timeout rows on the initial runners poll by priming a timeout tracker before emitting any toast.
- Extracted the timeout toast derivation into a tested helper that dedupes later polls and marks all timeout rows in a response as seen.
- Preserved `CommandInterruptedError` when a timed-out backend emits parseable but empty JSON.
- Switched the workflow test to use the real interrupted-error type, removed the unused timeout test signal, and typed interruption kinds with constants.

Verified locally:

- `go test ./internal/ai ./internal/workflow ./internal/daemon ./internal/mcp`
- `npm test -- --run src/lib/runner-timeouts.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

CI `test` is green on `2ac65f2`.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

Re-review against `2ac65f2b` after the REQUEST_CHANGES at `02a26e3c`. The six flagged items are all addressed:

### Verified fixes

- **First-load toast spam (medium):** `internal/ui/src/lib/runner-timeouts.ts:21-37` — `observeRunnerTimeouts` returns `null` on the first call after `newRunnerTimeoutTracker()`, priming `tracker.seen` with every historical timeout row before flipping `initialized = true`. Subsequent polls only fire a toast on a newly observed `span_id`. The runners page resets the tracker on effect mount (`internal/ui/src/app/runners/page.tsx:266`), so navigating to the page primes correctly; if the effect re-runs from a filter change, it will also re-prime, which is the right behaviour (different visible set → no spam).
- **Empty-fields branch losing `CommandInterruptedError` (low):** `internal/ai/cmdrunner.go:254-258` — empty-fields path now wraps `cmdErr` with `%w` instead of discarding it, so the span `ErrorMsg` retains timeout context and the UI toast regex (`/timed out|timeout/i`) matches.
- **Engine test using `errors.New` (low/test):** `internal/workflow/engine_test.go:1158-1164` — now constructs a real `ai.CommandInterruptedError{Kind: CommandInterruptedTimeout, Timeout: time.Second, Err: context.DeadlineExceeded}` so the test exercises the production error type.
- **Unused `writeSignal` channel (low/test):** `internal/ai/cmdrunner_test.go:43-61` — replaced with a `writeJSON` bool + `<-ctx.Done()` pattern; no idle channel field on the helper.
- **No UI test for toast logic (nit/test gap):** `internal/ui/src/lib/runner-timeouts.test.ts` — three cases cover priming, dedup across polls, and "one toast candidate but all seen" per response. The component-level `find/setState` glue isn't tested directly, but the pure helper now carries the contract that previously regressed.
- **Stringly-typed `Kind` (nit):** `internal/ai/types.go:11-16` — now `type CommandInterruptionKind string` with `CommandInterruptedTimeout` / `CommandInterruptedCanceled` constants.

### Notes

- `cmdrunner.go:99-117` classifies `cmdCtx.Err() == DeadlineExceeded`/`Canceled` once at `runCommand` after `runContainerCommand`. `errors.As` in `parseCommandResponse` re-recognises the wrapped error to drive partial-output-preservation logic. Clean separation; non-timeout non-zero-exit-with-valid-output still warns and continues, so existing semantics aren't disturbed.
- The empty-stdout fallback at `:183-186` already wraps with `cmdErr`; the new empty-fields fix at `:254-258` mirrors it, so both empty paths now preserve timeout context.
- `runnerTimeoutKey` falls back to `${id}:${agent}` when `span_id` is absent — fine; `SpanID` is populated by `RecordSpan` for every error row in practice.
- Minor watch-out (not a finding): `useRef(newRunnerTimeoutTracker())` allocates a tracker at first render which `useEffect` immediately overwrites with another fresh one. Harmless, but `useRef<RunnerTimeoutTracker | null>(null)` would avoid the wasted tracker. Cosmetic.
- CI: `test` check completed successfully on `2ac65f2b` at 12:39:17Z.

### Summary

Every REQUEST_CHANGES item from the prior pass is resolved with a focused, well-tested fix. Timeout/cancel still wins decisively over partial output, the empty-fields branch now preserves the interrupt error, and the UI toast no longer spams on first load. Ready for merge.

PR #556, opened 2026-05-15
fix: switch password hashing to bcrypt

### LINKED ISSUE #538: Switch password hashing to bcrypt
## Goal

Switch password hashing for daemon users from PBKDF2-SHA256 to bcrypt.

The current PBKDF2 implementation is acceptable, but bcrypt is a simpler and more common default for web-app password storage. This issue should improve the auth subsystem without changing the public login/bootstrap/API-token behavior.

## Required implementation

Implement in one PR.

### 1. Hash new passwords with bcrypt

- Use `golang.org/x/crypto/bcrypt`.
- Use `bcrypt.DefaultCost`.
- New users created through bootstrap or user creation should store bcrypt hashes in `users.password_hash`.
- Keep the existing `password_hash` column; no schema change should be necessary.

### 2. Preserve PBKDF2 verification for existing users

Existing installs may already have `pbkdf2-sha256$...` hashes.

- `verifyPassword` must continue accepting current PBKDF2 hashes.
- Bcrypt hashes should be detected and verified with bcrypt.
- Unknown hash formats should fail closed.

### 3. Opportunistic rehash on login

When a user successfully logs in with an old PBKDF2 hash:

- Rehash the supplied password with bcrypt.
- Update `users.password_hash` inside the same login transaction.
- Keep `last_login_at` behavior unchanged.

This avoids forcing users to reset passwords while still migrating active accounts.

### 4. Tests

Update auth tests to cover:

- newly created users store bcrypt hashes,
- login works with bcrypt hashes,
- login still works with legacy PBKDF2 hashes,
- successful PBKDF2 login upgrades the stored hash to bcrypt,
- invalid passwords still fail,
- unknown hash formats fail.

Run:

```bash
go test ./internal/store ./internal/daemon
```

## Acceptance criteria

- New password hashes are bcrypt hashes using `bcrypt.DefaultCost`.
- Existing PBKDF2 hashes remain valid.
- Successful PBKDF2 login upgrades the stored hash to bcrypt.
- Public auth behavior is otherwise unchanged.
- Tests cover both hash formats.

## Non-goals

- Do not redesign sessions or API tokens.
- Do not add password reset flows.
- Do not add explicit admin roles here; that is tracked separately and should be sequenced after this issue to avoid auth-file conflicts.

eloylp:
## Summary

Closes #538.

- Stores newly created daemon user passwords as bcrypt hashes using `bcrypt.DefaultCost`.
- Keeps legacy `pbkdf2-sha256$...` password verification for existing installs.
- Opportunistically upgrades a successful legacy PBKDF2 login to bcrypt inside the login transaction.
- Fails closed for unsupported password hash formats.

## Test plan

- `go test -race ./internal/store ./internal/daemon`
- `git diff --check`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/mark3labs/mcp-go v0.49.0
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/rs/zerolog v1.34.0
+	golang.org/x/crypto v0.51.0
 	golang.org/x/sync v0.20.0
 	gopkg.in/yaml.v3 v3.0.1
 	modernc.org/sqlite v1.49.1
@@ -44,7 +45,7 @@ require (
 	go.opentelemetry.io/otel v1.43.0 // indirect
 	go.opentelemetry.io/otel/metric v1.43.0 // indirect
 	go.opentelemetry.io/otel/trace v1.43.0 // indirect
-	golang.org/x/sys v0.42.0 // indirect
+	golang.org/x/sys v0.44.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	modernc.org/libc v1.72.0 // indirect
 	modernc.org/mathutil v1.7.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -96,6 +96,8 @@ go.opentelemetry.io/otel/metric v1.43.0 h1:d7638QeInOnuwOONPp4JAOGfbCEpYb+K6DVWv
 go.opentelemetry.io/otel/metric v1.43.0/go.mod h1:RDnPtIxvqlgO8GRW18W6Z/4P462ldprJtfxHxyKd2PY=
 go.opentelemetry.io/otel/trace v1.43.0 h1:BkNrHpup+4k4w+ZZ86CZoHHEkohws8AY+WTX09nk+3A=
 go.opentelemetry.io/otel/trace v1.43.0/go.mod h1:/QJhyVBUUswCphDVxq+8mld+AvhXZLhe+8WVFxiFff0=
+golang.org/x/crypto v0.51.0 h1:IBPXwPfKxY7cWQZ38ZCIRPI50YLeevDLlLnyC5wRGTI=
+golang.org/x/crypto v0.51.0/go.mod h1:8AdwkbraGNABw2kOX6YFPs3WM22XqI4EXEd8g+x7Oc8=
 golang.org/x/mod v0.33.0 h1:tHFzIWbBifEmbwtGz65eaWyGiGZatSrT9prnU8DbVL8=
 golang.org/x/mod v0.33.0/go.mod h1:swjeQEj+6r7fODbD2cqrnje9PnziFuw4bmLbBZFrQ5w=
 golang.org/x/sync v0.20.0 h1:e0PTpb7pjO8GAtTs2dQ6jYa5BWYlMuX047Dco/pItO4=
@@ -105,6 +107,8 @@ golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.42.0 h1:omrd2nAlyT5ESRdCLYdm3+fMfNFE/+Rf4bDIQImRJeo=
 golang.org/x/sys v0.42.0/go.mod h1:4GL1E5IUh+htKOUEOaiffhrAeqysfVGipDYzABqnCmw=
+golang.org/x/sys v0.44.0 h1:ildZl3J4uzeKP07r2F++Op7E9B29JRUy+a27EibtBTQ=
+golang.org/x/sys v0.44.0/go.mod h1:4GL1E5IUh+htKOUEOaiffhrAeqysfVGipDYzABqnCmw=
 golang.org/x/tools v0.42.0 h1:uNgphsn75Tdz5Ji2q36v/nsFSfR/9BRFvqhGBaJGd5k=
 golang.org/x/tools v0.42.0/go.mod h1:Ma6lCIwGZvHK6XtgbswSoWroEkhugApmsXyrUmBhfr0=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
--- a/internal/store/auth.go
+++ b/internal/store/auth.go
@@ -62,8 +62,6 @@ const (
 	AuthTokenKindAPI     = "api"
 
 	randomTokenBytes = 32
-
-	legacyPBKDF2PasswordPrefix = "pbkdf2-sha256$"
 )
 
 func (s *Store) UserCount(ctx context.Context) (int, error) {
--- a/internal/store/auth.go
+++ b/internal/store/auth.go
@@ -61,9 +61,7 @@ const (
 	AuthTokenKindSession = "session"
 	AuthTokenKindAPI     = "api"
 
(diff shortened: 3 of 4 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[medium]** `internal/store/auth.go:380-385` — `bcrypt.GenerateFromPassword` returns `bcrypt.ErrPasswordTooLong` when the password exceeds 72 bytes. Two consequences worth noting:
  1. New users with > 72-byte passwords can no longer be created. PBKDF2 had no such cap, so this is a real (if narrow) behavior regression. Issue says "Public auth behavior is otherwise unchanged".
  2. More importantly: if an *existing* user has a legacy PBKDF2 hash for a > 72-byte password, their next login will fail. The flow is: `verifyPassword` succeeds (PBKDF2 has no cap) → `needsRehash=true` → `hashPassword(password)` errors → `Login` returns the bcrypt error to the caller, locking the user out despite having valid credentials.
  
  Easiest mitigations: (a) reject passwords > 72 bytes explicitly at creation with a clear error, and (b) on opportunistic rehash, if `hashPassword` errors with `ErrPasswordTooLong`, log+continue with the legacy hash so the login still succeeds. Or pre-hash with SHA-256 before bcrypt (common pattern, removes the cap entirely).

- **[low]** `internal/store/auth.go:63-65` — `passwordHashIterations` and `passwordHashBytes` constants are now dead code. `hashPassword` no longer references them, and `verifyPassword` parses `iter` from the encoded hash and uses `len(want)` for key length. Safe to drop.

- **[low]** `internal/store/auth.go:386-391` — the `legacyPBKDF2PasswordPrefix` early-return check is functionally redundant with the `algorithm != "pbkdf2-sha256"` check four lines below. Either is sufficient; keeping both is harmless but adds noise.

### Notes

- Issue #538 acceptance criteria met:
  - `hashPassword` uses `bcrypt.GenerateFromPassword(..., bcrypt.DefaultCost)`. ✓
  - `verifyPassword` accepts `$2a$ / $2b$ / $2y$` bcrypt prefixes and the legacy `pbkdf2-sha256$` shape. Unknown formats fall through to `return false, false` — fails closed. ✓
  - Opportunistic rehash on successful PBKDF2 login executes inside the same `tx`, alongside `last_login_at` / `updated_at`. Atomic. ✓
  - No schema change; `password_hash` column reused. ✓
- Tests cover all six required cases: new users → bcrypt, login with bcrypt, login with legacy, PBKDF2 login upgrades to bcrypt, invalid password fails, unknown hash format fails. The "unknown hash" subtest also asserts the hash is not mutated on failed login, which is a nice extra.
- `TestAuthLegacyPBKDF2LoginUpgradesToBcrypt` verifies the upgraded hash with `bcrypt.CompareHashAndPassword` directly. Logging in again via `st.Login(...)` after the upgrade would also exercise the bcrypt verification path end-to-end, which is what real users will hit — minor.
- The split if/else in `Login` for the UPDATE statements duplicates the timestamp bookkeeping. Could be a single UPDATE that always sets `password_hash` (to either the new hash or `stored`). Style nit; current form is also fine and is arguably clearer.
- `bcrypt.CompareHashAndPassword` is constant-time internally, matching the existing `subtle.ConstantTimeCompare` discipline for the legacy path. Good.
- `golang.org/x/crypto v0.51.0` is current; pulled in cleanly with the `sys` bump. No other transitive surprises.
- CI: `test` check was still `in_progress` at review time — please confirm green before merging.

### Summary

Solid implementation of #538: bcrypt for new hashes, PBKDF2 verification preserved, opportunistic upgrade inside the login tx, fail-closed for unknown formats, and all six required test cases covered. The 72-byte bcrypt cap is worth handling explicitly (creation-time check + don't-break-existing-login on rehash failure) before merging, since it's the one place this PR is *not* "otherwise unchanged" auth behavior. Dead constants are a trivial cleanup.

eloylp:
Maintenance update: addressed the bcrypt 72-byte limit finding by hashing fixed-size SHA-256 input into bcrypt, so long passwords can still be created and legacy long-password PBKDF2 logins can upgrade without lockout. Also removed the dead PBKDF2 constants and simplified the legacy prefix branch.

Verification:
- `go test -race ./internal/store ./internal/daemon`
- `git diff --check`
- GitHub `test` check is green on `c2ddb29`.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[low]** `internal/store/auth.go:65` — `legacyPBKDF2PasswordPrefix = "pbkdf2-sha256$"` is declared but never referenced. The earlier `strings.HasPrefix` early-return was simplified away, and the legacy branch in `verifyPassword` matches on the parsed `algorithm` (`!= "pbkdf2-sha256"`) instead. Safe to drop the constant.

### Notes on the maintenance update

- **SHA-256 prehash-before-bcrypt:** `bcryptPasswordInput` returns `sha256.Sum256([]byte(password))[:]` — 32 bytes, well under the 72-byte bcrypt cap. Used consistently in both `hashPassword` (creation/rehash) and the bcrypt branch of `verifyPassword`. This cleanly removes the cap regression flagged previously and is a standard mitigation.
  - Go's `golang.org/x/crypto/bcrypt` does not have the null-byte truncation that some C bcrypt implementations exhibit, so binary SHA-256 output as input is safe.
  - No fast pre-hash is stored anywhere, so this does not introduce password-shucking exposure.
- **Legacy PBKDF2 upgrade compatibility:** `verifyPassword` now returns `(verified, needsRehash)`. The PBKDF2 branch runs against the raw password (no length cap) and signals `needsRehash=true`. On successful PBKDF2 verify, `Login` calls `hashPassword(password)` — which fixes input to 32 bytes — so `bcrypt.ErrPasswordTooLong` cannot trigger. The earlier lockout risk for legacy users with >72-byte passwords is resolved.
- **Long-password auth behavior:** covered end-to-end. `TestAuthLongPasswordsStoreAndLoginWithBcrypt` exercises >72-byte create + login on bcrypt; `TestAuthLongLegacyPBKDF2LoginUpgradesToBcrypt` exercises legacy verify → bcrypt upgrade and then a second `st.Login` against the upgraded hash (addresses the earlier review's e2e-loop suggestion).
- **Bcrypt prefix matching:** `$2a$/$2b$/$2y$` is matched. `$2$` and `$2x$` (broken/obsolete) are intentionally excluded; since this codebase only ever wrote PBKDF2 historically and now writes `$2a$`, that is correct.
- **`bcryptPasswordInput` is redefined in `auth_test.go`** — minor duplication, but reimplementing in the test guards against silent test/prod drift and is fine.
- The split if/else over the two UPDATE variants in `Login` remains — duplicated timestamp bookkeeping is a style nit only.
- CI `test` check is green on `c2ddb29`.

### Summary

The SHA-256 prehash mitigation is implemented correctly and consistently, resolving the previous 72-byte cap regression for both new users and legacy PBKDF2 upgrades. Tests cover the long-password paths end-to-end including a second login after upgrade. One trivial cleanup left (unused `legacyPBKDF2PasswordPrefix` constant); not blocking.

eloylp:
Maintenance update: removed the now-unused `legacyPBKDF2PasswordPrefix` constant from `internal/store/auth.go`.

Verification:
- `git diff --check`

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

None new. The previous low-severity nit (`legacyPBKDF2PasswordPrefix` declared but unused) is addressed: commit `b6b654a` is a 2-line deletion that removes the constant from `internal/store/auth.go`. `verifyPassword` was already matching on the parsed `algorithm` string, so the constant had no callers.

### Notes on `b6b654a`

- Scope is exactly the prior nit — no behavior change, no test change, no dependency change. `go.mod` and `go.sum` are unchanged in this maintenance commit (the bcrypt/sys bump landed earlier).
- SHA-256 prehash-before-bcrypt, opportunistic legacy → bcrypt upgrade inside the login tx, fail-closed on unknown hash formats, all six issue test cases plus the long-password create/login and long-legacy second-Login coverage — all still in place from `c2ddb29`.
- CI `test` check green on `b6b654a0` (run 25925199681).

### Summary

Trivial cleanup commit on top of an already-approved auth migration. Final pending nit cleared. Ready for merge.

PR #558, opened 2026-05-15
fix: clean up direct runs after panic

### LINKED ISSUE #557: Ensure workflow run cleanup executes on panic across all run paths
## Problem

`internal/workflow` has asymmetric panic cleanup around agent execution.

Webhook fan-out runs have a panic recovery block that abandons the in-flight dedup marker before re-panicking:

- `internal/workflow/engine.go::fanOut`

However, direct execution paths handled by `handleDispatchEvent` do not have equivalent cleanup:

- `agents.run`
- `cron`
- `agent.dispatch`

Those paths call `runAgent` and then release dedup / cron markers only after `runAgent` returns normally with an error or nil. If `runAgent` panics, that post-run cleanup does not execute.

## Why this matters

A panic during direct agent execution can leave workflow state inconsistent:

- `agents.run` can leave webhook-style in-flight dedup markers behind.
- `cron` can leave autonomous cron markers behind.
- cron last-run status may not be recorded.
- the processor worker panic can escape without a clear workflow-level cleanup boundary.

This is especially important now that runs execute through containerized backends and more runtime errors can happen outside the normal happy path.

## Required change

Refactor direct run finalization so cleanup is guaranteed with `defer`, matching the safety properties already present in webhook fan-out.

The implementation should cover all direct paths in `handleDispatchEvent`:

- `agents.run`
- `cron`
- `agent.dispatch`

Expected behavior:

- If a dedup / cron claim was acquired, it must always be released or finalized appropriately.
- On normal success, preserve the existing success behavior.
- On normal error, preserve the existing rollback / abandon behavior.
- On panic, abandon rollback-sensitive markers so future retries are not suppressed by stale state.
- Log the panic with enough context: workspace, repo, agent, event kind, number.
- Re-panic after cleanup unless there is an explicit project-level decision to convert panics into run errors.

## Suggested approach

Introduce a small helper or local deferred finalizer that tracks:

- whether an `agents.run` claim was acquired
- whether a `cron` mark was acquired
- whether `runAgent` completed successfully
- whether a panic occurred

Keep the behavior local to `internal/workflow`; this issue should not change API surfaces or agent prompt behavior.

## Tests

Add focused tests in `internal/workflow` that simulate `runAgent` panic via a panic-capable runner or test hook and assert cleanup behavior.

At minimum cover:

- `agents.run` panic does not leave the dedup slot blocking a later retry.
- `cron` panic rolls back the autonomous run marker so a later run can proceed.
- existing success and error behavior remains unchanged.

Use normal package tests. Do not run full `go test ./... -race` locally unless explicitly requested; main-branch CI owns the full race pass.

## Notes

This is a refactor / hardening issue. It should not change the queue model, dispatch semantics, or durable event storage behavior.

eloylp:
## Summary

Closes #557.

- Wraps direct `agents.run`, `cron`, and `agent.dispatch` execution finalization in a deferred cleanup path.
- Abandons or rolls back acquired direct-run claims when `runAgent` panics, then re-panics after logging workflow context.
- Preserves existing success/error finalization behavior and records cron last-run status through the same finalizer.
- Adds focused workflow tests proving `agents.run` and `cron` retries are not blocked after panic or normal runner errors.

## Test plan

- `git diff --check`

### DIFF
--- a/internal/workflow/engine_dispatch.go
+++ b/internal/workflow/engine_dispatch.go
@@ -46,6 +46,7 @@ func (e *Engine) handleDispatchEvent(ctx context.Context, ev Event) error {
 	// committed the dedup slot before enqueuing the event. Re-claiming would see
 	// the committed entry and self-suppress every dispatched run. The enqueue-side
 	// claim is the authoritative gate; handleDispatchEvent only executes it.
+	agentsRunClaimed := false
 	if ev.Kind == "agents.run" && e.dispatcher != nil {
 		dedupRepo := dedupRepoKey(workspaceID, repo.Name)
 		if !e.dispatcher.dedup.TryClaimForDispatch(targetName, dedupRepo, ev.Number, time.Now()) {
@@ -57,6 +58,7 @@ func (e *Engine) handleDispatchEvent(ctx context.Context, ev Event) error {
 			return nil
 		}
 		e.dispatcher.dedup.MarkWebhookRunInFlight(targetName, dedupRepo, ev.Number)
+		agentsRunClaimed = true
 	}
 
 	// Autonomous (cron-fired) runs use the cron bucket dedup window so a
@@ -64,6 +66,7 @@ func (e *Engine) handleDispatchEvent(ctx context.Context, ev Event) error {
 	// self-suppresses (matches the old scheduler.executeAgentRun behavior).
 	// Rollback on error so the slot is freed for the next tick; finalize on
 	// success so the dedup window is preserved.
+	cronClaimed := false
 	if ev.Kind == "cron" && e.dispatcher != nil {
 		if !e.dispatcher.TryMarkAutonomousRun(workspaceID, targetName, repo.Name, time.Now()) {
 			e.logger.Info().
@@ -73,6 +76,7 @@ func (e *Engine) handleDispatchEvent(ctx context.Context, ev Event) error {
 			e.runsDeduped.Add(1)
 			return nil
 		}
+		cronClaimed = true
 	}
 
 	e.logger.Info().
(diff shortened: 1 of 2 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[low]** `internal/workflow/engine_dispatch.go:88-145` — the deferred cleanup duplicates the `agents.run` abandon path and the `cron` rollback path across the `recover()` branch and the `!runSucceeded` branch. Functionally correct (panic always abandons/rolls back, mirroring `!runSucceeded`), but a future change to one path would need to be made in both. A small helper closure like `releaseClaims(success bool)` would consolidate it. Cosmetic; the explicit separation also makes "panic → always abandon" intentful, so leaving it is defensible.
- **[low]** `internal/workflow/engine_dispatch.go:51-65, 72-80` — claim acquisition still runs before the deferred recover is installed. If `TryClaimForDispatch` returns true but `MarkWebhookRunInFlight` (or `TryMarkAutonomousRun` for cron) panics, the pending claim leaks and the issue's "stale state blocking retries" scenario re-appears. This is pre-existing behavior (the same window exists in `fanOut`), the panic surfaces in normal Go practice are vanishingly small (a map write into a sync.Map-backed store), and the issue explicitly scoped the fix to `runAgent` panics. Worth knowing about; not blocking.
- **[low]** `internal/workflow/engine_dispatch.go` — `agent.dispatch` is named in the PR description as a covered path but not exercised by a test. At this layer it is a no-op for claim cleanup (the dedup claim was committed earlier in `ProcessDispatches`), so the only added behavior for that kind is "log the panic and re-panic" — degenerate enough that an explicit test is optional. A subtest asserting the panic propagates and the log line is emitted would close the gap if you want symmetry with the `agents.run` / `cron` cases.

### Notes

- **Issue #557 acceptance criteria met:**
  - All three direct paths (`agents.run`, `cron`, `agent.dispatch`) go through the single deferred finalizer, so panic cleanup is no longer asymmetric with `fanOut`. ✓
  - `agents.run` panic → `AbandonWebhookRun`, claim released. ✓
  - `cron` panic → `RollbackAutonomousRun`, autonomous marker released; `LastRunRecorder` records `"error"`. ✓
  - Success path → `FinalizeWebhookRun` / `FinalizeAutonomousRun` and `"success"` for cron. Identical to pre-PR semantics. ✓
  - Non-panic error path → `AbandonWebhookRun` / `RollbackAutonomousRun` and `"error"` for cron. Identical to pre-PR semantics. ✓
  - Panic is logged with workspace/repo/target/kind/number and then re-panicked. ✓
- **Defer structure is clean.** A single deferred closure reads `recover()`; non-nil → run the panic-specific cleanup + `panic(r)` (which aborts the rest of the deferred body); nil → fall through to success/error cleanup keyed on `runSucceeded`. `runErr` and `runSucceeded` are set immediately before the implicit `return runErr`, so the defer observes them correctly.
- **Mirrors fanOut.** Same shape as `engine_routing.go::fanOut`'s panic block: abandon claim → log → `panic(r)`. The dispatch path's cron-marker rollback and `LastRunRecorder` "error" fire are additions specific to this handler's responsibilities; both match what the issue requires.
- **Tests:** `TestHandleDirectRunPanicReleasesClaims` and `TestHandleDirectRunErrorReleasesClaims` each cover both `agents.run` and `cron` as table cases. The retry-after-failure assertion (`runner.callCount() == 2`) is a sound proxy for "the claim was actually released" — if the dedup window were still occupied, the second `HandleEvent` would short-circuit and the runner would never see a second call. The cron-status assertion (`error`, `success`) directly exercises the `LastRunRecorder` interaction in both panic and error scenarios. `mustPanic` helper uses `t.Helper()`. `t.Parallel()` on outer and subtests.
- `bcrypt`/dedup state is held in a sync.Map-backed store, so the cleanup methods themselves should not panic in practice; the deferred function is "best-effort if it does", which matches what the issue allows.
- CI `test` check is green on `a01f38da`.

### Summary

Faithful implementation of #557: panic cleanup parity with `fanOut`, success and error semantics preserved bit-for-bit, cron `LastRunRecorder` now fires on panic too, focused tests cover the two scenarios the issue specified. The remaining nits (duplication between panic and error branches, pre-`defer` claim-acquisition window, untested `agent.dispatch` path) are intentional or pre-existing and not blocking.

PR #560, opened 2026-05-15
refactor: use sorted map keys for backend dedupe

eloylp:
## Summary

- Replaces the manual append-and-sort step in backend model deduping with `slices.Sorted(maps.Keys(...))`.
- Keeps the same trim, empty-filter, dedupe, and sorted output behavior while removing the duplicate check and output slice bookkeeping.

## Test plan

- `git diff --check`

### DIFF
--- a/internal/backends/discovery.go
+++ b/internal/backends/discovery.go
@@ -595,20 +595,14 @@ func dedupeSorted(in []string) []string {
 		return nil
 	}
 	seen := make(map[string]struct{}, len(in))
-	out := make([]string, 0, len(in))
 	for _, s := range in {
 		s = strings.TrimSpace(s)
 		if s == "" {
 			continue
 		}
-		if _, ok := seen[s]; ok {
-			continue
-		}
 		seen[s] = struct{}{}
-		out = append(out, s)
 	}
-	slices.Sort(out)
-	return out
+	return slices.Sorted(maps.Keys(seen))
 }
 
 func resolveCommandInRuntime(ctx context.Context, runner runtimeexec.Runner, settings fleet.RuntimeSettings, defaultName, preferred string) (string, bool) {


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Notes
- Behavior preserved. Both old and new pipelines produce a trimmed, empty-filtered, deduplicated, lexically sorted slice. The map-based dedupe is idempotent, so dropping the explicit `if _, ok := seen[s]; ok { continue }` check is safe.
- Idiom matches in-file precedent: `unavailableBackendStatuses` in the same file already uses `slices.Sorted(maps.Keys(seen))`, so this brings `dedupeSorted` into alignment.
- Go version is fine: `go.mod` is `go 1.25.5`; `maps.Keys` (returning `iter.Seq`) and `slices.Sorted` are both Go 1.23+. The `maps` import was already present before this PR.
- Existing tests `TestParseModelsCodexDebugCatalog` and `TestParseModelsClaudeMarkdownTable` exercise `parseModels → dedupeSorted` end-to-end with realistic deduped+sorted expectations. CI green on `89c3134a`.
- Subtle (invisible) edge case worth noting for the record: when `in` is non-empty but every entry trims to `""`, the old code returned a non-nil empty slice (`make([]string, 0, len(in))`), while the new code returns `nil` (`slices.Sorted` over an empty `iter.Seq` starts from a nil base via `slices.Collect`). All current callers branch on `len(...) == 0` and the `Models` field is `json:"models,omitempty"`, so this is not observable. Not a finding; flagging only because someone reading `git blame` may eventually wonder.

### Summary
Clean, idiomatic refactor matching existing in-file style; behavior is equivalent and well-covered by existing tests.

PR #561, opened 2026-05-15
fix: persist explicit admin role

### LINKED ISSUE #539: Add explicit admin role for users
## Goal

Replace implicit admin detection (`id = MIN(id)`) with an explicit `users.is_admin` role flag.

Blocked by: #538

This should land after #538 to avoid concurrent conflicts in `internal/store/auth.go`, auth migrations, and auth tests.

## Why

The current model treats the first-created user as admin by calculating `id = (SELECT MIN(id) FROM users)`. That works, but it is implicit and fragile:

- admin status changes if data is repaired or imported unexpectedly,
- code readers have to know the convention,
- future multi-admin support is awkward,
- last-admin protection is harder to express clearly.

Admin status should be explicit persisted state.

## Required implementation

Implement in one PR.

### 1. Add explicit admin column

Add a migration:

```sql
ALTER TABLE users ADD COLUMN is_admin INTEGER NOT NULL DEFAULT 0;
```

Then backfill existing installs:

- If users exist, mark the current first user (`MIN(id)`) as admin.
- New empty installs remain empty until bootstrap.

### 2. Bootstrap creates an admin user

`BootstrapUser` must create the first user with `is_admin = 1`.

### 3. Scan admin from the DB

Update user scans/auth identity scans so `User.IsAdmin` comes from `users.is_admin`, not `id = MIN(id)`.

Update:

- `ListUsers`
- `GetUser`
- `AuthenticateToken`
- any other auth query currently calculating admin status.

### 4. Protect the last admin

- Deleting the last admin must be forbidden.
- Deleting a non-admin remains allowed.
- Deleting one admin while at least one other admin exists should be allowed.

If there is no existing admin-demotion endpoint, do not add one in this issue. If one exists, it must also protect the last admin.

### 5. User management rules

Keep current intended authorization semantics:

- Non-admin users may create/revoke/list their own API tokens.
- Only admins may manage users.
- The first bootstrapped dashboard user is admin.

The authorization middleware/handlers should continue enforcing these rules, but now based on explicit `is_admin`.

### 6. Tests

Update auth tests to cover:

- migration marks existing first user admin,
- bootstrap creates admin user,
- `ListUsers` and `AuthenticateToken` return DB-backed admin status,
- deleting last admin is forbidden,
- deleting non-admin is allowed,
- deleting one admin is allowed when another admin remains,
- non-admin users can still manage their own API tokens,
- non-admin users cannot manage users.

Run:

```bash
go test ./internal/store ./internal/daemon
```

## Acceptance criteria

- `users.is_admin` exists and is the source of truth.
- `User.IsAdmin` is not derived from minimum user ID anymore.
- Bootstrap user is admin.
- Existing installs get one admin during migration.
- Last admin cannot be removed.
- User/API-token authorization behavior remains aligned with the current product model.

## Non-goals

- Do not add full role-based access control.
- Do not add admin demotion UI unless required by an existing API surface.
- Do not redesign sessions/API tokens.

eloylp:
## Summary

Closes #539.

- Adds a `users.is_admin` migration and backfills the current first user as admin for existing installs.
- Persists the bootstrap user as admin and reads admin status from the database in user and token auth paths.
- Protects the last enabled admin from deletion while allowing admin deletion when another admin remains.
- Adds store and migration coverage for DB-backed admin status, bootstrap/admin auth identity, and last-admin behavior.

## Test plan

- `env -u GH_TOKEN -u GITHUB_TOKEN -u CLAUDE_CODE_OAUTH_TOKEN -u CODEX_AUTH_JSON_BASE64 -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u ANTHROPIC_AUTH_TOKEN go test -race ./internal/store ./internal/daemon`
- `git diff --check`

Note: an initial unsanitized `go test ./...` hit the existing backend discovery environment-sensitive failure in this runner; rerunning with backend credential variables unset passed.

### DIFF
--- a/internal/store/auth.go
+++ b/internal/store/auth.go
@@ -74,7 +74,7 @@ func (s *Store) UserCount(ctx context.Context) (int, error) {
 
 func (s *Store) ListUsers(ctx context.Context) ([]User, error) {
 	rows, err := s.db.QueryContext(ctx, `
-		SELECT id,username,created_at,updated_at,last_login_at,disabled_at,id=(SELECT MIN(id) FROM users)
+		SELECT id,username,created_at,updated_at,last_login_at,disabled_at,is_admin
 		FROM users ORDER BY username ASC, id ASC`)
 	if err != nil {
 		return nil, fmt.Errorf("auth: list users: %w", err)
@@ -121,7 +121,7 @@ func (s *Store) CreateUser(ctx context.Context, username, password string) (User
 
 func (s *Store) GetUser(ctx context.Context, id int64) (User, error) {
 	row := s.db.QueryRowContext(ctx, `
-		SELECT id,username,created_at,updated_at,last_login_at,disabled_at,id=(SELECT MIN(id) FROM users)
+		SELECT id,username,created_at,updated_at,last_login_at,disabled_at,is_admin
 		FROM users WHERE id=?`, id)
 	user, err := scanUser(row)
 	if errors.Is(err, sql.ErrNoRows) {
@@ -143,12 +143,20 @@ func (s *Store) DeleteUser(ctx context.Context, id int64) error {
 	}
 	defer tx.Rollback()
 
-	var adminID int64
-	if err := tx.QueryRowContext(ctx, "SELECT COALESCE(MIN(id), 0) FROM users").Scan(&adminID); err != nil {
-		return fmt.Errorf("auth: admin lookup: %w", err)
+	var isAdmin int
+	if err := tx.QueryRowContext(ctx, "SELECT is_admin FROM users WHERE id=?", id).Scan(&isAdmin); errors.Is(err, sql.ErrNoRows) {
+		return ErrAuthNotFound
+	} else if err != nil {
+		return fmt.Errorf("auth: lookup delete user: %w", err)
 	}
-	if adminID == id {
-		return ErrAuthForbidden
+	if isAdmin != 0 {
+		var adminCount int
+		if err := tx.QueryRowContext(ctx, "SELECT COUNT(*) FROM users WHERE is_admin != 0 AND disabled_at IS NULL").Scan(&adminCount); err != nil {
+			return fmt.Errorf("auth: count admins: %w", err)
+		}
+		if adminCount <= 1 {
+			return ErrAuthForbidden
+		}
 	}
 	res, err := tx.ExecContext(ctx, "DELETE FROM users WHERE id=?", id)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[low]** `internal/store/auth.go:155` — admin-count predicate filters on `is_admin != 0 AND disabled_at IS NULL`, so a *disabled* admin row cannot be removed unless another *enabled* admin exists. This narrows issue #539's "deleting one admin while at least one other admin exists should be allowed" to "another *enabled* admin remains." The PR description is explicit about the "enabled" qualifier, and there is no `DisableUser` API in the auth surface today, so the case is theoretical — but if a `disabled_at` is ever set out of band, the only way to delete that admin row would be to first promote and re-enable a second admin. Worth a comment on the predicate or a follow-up issue.
- **[low]** `internal/store/auth.go:144-159` — `DeleteUser` reads `is_admin` and counts admins inside a `BeginTx(ctx, nil)` (deferred-mode) transaction, with no explicit row lock. Two concurrent admin deletions on different connections could each observe `adminCount=2` and both succeed, dropping to zero admins. SQLite's single-writer behavior typically serializes writes (and the second write would likely surface `SQLITE_BUSY`), and this race shape pre-existed the change, so it's not a regression. Mentioning it for visibility — `BEGIN IMMEDIATE` or a conditional `DELETE … WHERE id=? AND (is_admin=0 OR (SELECT COUNT(*) FROM users WHERE is_admin!=0 AND disabled_at IS NULL) > 1)` checked via `RowsAffected` would close the gap atomically.
- **[low]** `internal/store/migrations/029_auth_admin_role.sql:1-9` — the `CREATE TABLE IF NOT EXISTS users (...)` at the head of 029 is a no-op in production (migration 019 already created the table) and the column list duplicates 019's definition. If 019 ever evolves (added column, changed default), this redefinition silently goes stale. Consider dropping it and trusting the migration sequence.

### Summary
Clean, well-scoped fix that satisfies the issue's acceptance criteria: explicit `users.is_admin` is now the source of truth across `ListUsers`, `GetUser`, `AuthenticateToken`, and `DeleteUser`; bootstrap persists `is_admin=1`; the migration backfills the existing first user idempotently via `NOT EXISTS`; last-admin deletion is blocked while a second admin still allows the first to be removed. Tests cover migration backfill against a seeded pre-029 schema, DB-driven admin status across read paths and bootstrap auth identity, and both halves of the last-admin behavior.

PR #563, opened 2026-05-15
refactor: simplify runtime command slices

eloylp:
## Summary

- Replaces the manual nested append used to inject Claude MCP args with `slices.Insert`.
- Uses `slices.Concat` when building the shell entrypoint command prefix plus backend command.
- Keeps runtime command ordering and environment behavior unchanged.

## Test plan

- `git diff --check`

### DIFF
--- a/internal/runtime/backend_setup.go
+++ b/internal/runtime/backend_setup.go
@@ -65,7 +65,7 @@ func WrapBackendCommand(backendName string, command []string, env []string, opts
 		return command, env
 	}
 	if strings.HasPrefix(backendName, "claude") && getEnv(env, "GITHUB_TOKEN") != "" && !slices.Contains(command[1:], "--mcp-config") {
-		command = append(command[:1], append([]string{"--mcp-config", RunnerClaudeMCPPath}, command[1:]...)...)
+		command = slices.Insert(command, 1, "--mcp-config", RunnerClaudeMCPPath)
 	}
 	return shellEntrypoint(command, BackendSetupScript(backendName, opts)), env
 }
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/runtime/backend_setup.go:68` — `slices.Insert(command, 1, "--mcp-config", RunnerClaudeMCPPath)` is element-for-element equivalent to the old nested `append(command[:1], append([]string{...}, command[1:]...)...)`. Result is `[command[0], "--mcp-config", PATH, command[1:]...]` in both cases. The `len(command) == 0` early return above guarantees `len(command) >= 1`, so index `1` is always a valid insertion point (inserts at end when `len == 1`). Command/argument ordering is preserved exactly. The new form also removes the well-known nested-append aliasing footgun — a readability win with no behavior change.
- **[info]** `internal/runtime/backend_setup.go:80` — `slices.Concat(out, command)` yields the same `[/bin/sh, -lc, setup, agents-runner, command...]` sequence as `append(out, command...)`. `out` is a fresh literal each call, so neither form risks caller aliasing; `$0=agents-runner`, then `cmd=$1; shift; exec "$cmd" "$@"` still receives `command` in original order.
- **[info]** Go version compatibility verified: `go.mod` declares `go 1.25.5`. `slices.Insert` (Go 1.21+) and `slices.Concat` (Go 1.22+) are both well within range. CI `test` job is green on `9e615b8`.
- **[note]** Both the old and new code can mutate the caller's `command` backing array in place when it has spare capacity — this is pre-existing behavior, not introduced or worsened by this PR, and the caller does not reuse the slice. No action needed.

### Summary
Clean, idiomatic, behavior-preserving simplification; ordering and Go-version compatibility both verified safe. APPROVE.

PR #564, opened 2026-05-15
fix: add self-service password change

### LINKED ISSUE #562: Add self-service password change for dashboard users
## Problem

Dashboard users cannot currently change or reset their own password.

The current auth surface supports:

- first-user bootstrap
- username/password login
- logout
- admin-created users
- admin removal of non-admin users
- API token creation/revocation

But there is no endpoint or UI flow for changing the password of the currently authenticated user.

Because passwords are stored as one-way hashes, the current password cannot be recovered or displayed. If the bootstrapped admin forgets the password and the browser session expires, the install can lose admin access without direct database surgery.

## Required behavior

Add self-service password change for authenticated dashboard users.

Suggested REST endpoint:

- `POST /auth/me/password`

Request body:

```json
{
  "current_password": "old password",
  "new_password": "new password"
}
```

Expected semantics:

- Requires an authenticated browser/API session like other sensitive auth routes.
- Verifies `current_password` against the current user's stored password hash.
- Rejects empty username/password-equivalent inputs.
- Updates only the authenticated user's password hash.
- Updates `users.updated_at`.
- Does not expose password hashes or current password material.
- Works for both admin and non-admin users.
- Keeps MCP free of password administration tools; this is UI/REST only.

Session/token handling:

- At minimum, keep the current session valid so the user is not logged out mid-flow.
- Prefer revoking other browser sessions for the same user after password change.
- Do not revoke API tokens automatically unless the implementation explicitly documents that behavior and the UI warns the user.

## Store changes

Add store-level functionality rather than doing SQL directly in the handler.

Suggested store API:

- `ChangeUserPassword(ctx, userID int64, currentPassword, newPassword string) error`

The store method should:

- load the user by ID
- reject disabled/missing users
- verify `currentPassword`
- hash `newPassword` using the same password hashing path as user creation/bootstrap
- update `password_hash` and `updated_at`

Keep password hashing centralized.

## UI changes

Add a small password-change form under `Config -> Authentication`.

Fields:

- Current password
- New password
- Confirm new password

UI behavior:

- All fields blank by default.
- Validate new password confirmation client-side before submitting.
- Show clear success/error feedback.
- Clear all password fields after success.
- Do not add this to MCP.

## Tests

Add focused tests covering:

- authenticated user can change own password
- old password no longer works after change
- new password works after change
- wrong current password is rejected
- unauthenticated requests are rejected
- non-admin users can change their own password
- one user cannot change another user's password through this endpoint
- empty new password is rejected

Also keep existing auth route protection tests updated so `/auth/me/password` is covered by the auth barrier.

## Docs

Update:

- `docs/api.md`
- `docs/security.md`
- `docs/ui.md`
- `AGENTS.md` / `CLAUDE.md` operational auth notes if needed

Document clearly that passwords cannot be recovered, only changed while authenticated.

## Non-goals

- No forgot-password email flow.
- No MCP password management surface.
- No admin password reset UI in this issue.
- No role model beyond the existing bootstrapped admin / non-admin distinction.

eloylp:
## Summary

Closes #562.

- Adds `Store.ChangeUserPassword` with current-password verification, centralized bcrypt hashing, disabled-user rejection, and `updated_at` updates.
- Adds `POST /auth/me/password` for authenticated users and covers admin/non-admin password-change flows.
- Adds a Config -> Authentication password form, UI tests for submit/mismatch behavior, and REST/security/UI docs.

## Test plan

- `go test ./internal/store`
- `npm test -- --run` in `internal/ui`
- `npm run build` in `internal/ui`
- `git diff --check`


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[nit]** `internal/daemon/auth.go:243` — `req.CurrentPassword == "" || req.NewPassword == ""` rejects empty inputs with `400 invalid password request`, while a wrong-but-non-empty current password returns `401 invalid credentials`. That split is consistent with `handleAuthBootstrap`/`handleAuthLogin` precedent in this file, so this is fine as-is — just calling it out so future readers don't try to "unify" them.
- **[nit]** `internal/store/auth.go:194` — `verified, _ := verifyPassword(currentPassword, stored)` drops the `needsRehash` bool. That's correct here because `ChangeUserPassword` immediately writes a fresh `bcrypt` hash via `hashPassword(newPassword)`, so a legacy `pbkdf2-sha256` stored hash is replaced regardless. Worth a one-line comment only if you think a future maintainer might "fix" the underscore.
- **[suggestion]** `internal/store/auth.go:178` — Successful password change leaves other browser sessions (`auth_tokens` rows with `kind='session'`) for the same `user_id` valid. Issue #562 only "prefers" revoking those, so this is not a blocker — but if/when you decide to harden, a single `UPDATE auth_tokens SET revoked_at=datetime('now') WHERE user_id=? AND kind=? AND id<>? AND revoked_at IS NULL` inside the same tx, keyed off the caller's current session token id, is the natural shape. Document API-token retention in the UI when you do.
- **[info]** Issue acceptance bullet "one user cannot change another user's password through this endpoint" is enforced by construction — `handleAuthMePassword` passes `identity.User.ID` from the authenticated session, never a body-supplied id. `TestBuildHandlerAuthNonAdminChangesOnlyOwnPassword` indirectly confirms by checking the admin's password is unaffected after the operator's change. No direct test possible without adding a parameter that doesn't exist.
- **[info]** Disabled-user defense in `ChangeUserPassword` (`disabled.Valid || !verified` collapses both into `ErrAuthInvalid`) is good — avoids leaking enable/disable state and matches the `Login` pattern. The middleware should already block disabled users via `AuthenticateToken`, so this is layered defense.
- **[info]** UI form lets the user click "Change password" with all fields blank (no client-side empty check); the server returns `400 invalid password request` and the text is surfaced. Functional, just a roundtrip you could skip with a client-side guard. Not worth a rev on its own.

### Summary
Clean, well-scoped fix for #562. Store API is centralized (`Store.ChangeUserPassword`), handler is thin, identity is session-derived so cross-user changes are impossible, `MaxBodyBytes` bounds input, errors funnel to `ErrAuthInvalid`/`ErrAuthNotFound` → `401` without leaking which one. Tests cover wrong/empty/success at both store and handler layers plus old-password-fails / new-password-works and the non-admin self-service path; protected-route table picks up `/auth/me/password`. UI tests exercise both submit and mismatch paths. Docs (api/security/ui/AGENTS/CLAUDE) updated in the same PR. CI was still pending at review time; human merge after green.

PR #566, opened 2026-05-15
fix: preserve omitted runtime settings

### LINKED ISSUE #546: Make runtime settings updates preserve omitted fields
## Problem

`update_runtime` on MCP and `PATCH /runtime` on REST look like partial-update surfaces, but currently behave like replacement-with-defaults.

MCP `runtimeSettingsFromRequest` starts from an empty `fleet.RuntimeSettings`, applies only supplied fields, then calls `WriteRuntimeSettings`. `WriteRuntimeSettings` normalizes missing `runner_image` to the default image and persists the whole runtime settings object. This means a caller who only wants to update one field, such as `cpus`, can accidentally clear or reset other runtime settings.

REST has a similar risk because `HandleUpdateRuntime` decodes a full `fleet.RuntimeSettings` for both `PUT` and `PATCH`.

## Desired behavior

- `PATCH /runtime` and MCP `update_runtime` should be true partial updates.
- `PUT /runtime` may remain full replacement if useful, but its behavior should be explicit in code/docs/tests.
- Updating one field must preserve all omitted current runtime fields.

## Required changes

1. Add a runtime patch shape with pointer fields, for example:

```go
type RuntimeSettingsPatch struct {
    RunnerImage *string
    Constraints struct {
        CPUs *string
        Memory *string
        PidsLimit *int64
        TimeoutSeconds *int
        NetworkMode *string
    }
}
```

The exact location can be store/fleet/daemon as appropriate, but validation should remain centralized enough that REST and MCP use the same semantics.

2. Add a store method that reads current settings, applies the patch, normalizes/validates, then persists.

3. Update REST:

- `PATCH /runtime` uses patch semantics.
- `PUT /runtime` can continue full replacement, or if simpler, make both methods patch semantics but document that explicitly.

4. Update MCP:

- `update_runtime` must read current settings or call the new patch method so omitted fields are preserved.
- Keep accepting the current arguments: `runner_image`, `cpus`, `memory`, `pids_limit`, `timeout_seconds`, `network_mode`.
- Allow explicit empty strings for string fields where clearing is valid, e.g. `cpus`, `memory`, `network_mode`. Be careful not to confuse omitted with empty.

5. Update docs/tool descriptions if they currently imply a behavior that changes.

## Tests

Add tests proving:

- Existing runtime settings with `runner_image`, `cpus`, `memory`, and `network_mode` are preserved when MCP updates only `timeout_seconds`.
- REST `PATCH /runtime` preserves omitted fields.
- Full replacement behavior for `PUT /runtime` is tested if it remains distinct.
- Explicit empty string can clear clearable string constraints without resetting `runner_image` accidentally.

## Acceptance criteria

- Partial runtime updates no longer reset omitted fields.
- REST and MCP runtime update semantics are aligned and documented.
- `go test ./internal/store ./internal/daemon ./internal/mcp` passes.

eloylp:
## Summary

Closes #546.

- Adds a shared `store.RuntimeSettingsPatch` path that reads current settings, applies only supplied fields, normalizes/validates, and persists.
- Keeps `PUT /runtime` as full replacement while making `PATCH /runtime` preserve omitted fields.
- Routes MCP `update_runtime` through patch semantics, including explicit empty-string clearing for clearable string constraints and explicit zero for numeric constraints.
- Updates REST/MCP docs and tool description to document patch behavior.

## Test plan

- `go test ./internal/store ./internal/daemon ./internal/mcp`
- `git diff --check`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -136,7 +136,7 @@ Duplicate webhook deliveries are suppressed via `X-GitHub-Delivery` with a TTL c
 | Method | Path | Description |
 |---|---|---|
 | `GET` | `/runtime` | Read global runner image and container constraints. |
-| `PUT/PATCH` | `/runtime` | Replace global runner image and constraints. Secret values are not accepted here. |
+| `PUT/PATCH` | `/runtime` | `PUT` replaces global runner image and constraints; `PATCH` updates only supplied fields and preserves omitted settings. Secret values are not accepted here. |
 | `PUT/PATCH` | `/workspaces/{workspace}/runtime` | Set or clear the selected workspace's runner image override. |
 
 Runtime settings are also included in `/config`, `/export`, and `/import`. Credentials are daemon environment variables and are never returned by these routes.
--- a/docs/api.md
+++ b/docs/api.md
@@ -140,6 +140,7 @@ Duplicate webhook deliveries are suppressed via `X-GitHub-Delivery` with a TTL c
 | `PUT/PATCH` | `/workspaces/{workspace}/runtime` | Set or clear the selected workspace's runner image override. |
 
 Runtime settings are also included in `/config`, `/export`, and `/import`. Credentials are daemon environment variables and are never returned by these routes.
+For global `/runtime`, empty string clears string constraints such as `cpus`, `memory`, and `network_mode`; empty `runner_image` resets to the built-in default because the field cannot be unset.
 
 ## AI runner contract
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -96,6 +96,7 @@ runtime:
 Runtime settings are fleet state because operators may need to switch runner images or constraints without rebuilding the daemon. They are stored in SQLite, returned by `/config`, included in import/export, and editable through Config -> Runtime, REST, and MCP. They are not startup env overrides; use the database-backed surfaces for runner image and container policy changes.
 
 The global runner image applies to every run unless a workspace sets `runner_image`. Constraints are passed to Docker where supported: CPU, memory, PID limit, timeout, and network mode. Advanced egress filtering is not part of the v1 runtime settings.
+When patching global runtime settings through REST or MCP, omitted fields are preserved. Empty string clears string constraints such as `cpus`, `memory`, and `network_mode`; empty `runner_image` resets to the built-in default because the global image cannot be unset.
 
 ## `backends`
 
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -116,7 +116,7 @@ These tools mirror the `/runners` REST surface; see [api.md](api.md#runners-mana
 |---|---|
 | `get_config` | Current fleet config snapshot. |
 | `get_runtime` | Read global runner image and container constraints. |
-| `update_runtime` | Replace global runner image and basic constraints. Secret values are not accepted. |
+| `update_runtime` | Patch global runner image and basic constraints, preserving omitted settings. Secret values are not accepted. |
 | `update_workspace_runtime` | Set or clear one workspace's runner image override. |
 | `export_config` | Fleet config as YAML (round-trippable via `import_config`). |
 | `import_config` | Import YAML config. `mode=replace` prunes missing entries. |
--- a/internal/daemon/config/config.go
+++ b/internal/daemon/config/config.go
@@ -77,6 +77,22 @@ func (h *Handler) HandleRuntime(w http.ResponseWriter, _ *http.Request) {
 // HandleUpdateRuntime serves PUT/PATCH /runtime. Secret values are not part of
 // this shape; credentials are injected from daemon environment at run time.
 func (h *Handler) HandleUpdateRuntime(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodPatch {
+		var patch store.RuntimeSettingsPatch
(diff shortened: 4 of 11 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[low]** `internal/store/runtime_settings.go:47` — `PatchRuntimeSettings` is read‑modify‑write without a transaction (`ReadRuntimeSettings` then `WriteRuntimeSettings`). Two concurrent patches can lose updates: A reads, B reads, A writes its patched view, B writes its patched view and silently clobbers A's change. Runtime config rarely sees concurrent writers, so practical risk is low, but it's exactly the kind of TOCTOU a partial-update API is expected to defend against. Wrapping the read+write in a `BeginTx`/`Commit` (same pattern as `Store.ChangeUserPassword` introduced in #564) closes the gap; SQLite's row-level write lock will serialize correctly.
- **[low]** `internal/store/runtime_settings.go` + `internal/mcp/tools_config.go` — explicit empty‑string for `runner_image` (REST PATCH `{"runner_image": ""}` or MCP `runner_image=""`) silently resets to `DefaultRunnerImage` because `NormalizeRuntimeSettings` re-defaults blank images. That's a defensible choice (the daemon can't run without a runner image), but it's not documented anywhere a caller would see it. The MCP tool's `runner_image` description and the REST `PATCH /runtime` docs both call out partial updates and empty-string clearing for `cpus`/`memory`/`network_mode`; they should add one sentence noting that `runner_image` is exempt — empty/whitespace resets to the default, it does not "clear" the field. Worth a one-line test asserting the behavior so the contract is pinned.
- **[low]** Tests cover preservation and empty‑string clearing, but neither the REST nor the MCP path exercises the explicit‑zero numeric case (`{"constraints": {"pids_limit": 0}}` or `{"constraints": {"timeout_seconds": 0}}`). This is exactly the semantic change the PR introduces over the prior MCP behavior (where `pids > 0` was the gate), and the dispatch reason flagged it. With pointer fields the code is correct — `*int64` is non‑nil when JSON has `0`, and `validateRuntimeSettings` allows `>= 0` — but without a test, a future refactor back to value semantics would silently re-break it.
- **[nit]** `internal/store/facade.go:78` — `PatchRuntimeSettings` is glued onto the runtime block with no blank line separating it from `WriteRuntimeSettings` and `ReadWorkspace`. The rest of facade.go uses single blank lines or section comments between method groups; consistent spacing is a small readability win.

### Summary
Correct, well-tested implementation of the issue's required shape (pointer-field patch, shared store path, REST `PATCH` partial / `PUT` replace, MCP routed through the same patch). Issue acceptance criteria are all met. Suggestions above are non-blocking polish: a transactional read-modify-write, a couple of doc/test additions to pin contract edges (`runner_image` empty-string, explicit zero numeric updates), and one whitespace nit.

eloylp:
Updated PR #566 for the review feedback:

- wrapped runtime patch read-modify-write in a SQL transaction
- documented that global `runner_image: ""` resets to the built-in default rather than clearing the field
- added store/MCP coverage for empty runner image reset
- added REST/MCP coverage for explicit zero numeric patch values
- fixed the facade spacing nit

Verification:
- `go test ./internal/store ./internal/daemon ./internal/mcp`
- `go test -race ./internal/store ./internal/daemon ./internal/mcp`
- `git diff --check`

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- All four follow-up items from the prior review on `62924d58` are addressed in `2c8c8d2`:
  - `internal/store/runtime_settings.go:48` — `PatchRuntimeSettings` now wraps `ReadRuntimeSettings` + `writeRuntimeSettings` in a `BeginTx`/`Commit` with `defer tx.Rollback()`. Uses the same `querier` for read and a new `runtimeSettingsWriter` (Exec-only) interface for the write so the existing `WriteRuntimeSettings(*sql.DB, …)` exported signature is preserved. SQLite serializes the two operations under one write transaction; the prior TOCTOU window is closed.
  - `docs/api.md`, `docs/configuration.md`, and the `update_runtime` MCP tool description (`internal/mcp/tools.go:312-313`) all state that empty `runner_image` resets to the built-in default rather than clearing the field. Wording is consistent across the three surfaces.
  - REST `TestStoreCRUDRuntimePatchPreservesOmittedFields` exercises `pids_limit: 0` (was 256) and asserts the result becomes 0 — pins the pointer-field semantics against future refactors. MCP `TestToolUpdateRuntimePreservesOmittedFields` mirrors it with `float64(0)`.
  - `TestPatchRuntimeSettingsEmptyRunnerImageResetsDefault` (store) and `TestToolUpdateRuntimeEmptyRunnerImageResetsDefault` (MCP) pin the empty-string-resets-to-default contract.
  - `internal/store/facade.go` spacing nit fixed: blank line before `PatchRuntimeSettings`.
- **[nit]** `internal/mcp/tools_config.go:96` — the `float64` arm of `optionalIntArg` uses `v < float64(math.MinInt) || v > float64(math.MaxInt)`. Because `float64` has only ~52 bits of mantissa, `float64(math.MaxInt64)` rounds up to `math.MaxInt64+1`, so values near 2^63 can pass the bounds check yet be unrepresentable by `int64` on cast. Not a real risk for `pids_limit`/`timeout_seconds` (sane operator inputs are six digits at most), so flagging for awareness only.

### Summary
Follow-up cleanly addresses the prior review. Patch path is now transactional, the runner_image reset contract is documented in all three caller-facing places, and the explicit-zero numeric + empty runner_image edges are pinned by tests on both REST and MCP. CI test job green on `2c8c8d2`.

PR #567, opened 2026-05-15
fix: persist observability records synchronously

### LINKED ISSUE #529: Make observability persistence shutdown-safe
## Problem

The daemon lifecycle is intended to be structured: components that spawn goroutines expose `Run(ctx) error`, own their children, and do not return until those children have exited. That lets daemon shutdown stop producers, drain consumers, and wait for in-flight work in a predictable order.

`internal/observe.Store` currently violates that model for several persistence paths:

- `RecordEvent`
- `RecordSpan`
- `RecordDispatch`

Each method starts a fire-and-forget goroutine to write to SQLite. Those goroutines are not owned by any `Run(ctx)` component and are not waited during shutdown. This means event/span/dispatch rows can theoretically be lost if the process exits while those goroutines are still running. It also means SSE can publish an event/span before the DB row exists.

`RecordStep` already writes synchronously before publishing, which is the safer model for this daemon.

## Desired Direction

Remove the unowned persistence goroutines from `internal/observe.Store` and make observability persistence deterministic and shutdown-safe.

Preferred implementation for this issue:

- Make `RecordEvent`, `RecordSpan`, and `RecordDispatch` persist synchronously.
- Publish SSE only after the DB write attempt has completed.
- Keep existing logging behavior for persistence errors.
- Do not introduce a new async writer unless there is a demonstrated need. If async persistence is introduced, it must be an owned component with `Run(ctx) error`, a bounded queue, flush-on-shutdown semantics, and daemon lifecycle wiring. Synchronous writes are expected to be enough for this issue.

## Acceptance Criteria

- `RecordEvent` no longer starts a goroutine for the SQLite insert.
- `RecordSpan` no longer starts a goroutine for the SQLite insert.
- `RecordDispatch` no longer starts a goroutine for the SQLite insert.
- `RecordStep` behavior remains synchronous and unchanged unless cleanup is needed for consistency.
- SSE publication for events/traces/dispatches happens after the synchronous persistence path returns.
- Existing tests pass.
- Add or update tests proving the rows are visible immediately after calling the recorder method, without sleeps or polling.
- Add a test or strengthen an existing test to cover shutdown-safety expectations if there is a natural daemon/observe test hook.

## Non-goals

- Do not redesign the full daemon shutdown model in this issue.
- Do not change trace/event/dispatch schemas.
- Do not change frontend behavior.
- Do not make SQLite writes best-effort async again under a different name.

## Files To Inspect

- `internal/observe/observe.go`
- `internal/observe/observe_test.go`
- `internal/daemon/daemon.go`
- `internal/workflow/engine.go`

## Rationale

Agent runs are much slower than these small local SQLite inserts, and observability rows are part of the operator-facing source of truth. Correctness and deterministic shutdown are more important than avoiding a few milliseconds in the workflow path.

eloylp:
## Summary

Closes #529.

- Removes fire-and-forget SQLite writes from `RecordEvent`, `RecordSpan`, and `RecordDispatch`.
- Keeps existing persistence error logging while publishing event/trace SSE only after the write attempt returns.
- Updates observe, runners, and MCP tests to rely on immediate row visibility instead of sleeps or polling.

## Test plan

- `go test ./internal/observe`
- `go test -race ./internal/observe`
- `go test ./internal/observe ./internal/daemon/runners ./internal/mcp`
- `go test -race ./internal/observe ./internal/daemon/runners ./internal/mcp`
- `git diff --check`

### DIFF
--- a/internal/daemon/runners/runners_test.go
+++ b/internal/daemon/runners/runners_test.go
@@ -109,15 +109,6 @@ func TestList_CompletedEventFannedOutToTwoAgentsShowsTwoRows(t *testing.T) {
 		StartedAt: now, FinishedAt: now.Add(2 * time.Second),
 		Status: "error", ErrorMsg: "claude command timed out after 10m0s",
 	})
-	// RecordSpan persists asynchronously into SQLite; poll until both
-	// rows are visible before asking the handler to JOIN.
-	deadline := time.Now().Add(2 * time.Second)
-	for time.Now().Before(deadline) {
-		if len(fx.observe.TracesByRootEventID("ev-fan")) >= 2 {
-			break
-		}
-		time.Sleep(20 * time.Millisecond)
-	}
 	_ = fx.store.MarkEventCompleted(id)
 
 	req := httptest.NewRequest(http.MethodGet, "/runners", nil)
--- a/internal/mcp/tools_test.go
+++ b/internal/mcp/tools_test.go
@@ -831,9 +831,7 @@ func TestToolTriggerAgentMissingArgs(t *testing.T) {
 }
 
 // seedEvent inserts an event row directly into the SQLite events table so
-// observe.Store.ListEvents reads it back deterministically. The Store's own
-// RecordEvent is async (goroutine), which would race with the immediate
-// ListEvents call our tool tests need.
+// observe.Store.ListEvents reads fixture rows back deterministically.
 func seedEvent(t *testing.T, db *sql.DB, ev observe.TimestampedEvent) {
 	t.Helper()
 	payload, _ := json.Marshal(ev.Payload)
@@ -847,7 +845,6 @@ func seedEvent(t *testing.T, db *sql.DB, ev observe.TimestampedEvent) {
 }
 
 // seedSpan inserts a trace span row directly into the SQLite traces table.
-// Same async-vs-sync rationale as seedEvent.
 func seedSpan(t *testing.T, db *sql.DB, sp observe.Span) {
 	t.Helper()
 	workspaceID := fleet.NormalizeWorkspaceID(sp.WorkspaceID)
@@ -1111,15 +1108,6 @@ func TestToolGetTracePromptReturnsBody(t *testing.T) {
 		Status: "success",
 		Prompt: "system: do the thing\n\nuser: please",
 	})
-	// RecordSpan persists asynchronously; poll briefly until visible.
-	deadline := time.Now().Add(2 * time.Second)
-	for time.Now().Before(deadline) {
-		if got, err := deps.Observe.PromptForSpan("sp-prompt"); err == nil && got != "" {
-			break
-		}
-		time.Sleep(20 * time.Millisecond)
-	}
-
 	req := mcpgo.CallToolRequest{}
 	req.Params.Arguments = map[string]any{"span_id": "sp-prompt"}
 	res, err := toolGetTracePrompt(deps)(context.Background(), req)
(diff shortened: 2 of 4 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[nit]** `internal/observe/observe.go:651-664,710-722` — On `s.db.Exec` failure the code still publishes to `EventsSSE` / `TracesSSE`. Issue #529 explicitly says "Publish SSE only after the DB write attempt has completed" (attempt, not success), so this matches the stated contract — flagging for awareness rather than as a change request. A SQLite write failure is now logged once and consumers see an event/span on SSE that won't be in `ListEvents` / `ListTraces`. Same posture as the prior async code, just deterministic now.
- **[nit]** `internal/observe/observe.go:732-742` — `RecordDispatch` has no SSE publish (and didn't before), so the SSE-ordering criterion in #529 is vacuously satisfied here. Worth noting in case future work adds a `DispatchSSE` hub — the synchronous-then-publish pattern would need to be applied consistently.
- **[nit]** Issue #529 mentions "Add a test or strengthen an existing test to cover shutdown-safety expectations if there is a natural daemon/observe test hook." No dedicated shutdown test was added, but the property follows logically from removing the unowned goroutines (no in-flight writers left at shutdown), and `TestStoreRecordEvent…`, `TestStoreRecordSpan…`, `TestStoreRecordDispatch…` now assert immediate row visibility without polling, which is the observable proof that the write completed before `RecordX` returned. Acceptable since the issue gates this on "if there is a natural hook."

### Notes on the explicit review focus

- **Shutdown safety**: All three persistence paths now block in `s.db.Exec` and return only after the write attempt completes. There are no remaining unowned goroutines spawned by these methods, so daemon shutdown can stop the workflow engine and trust that no observability writers are still queued. Matches the `RecordStep` / `RecordSteps` model already in use.
- **SSE ordering**: For `RecordEvent` and `RecordSpan`, `s.EventsSSE.Publish` / `s.TracesSSE.Publish` now run strictly after `s.db.Exec` returns. The prior `go func()` form let SSE fire while the goroutine was still inserting, which is the race the issue calls out. Order is now deterministic.
- **Blocking-write risk**: Callers are the workflow engine on agent-run completion, where agent runs dominate latency by orders of magnitude over a SQLite local insert (issue's own rationale). SQLite's single-writer lock already serializes writes from the prior async path too, so there's no new contention surface — the wait now happens on the caller stack instead of on a detached goroutine. No risk of deadlock since these methods don't take other locks while holding `db.Exec`. `RecordEvent` / `RecordSpan` / `RecordDispatch` are not holding `s.stepMu`, so they don't contend with `RecordStep` / `RecordSteps` at the Go level (only at the SQLite-driver level, which already serialized).

### Diff scope
- `RecordEvent`, `RecordSpan`, `RecordDispatch` rewritten to drop `go func()` and publish SSE after the write attempt.
- Tests in `internal/observe`, `internal/daemon/runners`, `internal/mcp` drop `time.Sleep` waits and polling loops, replaced with direct row-visibility assertions. Comments updated to match.
- No schema changes, no API changes, no frontend changes. Acceptance criteria 1–7 satisfied; criterion 8 satisfied implicitly via the rewritten tests.

CI: `test` job green on 94afc529. `mergeable_state: clean`.

### Summary
Small, focused refactor that converts three async persistence paths to synchronous, matching the model `RecordStep` already follows. SSE ordering is now deterministic and shutdown-safety follows from the absence of unowned writers. Findings are nits, not blockers.

PR #568, opened 2026-05-15
refactor: remove defensive migration table creates

eloylp:
## Summary

- Removes defensive `CREATE TABLE IF NOT EXISTS` blocks from later migrations when the table is owned by an earlier migration.
- Keeps migrations strict: missing earlier tables now fail loudly instead of being silently recreated with compatibility schemas.
- Updates the pre-021 migration test fixture so it seeds the earlier tables it marks as already applied.

## Rationale

Later migrations should not hide drifted or partially-applied database state. A migration that rebuilds or alters an existing table should assume prior migrations ran successfully. If that assumption is false, startup should fail with a clear migration error rather than continuing on a partial schema.

## Verification

- `GOCACHE=/tmp/go-build-agents go test ./internal/store`
- `TMPDIR=/tmp GOCACHE=/tmp/go-build-agents go test ./internal/store -run TestOpenAndMigrate -count=1 -v`
- `git diff --check`

## Notes

This PR intentionally does not touch PR #565's new `030_trace_error_metadata.sql` because that PR is still open. A maintainer comment was added there asking the author to remove the same defensive `traces` table creation from that migration.

### DIFF
--- a/internal/store/migrations/022_workspace_memory.sql
+++ b/internal/store/migrations/022_workspace_memory.sql
@@ -1,16 +1,6 @@
 -- Scope daemon-managed agent memory by workspace. Existing memory rows belong
 -- to the migrated Default workspace.
 
--- Keep this compatibility definition aligned with 013_memory_store.sql; fresh
--- installs may reach this migration without a pre-existing memory table.
-CREATE TABLE IF NOT EXISTS memory (
-    agent      TEXT NOT NULL REFERENCES agents(name) ON DELETE CASCADE,
-    repo       TEXT NOT NULL,
-    content    TEXT NOT NULL DEFAULT '',
-    updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-    PRIMARY KEY (agent, repo)
-);
-
 CREATE TABLE memory_new (
     workspace_id TEXT NOT NULL DEFAULT 'default' REFERENCES workspaces(id) ON DELETE CASCADE,
     agent        TEXT NOT NULL REFERENCES agents(name) ON DELETE CASCADE,
--- a/internal/store/migrations/023_workspace_graph_layouts.sql
+++ b/internal/store/migrations/023_workspace_graph_layouts.sql
@@ -1,16 +1,5 @@
 -- Scope existing agent graph layout rows to the migrated Default workspace.
 -- The table already stores a generic scope column from 020_agent_ids_and_graph_layouts.sql.
-CREATE TABLE IF NOT EXISTS graph_layouts (
-    id         INTEGER PRIMARY KEY AUTOINCREMENT,
-    scope      TEXT NOT NULL DEFAULT 'global',
-    node_kind  TEXT NOT NULL,
-    node_id    TEXT NOT NULL REFERENCES agents(id) ON DELETE CASCADE,
-    x          REAL NOT NULL,
-    y          REAL NOT NULL,
-    updated_at TEXT NOT NULL DEFAULT (datetime('now')),
-    UNIQUE(scope, node_kind, node_id)
-);
-
 UPDATE graph_layouts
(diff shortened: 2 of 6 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

None blocking.

- **[info]** Directly authorized by maintainer steering. This is the separate cleanup the maintainer announced on #565 (comment 4463374992): *"We are going to clean up similar defensive migration scaffolding separately. The rule should be: migrations can assume earlier migrations ran successfully; no compatibility table creation for drifted/artificial schema states unless there is a documented real production upgrade path."* The PR correctly leaves #565's `030_trace_error_metadata.sql` untouched (that PR's own scope; its author already removed the same block under direct steering).

- **[info]** Ownership trace is sound. Each removed defensive block targets a table owned by an earlier migration, confirmed by each migration's own header comments:
  - `022` — `memory` owned by `013_memory_store.sql`.
  - `023` — `graph_layouts` owned by `020_agent_ids_and_graph_layouts.sql`.
  - `024` — `traces`/`events`/`event_queue`/`dispatch_history`/`token_budgets` owned by earlier migrations; `024` only renames-then-rebuilds them, so the `IF NOT EXISTS` creates were dead in the normal path.
  - `027` — `skills` owned by migration `001`.
  - `028` — `config` owned by an earlier migration; `028` only `INSERT OR IGNORE`s into it.

- **[info]** Behavior-preserving on the normal path, intentional change on drift. In a clean forward migration the removed blocks were no-ops (the table already existed), so the upgrade path is unchanged. The deliberate change is that a *drifted/partial* schema now fails loudly at `ALTER`/`RENAME`/`INSERT` instead of being silently rebuilt with a compatibility schema — exactly the maintainer's stated rule.

- **[info]** Correctness is covered by CI, not just assertion. The PR's `TestOpenAndMigrate` run (canonical full migration on an empty DB) is green on `287b4f1f`: if any now-assumed table were *not* owned by an earlier migration, the fresh path would fail with `no such table`. The full `./internal/store` suite being green also rules out regressions in other snapshot-seeding tests.

- **[info]** `024` index-name safety is preserved. The updated fixture seeds `idx_event_queue_pending` on the pre-024 `event_queue`. Migration `024` renames `event_queue` → `event_queue_legacy_workspace_024`, then `DROP TABLE event_queue_legacy_workspace_024` (dropping the legacy index, since SQLite index names are schema-global) *before* `CREATE INDEX idx_event_queue_pending` on the new table — so no name collision. This ordering is unchanged by this PR; the fixture now just mirrors real production state more faithfully.

- **[nit]** `internal/store/store_test.go` — the hand-rolled pre-migration fixture (e.g. `token_budgets` now seeded with `UNIQUE(scope_kind, scope_name, period)`, which the *removed* 024 defensive block lacked) is closer to the real owning migration's schema than the old defensive block was — good direction. The residual risk is the same one accepted on #565: a hand-written snapshot can drift from the real migration output over time. Harmless here (`RENAME`/`ALTER` only need the table to exist) and consistent with the project's accepted fixture style; not worth a rev.

### Notes

- **Merge coordination (heads-up for the human):** #565 is still open and also edits migration-fixture seeding in `internal/store/store_test.go` (its `TestAuthAdminMigrationBackfillsFirstExistingUser` pre-030 `traces` seed). This PR edits `TestWorkspacePromptMigrationBackfillsExistingAgents` in the same file. Different functions, but expect a likely merge conflict in `store_test.go` whichever lands second — a mechanical rebase, not a design issue.

### Summary
Correct, maintainer-authorized cleanup: removes dead compatibility scaffolding so later migrations fail loudly on drifted state, with ownership verified by the green full-migration test and the fixture updated to seed the earlier-owned tables. Not auto-merged — this changes the DB migration/startup path across multiple migration files with an intentional failure-semantics change, so it warrants a human merge decision rather than the contained-refactor fast path.

PR #572, opened 2026-05-16
refactor: use max for daemon runner offset clamp

eloylp:
## Summary

- Replaces the manual negative offset clamp in the daemon runners list handler with Go's built-in `max`.
- Keeps the existing pagination defaulting and response behavior unchanged.

## Verification

- `GOCACHE=/workspace/agents/.gocache go test ./internal/daemon/runners`
- `git diff --check`

### DIFF
--- a/internal/daemon/runners/runners.go
+++ b/internal/daemon/runners/runners.go
@@ -148,9 +148,7 @@ func (h *Handler) List(workspace, status string, limit, offset int) (ListRespons
 	if limit <= 0 {
 		limit = 100
 	}
-	if offset < 0 {
-		offset = 0
-	}
+	offset = max(offset, 0)
 	rows := make([]RunnerRow, 0, len(events))
 	for _, ev := range events {
 		rows = append(rows, h.expand(ev)...)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/daemon/runners/runners.go:151` — `offset = max(offset, 0)` is exactly equivalent to the removed `if offset < 0 { offset = 0 }` clamp. `max` is a Go builtin since 1.21 and the module is `go 1.25.5`, so it is fully supported. Pagination behavior is preserved.
- **[nit, pre-existing]** The clamp runs *after* `h.store.ListWorkspaceRunners(workspace, st, limit, offset)` is called with the raw offset, so it only sanitizes the echoed `Offset` response field, not the store query. This ordering is unchanged by this PR and the PR's "response behavior unchanged" claim is accurate — flagging only for awareness, not as a blocker.
- **[suggestion, pre-existing]** `runners_test.go` has no case exercising limit/offset clamping. The refactor is trivially equivalent so this is non-blocking, but a small table-driven case on `List` (negative offset, zero limit) would lock the contract in.

### Summary
Correct, idiomatic, behavior-preserving one-line refactor; contained and CI-green.

PR #574, opened 2026-05-16
refactor: use slices.Contains for public routes

eloylp:
## Summary

- Replaces repeated public-route equality checks in daemon auth with `slices.Contains`.
- Keeps the existing status, webhook, auth, UI, and loopback proxy route behavior unchanged.

## Verification

- `GOCACHE=/workspace/agents/.gocache go test ./internal/daemon`
- `git diff --check`

### DIFF
--- a/internal/daemon/auth.go
+++ b/internal/daemon/auth.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/daemon/auth.go:96-101` — The three `slices.Contains` conversions are faithful and behavior-preserving. Each replaces an `||` chain over an exact-match set with no change to ordering or short-circuit semantics: the proxy branch still evaluates `d.proxy != nil` first, and the UI branch still ORs `strings.HasPrefix(path, "/ui/")` exactly as before. `slices` (Go 1.21+ stdlib) is the idiomatic choice here.
- **[nit]** `internal/daemon/auth.go:99` — Pre-existing, not introduced by this PR: `"/ui/"` in the set is redundant with the adjacent `strings.HasPrefix(path, "/ui/")`. Out of scope for a behavior-preserving refactor; flagging only so it isn't lost.
- **[nit]** `isPublicRoute` runs on every HTTP request, and each call now builds three `[]string{...}` literals. Escape analysis should keep these stack-allocated (the slices don't escape `slices.Contains`), so this is not a real cost — noting only that the original `||` chains were equally readable and marginally cheaper. Pure taste; not blocking.

### Summary
Clean, faithful, behavior-preserving refactor of the daemon public-route check; correct and idiomatic, CI green. APPROVE. Not auto-merging: this is the auth bypass decision (`isPublicRoute`), so it falls under the auth/security exclusion and is left for a human merge decision.

PR #577, opened 2026-05-16
refactor: name trace step limit

eloylp:
## Summary

- Adds a single named constant for the trace step persistence cap in `internal/observe`.
- Reuses the constant in both single-step and batch step recording paths.

## Verification

- `GOCACHE=/workspace/agents/.gocache go test ./internal/observe`
- `git diff --check`

### DIFF
--- a/internal/observe/observe.go
+++ b/internal/observe/observe.go
@@ -23,6 +23,8 @@ import (
 	"github.com/eloylp/agents/internal/workflow"
 )
 
+const maxTraceSteps = 100
+
 // ─── Types ───────────────────────────────────────────────────────────────────
 
 // TimestampedEvent is a workflow event with an ingestion timestamp attached.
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[nit, non-blocking]** `internal/observe/observe.go:750` & `:780` — the godoc comments still say "capped at 100 per span" with the literal value. Accurate while `maxTraceSteps = 100`, but they will silently go stale if the constant ever changes. Optional: phrase as "capped per span (see `maxTraceSteps`)". Out of scope for a pure naming refactor; not a merge blocker.
- **[context, pre-existing]** The two cap checks have different semantics — `RecordStep` (`:761`) caps on the DB `COUNT(*)` for the span (true per-span total), while `RecordSteps` (`:795`) caps on the slice index `i` (per-call). Sharing one constant is the right call since both express the same intent ("max trace steps persisted per span"), and this divergence predates the PR. Noted only so the shared constant isn't mistaken for unifying the two behaviors.

### Summary
Clean, behavior-preserving extraction of a magic number into an unexported package-level constant; `100 → maxTraceSteps` substitution is numerically identical in both call sites, refactor is complete (no missed literals), and no public surface is widened. CI green.

PR #580, opened 2026-05-17
refactor: clone graph edge maps

eloylp:
## Summary

- Replaces manual configured-edge map copy loops with `maps.Clone` in the HTTP graph handler and MCP graph tool.
- Keeps the existing observed-edge overlay and response sorting behavior unchanged.

## Verification

- `GOCACHE=/workspace/agents/.gocache go test ./internal/daemon/observe ./internal/mcp`
- `git diff --check`

### DIFF
--- a/internal/daemon/observe/observe.go
+++ b/internal/daemon/observe/observe.go
@@ -15,6 +15,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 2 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/daemon/observe/observe.go:422`, `internal/mcp/tools_observe.go:172` — `maps.Clone` returns `nil` for a nil input, and both call sites write to `edgeByKey` afterwards (a nil-map write would panic). Verified safe: `configuredEdges` is unconditionally `make(map[string]...)` at `observe.go:385` and `tools_observe.go:139` and never reassigned to nil, so `maps.Clone` always returns a non-nil map. In `tools_observe.go` a `ReadAgents` error returns early before the clone, so the map is always built first.
- **[nit]** The original `make(..., len(edges)+len(configuredEdges))` pre-sized for both configured and observed edges; `maps.Clone` sizes only for `configuredEdges`, so the observed-edge overlay may trigger one or more map growths. Negligible for realistic graph sizes and acceptable for a simplification refactor — not blocking.

### Summary
Behavior-preserving simplification; the only correctness risk (nil-map write) does not apply because both maps are unconditionally constructed. CI green. Approving.

PR #582, opened 2026-05-17
refactor: use bytes clone for runner line capture

eloylp:
## Summary

- Replaces the manual `append([]byte(nil), line...)` byte-slice clone in the container runner test callback with `bytes.Clone`.
- Keeps the captured line isolation behavior unchanged while using the stdlib helper already imported by the test file.

## Verification

- `GOCACHE=/workspace/agents/.gocache go test ./internal/ai`
- `git diff --check`

### DIFF
--- a/internal/ai/cmdrunner_test.go
+++ b/internal/ai/cmdrunner_test.go
@@ -190,7 +190,7 @@ func TestContainerCommandRunnerUsesRuntimeAndParsesOutput(t *testing.T) {
 		System:   "system",
 		User:     "user",
 		OnLine: func(line []byte) {
-			lines = append(lines, append([]byte(nil), line...))
+			lines = append(lines, bytes.Clone(line))
 		},
 	})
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/ai/cmdrunner_test.go:193` — `bytes.Clone(line)` is equivalent to `append([]byte(nil), line...)` for this test's purpose (isolating each captured line from the runner's reused buffer). `bytes` is already imported (line 4); go.mod is `go 1.25.5` so the stdlib helper is available. The only theoretical difference — an empty *non-nil* input yields an empty non-nil slice from `bytes.Clone` vs `nil` from the old append — is immaterial here since lines are compared by content and the test asserts on non-empty parsed output.

### Summary
Clean, behavior-preserving test-only refactor; CI green. Approving.

PR #583, opened 2026-05-17
refactor: use cutprefix for runtime env lookup

eloylp:
## Summary

- Replaces the manual `HasPrefix` plus `TrimPrefix` environment lookup in runtime backend setup with `strings.CutPrefix`.
- Keeps the same first-match and empty-string behavior while using the standard helper for this pattern.

## Verification

- `git diff --check`

### DIFF
--- a/internal/runtime/backend_setup.go
+++ b/internal/runtime/backend_setup.go
@@ -148,8 +148,8 @@ func setEnvValues(env []string, kvs ...string) []string {
 func getEnv(env []string, key string) string {
 	prefix := key + "="
 	for _, entry := range env {
-		if strings.HasPrefix(entry, prefix) {
-			return strings.TrimPrefix(entry, prefix)
+		if value, ok := strings.CutPrefix(entry, prefix); ok {
+			return value
 		}
 	}
 	return ""


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/runtime/backend_setup.go:151` — Behavior is exactly preserved. `strings.CutPrefix(entry, prefix)` returns `(entry[len(prefix):], true)` iff `entry` starts with `prefix`, else `(entry, false)`. The old code only called `strings.TrimPrefix` *after* `strings.HasPrefix` confirmed the prefix, so it always returned `entry[len(prefix):]` — identical to the `value` returned here on `ok`.
- **[info]** First-match semantics unchanged: both versions iterate `env` in order and return on the first entry whose prefix matches.
- **[info]** Empty-string handling unchanged: for an entry like `"KEY="`, old `TrimPrefix` → `""`; new `CutPrefix` → `("", true)` → `""`. No match anywhere → `""` return in both.
- **[info]** `strings` import remains in use elsewhere in the file (`HasPrefix`, `Cut`), so no orphaned import. `strings.CutPrefix` (Go 1.20+) is available under `go 1.25.5`.

### Summary
Clean, idiomatic, strictly behavior-preserving one-function refactor; CI green. Not auto-merging because `getEnv` sits in the runtime backend-setup path and feeds credential/token gating (e.g. the `GITHUB_TOKEN` check for the claude backend) — that puts it outside the contained non-runtime/non-auth class, so the merge decision is left to a human.

PR #584, opened 2026-05-17
refactor: use cutprefix for webhook signature

eloylp:
## Summary

- Replaces the manual `TrimPrefix` signature prefix strip in webhook HMAC verification with `strings.CutPrefix`.
- Keeps the existing behavior for already-trimmed signatures and missing prefixes while using the clearer standard helper.

## Verification

- `go test ./internal/webhook`
- `git diff --check`

### DIFF
--- a/internal/webhook/handler.go
+++ b/internal/webhook/handler.go
@@ -485,7 +485,7 @@ func (h *Handler) enqueue(ctx context.Context, w http.ResponseWriter, ev workflo
 // hmac.Equal is used for the final comparison to avoid timing attacks that
 // could leak information about the expected value through execution time.
 func verifySignature(payload []byte, secret, signature string) bool {
-	signature = strings.TrimPrefix(strings.TrimSpace(signature), "sha256=")
+	signature, _ = strings.CutPrefix(strings.TrimSpace(signature), "sha256=")
 	if signature == "" || secret == "" {
 		return false
 	}


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/webhook/handler.go:488` — `strings.CutPrefix(s, "sha256=")` returns a string identical to the previous `strings.TrimPrefix(s, "sha256=")` for every input; the only difference is the reported `found` boolean, which is intentionally discarded via `_`. Since the original code never branched on prefix presence (it trimmed unconditionally), behavior is exactly preserved for: prefixed signatures, already-trimmed signatures, and the empty-string case after `TrimSpace`.
- **[info]** Security posture unchanged — constant-time `hmac.Equal` final comparison and the fail-closed `signature == "" || secret == ""` guard are both untouched. No new inputs or branches are introduced.
- **[info]** `strings.CutPrefix` requires Go 1.20+; `go.mod` declares `go 1.25.5`, so this compiles cleanly. `strings` was already imported (`TrimSpace` remains), so no import churn. CI `test` job is green.
- **[nit, optional]** `TestVerifySignature` exercises the prefixed-signature path but not the no-prefix passthrough or whitespace-trim path. This is pre-existing coverage, not a regression from this PR; a small table-driven addition (no-prefix, leading/trailing space) would lock the equivalence in. Not blocking.

### Summary
Clean, idiomatic, behavior-preserving one-line refactor with no security regression — APPROVE. Left for a human merge decision since the change lives in webhook HMAC signature verification (security-relevant code), which is outside the contained auto-merge class.

PR #591, opened 2026-05-18
refactor: clarify mcp traversal component check

eloylp:
## Summary

- Replaces the manual boolean expression in the MCP traversal component helper with an explicit switch over the cleaned sentinel values.
- Keeps the same rejection behavior for `.` and `..` components while making the two accepted cases easier to scan.

## Verification

- `go test ./internal/mcp`
- `git diff --check`

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -1115,5 +1115,10 @@ func trimmedStringOptional(req mcpgo.CallToolRequest, key string) (string, bool)
 // canonicalised by ai.NormalizeToken downstream and cannot escape the store.
 func isTraversalComponent(s string) bool {
 	c := filepath.Clean(s)
-	return c == "." || c == ".."
+	switch c {
+	case ".", "..":
+		return true
+	default:
+		return false
+	}
 }


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/mcp/tools.go:1116` — `isTraversalComponent` rewrite is exactly behavior-preserving. A Go `switch c { case ".", "..": return true; default: return false }` matches iff `c == "." || c == ".."`, so the new form is a 1:1 equivalent of the prior boolean expression. `filepath.Clean(s)` is still evaluated exactly once before the comparison, so path-traversal canonicalisation is unchanged.
- **[info]** Contract and signature unchanged; the sole call site (`internal/mcp/tools_observe.go`, the `agent`/`repo` traversal guard returning `"invalid agent or repo path"`) is unaffected — `.`/`..` still reject, everything else still passes through to `ai.NormalizeToken` downstream.
- **[info]** No new imports (`filepath` already in use); diff is body-only, +6/-1, single file. `go test ./internal/mcp` and the full suite pass; CI green.

### Summary
Clean, contained, behavior-preserving readability refactor of a path-traversal sentinel check — rejects exactly `.` and `..` as before. Approving; leaving the merge decision to a human since this is security-adjacent code.

PR #592, opened 2026-05-18
refactor: parse first model field lazily

eloylp:
## Summary

- Replaces the plain-text model parser's `strings.Fields` allocation with `strings.FieldsSeq` when only the first field is needed.
- Keeps the same first-token extraction behavior for fallback model discovery output.

## Verification

- `git diff --check`

### DIFF
--- a/internal/backends/discovery.go
+++ b/internal/backends/discovery.go
@@ -582,11 +582,10 @@ func parseModels(raw string) []string {
 		if strings.HasPrefix(lower, "model") || strings.HasPrefix(lower, "available") || strings.HasPrefix(lower, "current") || strings.HasPrefix(lower, "you") || strings.HasPrefix(lower, "for ") {
 			continue
 		}
-		fields := strings.Fields(line)
-		if len(fields) == 0 {
-			continue
+		for field := range strings.FieldsSeq(line) {
+			names = append(names, field)
+			break
 		}
-		names = append(names, fields[0])
 	}
 	return dedupeSorted(names)
 }


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- No correctness, concurrency, security, or style issues found.
- Behavioral equivalence verified for the three relevant cases in the plain-text fallback at `internal/backends/discovery.go:585-588`:
  - **Empty / whitespace-only line:** old `strings.Fields` returns an empty slice → `continue`; new `strings.FieldsSeq` yields zero elements → loop body never executes → nothing appended. Identical. (This case is also defensive — `line` is already `TrimSpace`d and `line == ""` is filtered above.)
  - **Single-field line:** both append that single token.
  - **Multi-field line:** old appends `fields[0]`; new appends the first yielded field then `break` cleanly terminates the iterator before any further field. Identical first-token semantics.
- `strings.FieldsSeq` splits on runs of whitespace via `unicode.IsSpace`, exactly as `strings.Fields`; ordering is preserved. It is available (Go 1.24+; repo targets `go 1.25.5`).
- `break` in a range-over-func loop cleanly stops the iterator; `FieldsSeq` is a pure in-memory iterator with no resource/leak concern.
- No new import added (`strings` is already used in the same function). Consistent with the adjacent `strings.SplitSeq` usage in `parseModels`.

### Summary
Pure, behavior-preserving Go 1.24 `FieldsSeq` refactor that avoids the full-slice allocation when only the first field is needed; first-token extraction is unchanged for empty and multi-field output. CI green.

PR #597, opened 2026-05-18
refactor: simplify mcp graph edge sort

eloylp:
## Summary

- Replaces the manual two-key MCP graph edge comparator with `cmp.Or` and `cmp.Compare`.
- Keeps the same `from` then `to` ordering while matching the concise comparator style already used elsewhere.

## Verification

- `env -u GH_TOKEN -u GITHUB_TOKEN -u CLAUDE_CODE_OAUTH_TOKEN -u CODEX_AUTH_JSON_BASE64 -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u ANTHROPIC_AUTH_TOKEN go test ./internal/mcp`
- `git diff --check`

### DIFF
--- a/internal/mcp/tools_observe.go
+++ b/internal/mcp/tools_observe.go
@@ -195,10 +195,7 @@ func toolGetGraph(deps Deps) server.ToolHandlerFunc {
 			wireEdges = append(wireEdges, e)
 		}
 		slices.SortFunc(wireEdges, func(a, b mcpGraphEdge) int {
-			if byFrom := cmp.Compare(a.From, b.From); byFrom != 0 {
-				return byFrom
-			}
-			return cmp.Compare(a.To, b.To)
+			return cmp.Or(cmp.Compare(a.From, b.From), cmp.Compare(a.To, b.To))
 		})
 
 		return jsonResult(map[string]any{


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[none]** `internal/mcp/tools_observe.go:198` — Behavior-preserving. `cmp.Or(x, y)` returns the first non-zero argument, else the last; with `x = cmp.Compare(a.From, b.From)` and `y = cmp.Compare(a.To, b.To)` this is exactly the original "return byFrom if non-zero, else compare To" chain. Primary `From` / secondary `To` ascending ordering is preserved; `slices.SortFunc` ordering is unaffected since the comparator output is identical in every case.
- **[info]** `cmp` is already imported — no new import is introduced, and `cmp.Or`/`cmp.Compare` are both Go 1.21+ (the module already uses `cmp.Compare`/`slices.SortFunc`).
- **[nit]** `cmp.Or` evaluates both arguments eagerly (it is not short-circuiting). Since `cmp.Compare` on strings is pure and side-effect-free, this is a negligible, non-observable extra string comparison on the `From`-differs path — an accepted, idiomatic tradeoff for the more concise multi-key comparator (this is the canonical `cmp.Or` use case in the stdlib docs). The style is appropriate for this package.

### Summary
Correct, idiomatic, behavior-preserving comparator simplification; CI green and tests pass.

PR #607, opened 2026-05-18
fix: add graph prompt editing and transcripts

### LINKED ISSUE #600: Make graph editor support inline prompt editing and live trace transcript access
## Goal

Make the graph section more central as the primary workflow design and operations surface.

The graph already lets users manage agents/dispatch wiring and inspect recent activity. Two convenience gaps remain:

1. Users should be able to edit an agent's selected prompt directly from the graph editor.
2. Users should be able to open the live trace transcript from graph-side recent activity, not only from the dedicated traces/runners surfaces.

## Product context

Prompts are correctly modeled as reusable catalog entries. That should not change.

However, when a user is editing an agent from the graph, it is awkward to leave the graph, navigate to the prompt catalog, find the referenced prompt, edit it, and then return. The graph should feel like the central "agent studio" surface, so editing the prompt in place should be possible as a convenience layer over the existing prompt catalog/backend logic.

Similarly, the graph editor already shows latest activity for an agent and can lead users into traces. It should also expose the live transcript modal from that same context, so users can inspect an in-flight or recent run without changing mental context.

### Inline prompt editing from graph agent editor

- In the graph agent editor, show the currently selected prompt reference/id for the selected agent.
- Provide an edit-in-place affordance for the referenced prompt content.
- Reuse the existing frontend/backend prompt update logic. Do not introduce a second prompt persistence path.
- Preserve the catalog model: the agent still references a prompt by stable prompt id; editing in place updates that prompt catalog entry.
- Make it clear in the UI that editing the prompt affects every agent using the same prompt.
- If the agent has no resolvable prompt reference, surface a clear user-actionable error instead of silently creating unrelated state.
- Respect existing scoped prompt visibility rules. The editor must not allow selecting or editing prompts outside the agent's visible catalog scope.
- Keep validation centralized through the existing REST/MCP/service/store paths. Do not duplicate domain validation only in the UI.

### Live transcript access from graph activity

- From the graph-side latest activity / recent runs area for an agent, allow opening the same live transcript experience used elsewhere in the app.
- If the run is still active, the transcript should stream live using the existing trace steps stream endpoint.
- If the run is completed, the transcript should replay persisted trace steps from the database, matching the existing trace view behavior.
- Reuse the existing transcript modal/components and trace step parsing/rendering logic where possible.
- Do not create a second SSE implementation or a separate in-memory transcript path.
- Preserve auth behavior and existing error handling for protected trace endpoints.

## Implementation guidance

- Prefer component extraction/reuse over copying UI logic from the traces/runners pages.
- Keep the graph editor as the orchestration surface, but keep prompt data ownership in the prompt catalog API/service/store path.
- If existing components are too coupled to their current page, first extract small reusable components rather than introducing a large rewrite.
- Avoid changing the prompt catalog data model unless a concrete blocker is found.
- Avoid introducing new backend endpoints unless existing prompt and trace endpoints cannot support the flow cleanly.

## Acceptance criteria

- A user can select an agent node in the graph and edit the content of the prompt currently referenced by that agent.
- Saving from the graph updates the same prompt catalog entry visible in the prompts/config surfaces.
- The UI warns or clearly indicates that editing the prompt may affect other agents sharing it.
- A user can open a trace transcript from graph-side recent activity.
- Active traces stream live; completed traces render from persisted trace steps.
- Existing prompt and trace tests are updated or new focused tests are added for the graph integration.
- README / AGENTS.md / CLAUDE.md / relevant UI docs are updated if the user-facing graph workflow changes.

eloylp:
## Summary

- Adds a graph Prompt tab that resolves the selected agent's visible catalog prompt and edits it through the existing prompt PATCH path.
- Reuses the live trace transcript modal as a shared component and opens it from graph activity rows for active or completed spans.
- Updates the dashboard UI docs for graph prompt editing and transcript access.

Closes #600

## Verification

- `npm test`
- `npm run build`
- `git diff --check`

### DIFF
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -29,7 +29,9 @@ Primary workflow designer showing agents as draggable graph nodes and dispatch p
 
 The graph deliberately keeps one node per real agent. Repo-scoped agents are grouped inside thin dashed repo boundaries; workspace-scoped agents sit outside those boundaries. Repo trigger bindings are visualised as thin lines from the agent to passive repo anchors, so the user can see where an agent runs without duplicating the agent into one node per repo. Dispatch wiring remains the editable agent-to-agent relationship.
 
-The right-side **Agent editor** is the graph's main editing surface. Click an agent to inspect or edit its definition, run it against bound repos, manage repo triggers, add or remove dispatch wiring, and review recent runner rows / trace links. The editor is tabbed by concept: Overview, Settings, Triggers, Dispatch, and Activity. Operators can create agents without leaving the designer.
+The right-side **Agent editor** is the graph's main editing surface. Click an agent to inspect or edit its definition, run it against bound repos, manage repo triggers, add or remove dispatch wiring, edit the referenced catalog prompt, and review recent runner rows / trace links. The editor is tabbed by concept: Overview, Settings, Prompt, Triggers, Dispatch, and Activity. The Prompt tab updates the same reusable prompt catalog entry selected by the agent and calls out that shared ownership before saving. Operators can create agents without leaving the designer.
+
+The Activity tab can open the same live transcript modal used by the Runners page. In-flight spans stream from `/traces/{span_id}/stream`; completed spans replay the persisted `trace_steps` rows through the same surface, with a link back to the full trace detail.
 
 Dispatch wiring is always editable: drag from one agent to another to add a connection, or click an edge to inspect its Overview, History, and Danger tabs. Dispatch changes write back to the source agent's `can_dispatch` list and the target's `allow_dispatch` flag.
 
--- a/internal/ui/src/app/graph/page.tsx
+++ b/internal/ui/src/app/graph/page.tsx
@@ -24,11 +24,14 @@ import dagre from 'dagre'
 import Card from '@/components/Card'
 import AgentForm, { emptyAgentForm, type BackendOption } from '@/components/AgentForm'
 import BadgePicker from '@/components/BadgePicker'
+import LiveTraceModal, { type LiveTraceSpan } from '@/components/LiveTraceModal'
+import MarkdownEditor from '@/components/MarkdownEditor'
 import RunButton from '@/components/RunButton'
 import WorkspaceSelect from '@/components/WorkspaceSelect'
 import { useSelectedWorkspace, withWorkspace, type CatalogItem } from '@/lib/workspace'
 import { type Binding } from '@/lib/bindings'
 import { fmtDuration } from '@/lib/format'
+import { graphPromptIdentifier, resolveGraphPrompt, type GraphPromptItem } from '@/lib/graph-prompt'
 import {
   addCanDispatch,
   availableDispatchTargets,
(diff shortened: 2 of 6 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Issue #600 goal check
Inline prompt editing: prompt id/scope shown; in-place `MarkdownEditor` + save; reuses the exact catalog PATCH path; catalog model preserved; shared-ownership warning present; clear actionable error when no visible prompt resolves; scoped visibility respected; backend validation not duplicated. Live transcript: opens from graph activity rows; live stream + completed replay via the existing endpoint; modal/components reused; no second SSE path; auth/error handling preserved. Docs updated. All acceptance criteria met.

### Findings
- **[info]** `internal/ui/src/lib/graph-prompt.ts:11` — Prompt scope/visibility is **correct**. `resolveGraphPrompt` filters candidates through `visibleCatalogItems(prompts, workspace, repo)` with `repo` derived as `scope_type === 'repo' ? scope_repo : ''`, identical to the existing `AgentForm` picker (`AgentForm.tsx:64-65`). Out-of-scope (cross-repo) prompts resolve to `null`; resolution matches stable `prompt_id` first, then legacy `prompt_ref`/`prompt_scope`. The editor cannot select or edit a prompt outside the agent's visible catalog scope. `graph-prompt.test.ts` covers the in-scope, cross-repo-rejection, and legacy-ref cases.
- **[info]** `internal/daemon/observe/observe.go:225` — Completed-span replay via `/traces/{span_id}/stream` is **appropriate** from the graph activity surface. The endpoint replays persisted `trace_steps` rows then emits `event: end` for inactive spans (`!active && len(steps) > 0`), matching the existing trace view / Runners behavior. The modal is a faithful extraction of the former `LiveStreamModal` (same `openAuthenticatedSSE` path, same `StreamCard`/`stepToCardEntries`/`TranscriptFilter`), so no second SSE or in-memory transcript path is introduced.
- **[positive]** `internal/ui/src/components/LiveTraceModal.tsx:14` — The extracted modal's `traceHref` now uses `span.rootEventId || span.id`. The trace-detail page filters by `root_event_id` (`traces/page.tsx:238`), so the previous Runners link (`/traces/?id=<span.id>`) was a latent broken link for the common `span_id != root_event_id` case; this extraction quietly fixes it and both call sites pass `rootEventId: event_id`.
- **[low]** `internal/ui/src/app/graph/page.tsx:1606` — The "Open transcript" button renders for any row with `span_id` regardless of status. A completed span that recorded no `trace_steps` (non-claude backend / pre-migration) makes `/traces/{span_id}/stream` return 404, and the reused modal shows "Lost connection to the live stream. The run may still be in flight…", which reads oddly for a finished run. This is **inherited** behavior of the reused component (same on the Runners page) and the issue explicitly mandates reuse + matching existing behavior, so it is non-blocking — flagging only as a future polish candidate.
- **[nit]** No test asserts the graph transcript button wiring or the `rootEventId || span.id` href selection. The highest-risk logic (prompt resolution/scoping) is well covered and the modal change is a behavior-preserving extraction validated by the full green suite, so this is a minor coverage observation, not a blocker.

### Verification
- Cloned at `38f88d7`; `npm ci` + `npx vitest run` → **53/53 green** (incl. new `graph-prompt.test.ts` 3/3); `tsc --noEmit` clean. CI `test` job green.
- Confirmed `GET /prompts` returns full `content`/`description` (`storePromptJSON`, `content` has no `omitempty`), so the graph editor loads real content — no data-loss/overwrite risk. PATCH body shape and id resolution match the prompts catalog page exactly.

### Summary
Clean, well-scoped feature: faithful modal extraction, reused backend prompt/trace paths, correct prompt-scope enforcement, well-tested resolution logic. Prompt visibility and completed-span replay are both correct. APPROVE; only a non-blocking inherited-UX note and a minor coverage nit.

PR #608, opened 2026-05-19
refactor: simplify bcrypt prefix detection

eloylp:
## Summary

- Replaces the manual bcrypt hash prefix OR-chain with `slices.ContainsFunc` over the supported bcrypt prefixes.
- Keeps the same bcrypt verification path and PBKDF2 fallback behavior.

## Verification

- `go test ./internal/store`
- `git diff --check`

### DIFF
--- a/internal/store/auth.go
+++ b/internal/store/auth.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[positive]** `internal/store/auth.go:439` — The rewrite is an exact behavior-preserving equivalence. `slices.ContainsFunc([]string{"$2a$","$2b$","$2y$"}, func(p) { return strings.HasPrefix(encoded, p) })` is true iff `HasPrefix($2a$) || HasPrefix($2b$) || HasPrefix($2y$)`: the predicate is pure, elements are visited in slice order, and `ContainsFunc` short-circuits on the first match (it returns `IndexFunc(...) >= 0`, which breaks on first true). Same prefixes, same evaluation order, same short-circuit.
- **[positive]** Bcrypt-vs-PBKDF2 fallback is fully preserved (dispatch focus). The recognized bcrypt prefix set is unchanged, the `if` body (`bcrypt.CompareHashAndPassword(...) == nil, false`) is untouched, and the fall-through PBKDF2 path (`strings.Cut` + algorithm dispatch) is untouched. Any input routes to exactly the same branch as before.
- **[info]** No auth/security regression. The check operates on the hash *format identifier*, not on a secret; constant-time password comparison remains delegated to `bcrypt.CompareHashAndPassword`. New import is std-lib only (`slices`, available since Go 1.21; module is `go 1.25.5`), and the import block stays gofmt-ordered. CI `test` job is green.
- **[nit]** Taste only, non-blocking: the original three-term `||` chain was already clear and idiomatic, and it avoided constructing a 3-element slice + a closure capturing `encoded` on every `verifyPassword` call. The slice does not escape, so escape analysis likely keeps it off-heap and the cost is negligible — but `slices.ContainsFunc` is arguably more machinery for the same result here. This is consistent with the repo's ongoing std-lib-idiom refactor series, so it is a defensible direction; flagging only so the tradeoff is visible.

### Summary
Exact, behavior-preserving simplification of bcrypt prefix detection; bcrypt/PBKDF2 fallback and auth semantics fully preserved, CI green — APPROVE.

PR #614, opened 2026-05-19
refactor: use bytes trim for first byte scan

eloylp:
## Summary

- Replaces the manual ASCII whitespace scan in `firstNonSpace` with `bytes.TrimLeft` using the same whitespace set.
- Keeps the same zero return for all-whitespace input and the same first-byte result for non-empty JSON payloads.

## Verification

- `env -u GH_TOKEN -u GITHUB_TOKEN -u CLAUDE_CODE_OAUTH_TOKEN -u CODEX_AUTH_JSON_BASE64 -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u ANTHROPIC_AUTH_TOKEN go test ./internal/anthropic_proxy`
- `git diff --check`

### DIFF
--- a/internal/anthropic_proxy/translate.go
+++ b/internal/anthropic_proxy/translate.go
@@ -316,12 +316,11 @@ func parseContent(raw json.RawMessage) (blocks []ContentBlock, text string, err 
 }
 
 func firstNonSpace(b []byte) byte {
-	for _, c := range b {
-		if c != ' ' && c != '\t' && c != '\n' && c != '\r' {
-			return c
-		}
+	trimmed := bytes.TrimLeft(b, " \t\n\r")
+	if len(trimmed) == 0 {
+		return 0
 	}
-	return 0
+	return trimmed[0]
 }
 
 // toolResultContent extracts the string content from a tool_result block.


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/anthropic_proxy/translate.go:318` — `bytes.TrimLeft(b, " \t\n\r")` is exactly behavior-preserving here. The cutset is all-ASCII, so the stdlib takes the `makeASCIISet` fast path and trims leading bytes via a byte-by-byte scan against the same {`' '`,`'\t'`,`'\n'`,`'\r'`} set as the old explicit loop — no rune decoding affects the result. First non-whitespace byte returned is identical.
- **[info]** Zero-return preserved: empty input and all-whitespace input both yield `len(trimmed)==0 → return 0`, matching the old `return 0` fallthrough (`nil` input is also fine — `len(nil)==0`).
- **[info]** Edge case checked: a `0x00` byte as the first non-whitespace byte returns `0` in *both* implementations — the all-whitespace/null-byte sentinel ambiguity is pre-existing and unchanged, so this is not a regression.
- **[info]** No new import — `bytes` is already imported (`translate.go:4`). CI green (build + full `go test ./...` + targeted package test).

### Summary
Faithful, byte-for-byte equivalent refactor with whitespace handling and zero-return behavior preserved across empty, all-whitespace, null-byte, and non-ASCII inputs. APPROVE.

PR #622, opened 2026-05-19
feat: add persistent sidebar focus mode

### LINKED ISSUE #616: Add persistent sidebar collapse and graph focus mode
## Goal

Allow users to hide/collapse the application menu even on large screens, especially while working in the graph editor.

The graph editor should be able to feel like a focused design surface: graph canvas + editor tabs, with the menu available only when the user wants it.

## Product context

The graph section is becoming the central workflow designer. When editing agents and dispatch wiring, the left navigation can become visual noise even on desktop-sized screens. Users should be able to create a focus experience without losing access to navigation.

## Recommended approach

Implement persistent manual sidebar collapse globally, plus a graph-page shortcut for entering focus mode.

This avoids route-specific magic while still making the graph editor feel focused.

### Global persistent sidebar collapse

- The sidebar/menu can be collapsed on large screens as well as small screens.
- The collapsed/expanded preference persists locally, for example in `localStorage`.
- The user can always reopen the menu from a visible hamburger/menu button.
- The collapse behavior should not depend only on viewport size.
- Existing mobile behavior should continue working.
- Do not remove navigation items or change routes.

### Graph focus shortcut

- Add a graph-page affordance, for example a `Focus` button in the graph toolbar/header.
- Activating it collapses the sidebar/menu using the same global collapse state.
- The user can still reopen the menu through the normal hamburger/menu button.
- The graph canvas/editor should gain the reclaimed horizontal space.

### UX expectations

- On desktop, users can choose between:
  - full navigation visible
  - collapsed navigation / focus layout
- The menu should not auto-collapse just because the user visits `/graph`; avoid surprising route-specific behavior.
- The graph focus action is explicit user intent.
- The collapsed state should be stable across reloads and navigation until the user changes it.

## Implementation guidance

- Look first at `internal/ui/src/components/DashboardShell.tsx`, where the shell/sidebar/menu state is managed.
- The existing collapsed/mobile menu behavior should be reused rather than replaced.
- Avoid introducing a new global state library; local component state plus `localStorage` should be enough unless the current shell architecture requires otherwise.
- If the graph page needs to trigger collapse from inside the page, prefer a simple shared event/context callback rather than coupling the graph page tightly to shell internals.
- Keep styling consistent with the current square-corner direction.
- Reuse existing icon/button patterns; do not add a dependency just for the hamburger/focus icon.

## Acceptance criteria

- On a large desktop viewport, the user can collapse the sidebar/menu.
- The collapsed state persists across reloads.
- The user can reopen the menu from the collapsed state.
- The graph page exposes a clear focus action that collapses the menu.
- Graph/editor layout uses the freed space after collapse.
- Mobile navigation still works.
- No route/navigation behavior changes except the new collapse/focus affordance.
- UI tests are added or updated if existing tests cover shell/menu behavior.

eloylp:
## Summary

- Adds a persistent desktop sidebar collapse state in the dashboard shell, stored in `localStorage`.
- Keeps the hamburger control visible so users can collapse or reopen navigation, while preserving mobile menu behavior.
- Adds a graph toolbar `Focus` action that collapses the global sidebar so the graph/editor surface gains horizontal space.
- Restores the saved collapse preference after mount to avoid static-export hydration mismatch, and shares the shell-collapse event name between the shell and graph page.

Closes #616

## Verification

- `git diff --check`
- `npm test -- DashboardShell.test.tsx` (blocked: fresh checkout has no installed Node dependencies; `vitest` not found)
- `npm ci --offline` (blocked: npm cache is incomplete for `zwitch-2.0.4.tgz`; no network install attempted under run guardrails)
- `npm run build` (not run because Node dependencies are unavailable in this checkout)

### DIFF
--- a/internal/ui/src/app/graph/page.tsx
+++ b/internal/ui/src/app/graph/page.tsx
@@ -32,6 +32,7 @@ import { useSelectedWorkspace, withWorkspace, type CatalogItem } from '@/lib/wor
 import { type Binding } from '@/lib/bindings'
 import { fmtDuration } from '@/lib/format'
 import { graphPromptIdentifier, resolveGraphPrompt, type GraphPromptItem } from '@/lib/graph-prompt'
+import { SIDEBAR_COLLAPSE_EVENT } from '@/lib/shell-events'
 import {
   addCanDispatch,
   availableDispatchTargets,
@@ -1141,7 +1142,7 @@ export default function GraphPage() {
   }, [loadLookups, promptDraft, selectedPrompt])
 
   const enterFocusMode = useCallback(() => {
-    window.dispatchEvent(new Event('agents:shell-collapse-sidebar'))
+    window.dispatchEvent(new Event(SIDEBAR_COLLAPSE_EVENT))
   }, [])
 
   return (
--- a/internal/ui/src/app/graph/page.tsx
+++ b/internal/ui/src/app/graph/page.tsx
@@ -1140,6 +1140,10 @@ export default function GraphPage() {
     }
   }, [loadLookups, promptDraft, selectedPrompt])
 
+  const enterFocusMode = useCallback(() => {
+    window.dispatchEvent(new Event('agents:shell-collapse-sidebar'))
+  }, [])
+
   return (
     <div>
       <div style={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center', marginBottom: '1.5rem' }}>
@@ -1152,6 +1156,9 @@ export default function GraphPage() {
         <div style={{ display: 'flex', gap: '0.5rem', alignItems: 'center' }}>
           <WorkspaceSelect compact />
           <RepoFilter selected={repoFilter} onChange={setRepoFilter} workspace={workspace} />
+          <button onClick={enterFocusMode} style={{ background: 'var(--bg-card)', border: '1px solid var(--border)', color: 'var(--accent)', padding: '6px 12px', borderRadius: '6px', cursor: 'pointer', fontSize: '0.875rem', fontWeight: 500 }}>
+            Focus
+          </button>
           <button onClick={openCreateAgent} style={{ background: 'var(--btn-primary-bg)', border: '1px solid var(--btn-primary-border)', color: '#fff', padding: '6px 12px', borderRadius: '6px', cursor: 'pointer', fontSize: '0.875rem', fontWeight: 600 }}>
             + Create agent
           </button>
--- a/internal/ui/src/components/DashboardShell.test.tsx
+++ b/internal/ui/src/components/DashboardShell.test.tsx
@@ -1,5 +1,6 @@
-import { fireEvent, render, screen } from '@testing-library/react'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
 import { afterEach, describe, expect, it, vi } from 'vitest'
+import { SIDEBAR_COLLAPSE_EVENT } from '@/lib/shell-events'
 import DashboardShell from './DashboardShell'
 
 vi.mock('next/navigation', () => ({
@@ -65,9 +66,22 @@ describe('<DashboardShell />', () => {
     mockViewport(false)
     const { container } = render(<DashboardShell><div>Content</div></DashboardShell>)
 
-    fireEvent(window, new Event('agents:shell-collapse-sidebar'))
+    fireEvent(window, new Event(SIDEBAR_COLLAPSE_EVENT))
 
     expect(container.querySelector('.dashboard-shell.nav-collapsed')).toBeInTheDocument()
+    expect(window.localStorage.getItem('agents.sidebarCollapsed')).toBe('true')
+  })
+
+  it('restores collapsed desktop navigation after reload', async () => {
+    window.localStorage.setItem('agents.sidebarCollapsed', 'true')
+    mockShellFetch()
+    mockViewport(false)
+    const { container } = render(<DashboardShell><div>Content</div></DashboardShell>)
+
+    await waitFor(() => {
+      expect(container.querySelector('.dashboard-shell.nav-collapsed')).toBeInTheDocument()
+    })
+    expect(screen.getByTitle('Open navigation')).toBeInTheDocument()
     expect(window.localStorage.getItem('agents.sidebarCollapsed')).toBe('true')
   })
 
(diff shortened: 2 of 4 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

Reviewed against issue #616. The feature is functionally complete and well-tested: desktop collapse works independently of viewport, the hamburger stays visible and toggles both ways, the graph `Focus` button collapses the shell via a namespaced `window` event (loose coupling, exactly the "simple shared event/callback" the issue recommends — graph page knows nothing about shell internals), mobile drawer behavior is preserved, no nav items/routes change, and `.nav-collapsed` reclaims `margin-left` for the graph surface. The new `DashboardShell.test.tsx` cases are non-vacuous (verified: reverting only the production files to base makes 2/3 fail; the 3rd is a deliberate mobile-preservation guard that correctly passes on both). Full vitest suite green (57/57) at head; CI Go `test` job green.

One substantive issue blocks approval.

### Findings

- **[medium]** `internal/ui/src/components/DashboardShell.tsx:31-34` — the `useState` lazy initializer reads `localStorage` during render. `next.config.js` sets `output: 'export'`, so pages are prerendered to static HTML at build time with `window` undefined → `sidebarCollapsed=false` → `<div className="dashboard-shell ">`. A returning user who collapsed the sidebar hydrates with `localStorage` returning `'true'` → first client render produces `dashboard-shell nav-collapsed`, which does not match the static HTML. Result: a React hydration mismatch (console error) plus a visible flash of the full sidebar before it collapses, on **every reload** — precisely the reload path the issue's "collapsed state should be stable across reloads" expectation is about. The project already solves this exact problem in `@/lib/theme` `ThemeProvider` (the hook this component imports via `useTheme`): initialize to a stable default and read `localStorage` in a mount `useEffect`. Recommend matching that pattern, e.g. `useState(false)` plus `useEffect(() => { setSidebarCollapsed(window.localStorage.getItem(SIDEBAR_COLLAPSED_KEY) === 'true') }, [])`. The existing `[sidebarCollapsed]` write effect already self-corrects after the read effect, mirroring `ThemeProvider`.

- **[low]** `internal/ui/src/app/graph/page.tsx:1143` — the event name `'agents:shell-collapse-sidebar'` is a hardcoded string literal here, while `DashboardShell.tsx:8` defines `SIDEBAR_COLLAPSE_EVENT` for the same value. The two can drift silently and break the Focus button with no compile-time signal. Export the constant (or a tiny shared `collapseSidebar()` helper) from one module and use it on both sides. The event-based coupling itself is appropriately small and a good fit for the issue's guidance — only the duplicated literal is the concern.

- **[low]** `internal/ui/src/components/DashboardShell.test.tsx` — tests cover the event-driven collapse and the `localStorage` *write*, but not the *read-on-init / reload* path (render with `localStorage['agents.sidebarCollapsed']='true'` set before mount → expect `.nav-collapsed`). That path is the one most relevant to the persistence acceptance criterion and to the hydration concern above; a case there would lock the restored-on-reload behavior down.

### Summary
Feature meets all functional issue goals with solid, non-vacuous tests; requesting changes only for the SSR/hydration-mismatch + flash from the render-time `localStorage` read, which diverges from the project's established `ThemeProvider` pattern and undermines the "stable across reloads" UX expectation.

eloylp:
Addressed the requested changes in `8c9f8df`:

- moved the sidebar collapse preference read out of render and into a mount effect, with writes gated until the stored preference has been loaded;
- added a reload-persistence test for a pre-existing `agents.sidebarCollapsed=true` value;
- moved the shell-collapse event name into a shared UI module so the graph Focus button and shell listener cannot drift.

Verification: `go test ./...` and `git diff --check` pass. The targeted UI test remains blocked in this checkout because Node dependencies are not installed and `npm ci --offline` is missing `zwitch-2.0.4.tgz` from the cache.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

Re-review of `8c9f8df` against my prior REQUEST_CHANGES on `58dacbc` (issue #616). All three findings are addressed; full vitest suite green and non-vacuous; CI Go `test` job green.

### Prior findings — resolution

- **[medium → resolved]** `DashboardShell.tsx` render-time `localStorage` read / hydration mismatch. Now `useState(false)` (stable default matching the static-export prerendered HTML) + a mount `useEffect` that reads `localStorage`, with the write `useEffect` gated on a `sidebarPreferenceLoaded` flag. This is exactly the established `ThemeProvider` pattern. Effect ordering is correct: on first commit the write effect early-returns while `sidebarPreferenceLoaded` is still `false`, so the default `false` never clobbers a stored `true` before the read effect runs. Client first render now matches SSR output → no hydration mismatch, no reload flash.
- **[low → resolved]** Duplicated event-name literal. New `@/lib/shell-events.ts` exports `SIDEBAR_COLLAPSE_EVENT`; both the shell listener and the graph `Focus` dispatcher import it — drift is no longer possible.
- **[low → resolved]** Missing read-on-init/reload test. New `restores collapsed desktop navigation after reload` seeds `localStorage` before mount and asserts `.nav-collapsed` via `waitFor`.

### Verification

- Cloned at head `8c9f8df`, `npm install`, full `vitest run`: **12 files / 58 tests passed**.
- Non-vacuousness: reverting only the two production files (`DashboardShell.tsx`, `graph/page.tsx`) to base while keeping head tests + `shell-events.ts` makes the 3 behavioral cases fail (toggle-label, event-collapse persistence, reload-restore); the `keeps the hamburger opening mobile navigation` case correctly still passes — a deliberate mobile-preservation guard, not vacuous.
- `git diff --check` clean; CI Go `test` green.
- Issue #616 acceptance criteria all met: desktop collapse independent of viewport, persists across reloads (now hydration-safe), reopen from collapsed via always-visible hamburger, graph `Focus` action, reclaimed `margin-left` for the graph surface, mobile drawer preserved, no route changes, tests added.

### Findings

- **[low]** `DashboardShell.tsx` — on a mobile viewport, if the persisted `sidebarCollapsed` is `true` (set earlier on desktop, shared `localStorage` key), tapping the hamburger does `setSidebarCollapsed(false)` + `setSidebarOpen(true)`, i.e. opening the mobile drawer also clears the global desktop collapse preference. This is a defensible single-global-state design choice (explicit user interaction with the toggle) and breaks no acceptance criterion — flagging only for maintainer awareness, non-blocking.
- **[nit]** `lib/shell-events.ts` carries a `'use client'` directive while exporting only a string constant. Harmless (both consumers are already client components) but unnecessary; optional cleanup.

### Summary
Prior medium hydration defect and both low findings fully addressed using the project's own `ThemeProvider` precedent; tests green and proven non-vacuous — approving.

PR #628, opened 2026-05-20
refactor: centralize backend token fallback test setup

eloylp:
## Summary

- Adds a shared backend discovery test helper for the GITHUB_TOKEN to GH_TOKEN fallback setup.
- Applies the helper to both runtime backend tests so a host GH_TOKEN cannot shadow the intended fallback token.

## Verification

- `git diff --check`

### DIFF
--- a/internal/backends/discovery_test.go
+++ b/internal/backends/discovery_test.go
@@ -108,16 +108,7 @@ For building AI applications, default to the latest capable model.`
 }
 
 func TestDiagnoseGitHubCLIInRuntimeUsesRunnerContainer(t *testing.T) {
-	oldGH, hadGH := os.LookupEnv("GH_TOKEN")
-	t.Cleanup(func() {
-		if hadGH {
-			_ = os.Setenv("GH_TOKEN", oldGH)
-		} else {
-			_ = os.Unsetenv("GH_TOKEN")
-		}
-	})
-	_ = os.Unsetenv("GH_TOKEN")
-	t.Setenv("GITHUB_TOKEN", "test-token")
+	setGitHubTokenFallbackEnv(t)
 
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/backends/discovery_test.go:178` — `setGitHubTokenFallbackEnv` correctly preserves env isolation: `os.LookupEnv` captures presence + value, `t.Cleanup` is registered **before** the mutation (so cleanup runs even if a later step panics), then `os.Unsetenv("GH_TOKEN")` clears any host token and `t.Setenv("GITHUB_TOKEN", "test-token")` is auto-restored by the test framework. `t.Helper()` is set so failures point at the caller.
- **[info]** `internal/backends/discovery_test.go:157` — Applying the helper to `TestCheckGitHubMCPInRuntimeUsesBackendSetup` is a strict positive fix, not just a refactor: previously only `t.Setenv("GITHUB_TOKEN", "test-token")` was called, so on a developer/CI host with `GH_TOKEN` already set, the `GH_TOKEN=test-token` assertion in the container env would have been silently masked by the host token. The helper closes that gap on both call sites.
- **[info]** Tests do not call `t.Parallel()`, which is correct for process-global env mutation; the helper preserves that invariant.
- **[info]** Helper scope is correctly narrow (only `GH_TOKEN`/`GITHUB_TOKEN`); the broader env scrub of other backend tokens belongs at the `env -u …` invocation level documented in the PR verification, not in this fallback-specific helper.

### Summary
Clean, defensive test-only refactor; centralizes the GH_TOKEN host-leakage guard and fixes a previously-missing guard in one of the two call sites. No production behavior change.

PR #635, opened 2026-05-20
refactor: compare raw json null as bytes

eloylp:
## Summary

- Replaces the `rawJSONString` null sentinel check with `bytes.Equal` so the byte-oriented helper does not convert the trimmed raw JSON value to a string for comparison.
- Keeps the same empty and `null` handling before quoted string unwrapping.

## Verification

- `go test ./internal/ai`
- `git diff --check`

### DIFF
--- a/internal/ai/stream_steps.go
+++ b/internal/ai/stream_steps.go
@@ -269,7 +269,7 @@ func codexOutput(result, output json.RawMessage) string {
 // quoted strings to their textual content. Empty / null values become "".
 func rawJSONString(raw json.RawMessage) string {
 	trimmed := bytes.TrimSpace(raw)
-	if len(trimmed) == 0 || string(trimmed) == "null" {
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/ai/stream_steps.go:272` — Behaviour-equivalent stdlib swap. `trimmed` is `[]byte` (from `bytes.TrimSpace` over `json.RawMessage`), and `bytes.Equal(trimmed, []byte("null"))` matches the prior `string(trimmed) == "null"` byte-for-byte. The preceding `len(trimmed) == 0` short-circuit means nil/empty inputs never reach the equality check, so the nil-vs-empty treatment of `bytes.Equal` is irrelevant here. `bytes` is already imported (used on the line above), so no new dependency. Quoted-string unwrap path on the next branch is untouched.

### Summary
Tiny, provably equivalent comparator swap; CI green; no behavioural risk.

PR #638, opened 2026-05-20
Store: enforce config integrity constraints

eloylp:
## Summary

- Adds DB-level integrity constraints for mutable fleet configuration: agents, repos, backends, prompts, skills, guardrails, bindings, graph layouts, token budgets, and workspace guardrail selections.
- Moves agent `skills` and `can_dispatch` from JSON columns into normalized join tables with FK protection.
- Adds a second migration that decouples catalog database identity from public refs: prompts, skills, and guardrails now use opaque internal SQLite IDs while REST/MCP/import/export continue exposing stable public refs.
- Keeps user-facing validation in store/service paths so API callers get clear errors instead of raw SQLite constraint failures.
- Updates docs to describe stable public refs vs internal database IDs.

## Validation

- `GOCACHE=/tmp/go-build-agents go test ./internal/store -count=1 -timeout=120s`
- `GOCACHE=/tmp/go-build-agents go test ./internal/service ./internal/config ./internal/mcp ./internal/workflow ./internal/scheduler ./internal/daemon/fleet ./internal/daemon/runners ./internal/daemon/config -count=1 -timeout=180s`
- `GOCACHE=/tmp/go-build-agents go test ./internal/daemon ./internal/daemon/observe -count=1 -timeout=180s`
- `GOCACHE=/tmp/go-build-agents go test ./... -count=1 -timeout=240s`
- Applied migrations `032` + `033` to copied lonestar DB and checked `PRAGMA foreign_key_check;` plus `PRAGMA quick_check;`.
- Ran local daemon smoke test with disposable DB, REST import, repo-scoped catalog/agent creation, negative visibility validation, direct SQLite FK/ref inspection, and MCP `list_agents` / `get_prompt` by public ref.


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[low]** `internal/store/migrations/032_config_integrity_constraints.sql:277-284` — `graph_layouts` rows whose `scope` is not `workspace:%` collapse silently into `workspace_id='default'`. Per migration 023 only `'global'` and `'workspace:%'` should exist (and 023 already converted `'global'` to `'workspace:default'`), so on a well-formed DB this is a no-op. If a column-shape drift snuck in via a future code path, the silent default-bucket would be hard to debug; a `RAISE(ABORT)` for unknown formats would be more honest, but I'd only ask for that if you already had reason to suspect drift. Not a blocker.
- **[low]** `internal/store/budgets.go:138-176` — `importTokenBudgetsTx` only calls `validateBudget`, not `validateBudgetRefs`. The new `token_budgets_validate_insert` trigger covers workspace-qualified refs (workspace, workspace+*) and any backend ref, but the simple cross-workspace `repo`/`agent` scopes are validated nowhere on the import path. A YAML import can persist a `repo`/`agent` budget for a nonexistent repo/agent, and only `CheckBudgets` at run time will iterate it (it just won't match anything). Consistent with the documented "simple scopes are global by name", but worth surfacing in case stricter validation was intended.
- **[low]** `internal/store/guardrails.go:139-151` — `UpsertGuardrailTx`'s `ON CONFLICT(ref) DO UPDATE SET workspace_id = excluded.workspace_id, repo = excluded.repo, …` allows an operator POST to re-scope a built-in (e.g., move `security` from global to workspace-scoped). `is_builtin` and `default_content` are correctly preserved, but a re-scope quietly removes the row from global render order. Probably intentional given POST = full-replace semantics; a sentence in `docs/api.md` Guardrails could help if not.
- **[info]** `internal/store/migrations/033_catalog_internal_ids.sql:51-60` — `agents_copy_033` is populated via `INNER JOIN prompt_id_map_033`. If any agent had a dangling `prompt_id`, the agent row would silently drop. This is provably safe because migration 032 adds `prompt_id NOT NULL REFERENCES prompts(id) ON DELETE RESTRICT` and runs under `PRAGMA foreign_keys=ON`, so 032 would have aborted on a dangling ref before 033 ever ran. Worth a one-line comment in 033 noting the FK invariant the inner join relies on, so future maintainers don't loosen 032 and lose data here.
- **[info]** `internal/store/migrations/033_catalog_internal_ids.sql` — triggers on tables not dropped in 033 (`repos_scope_agent_delete`, `token_budgets_validate_insert/_update`, `token_budgets_workspace_delete`, `token_budgets_repo_delete`, `token_budgets_backend_delete`) survive the table rebuilds by name resolution at fire time; only `token_budgets_agent_delete` (ON `agents`) is correctly dropped and recreated. Solid trigger lifecycle handling.

### Summary
Migration 032 follows the SQLite 12-step rebuild pattern carefully (copy-out → drop → rebuild with FKs/CHECKs/triggers → copy-in), and migration 033 separates opaque internal IDs from stable public refs without breaking the existing API contract — `id = ? OR ref = ?` lookups make old REST/MCP/YAML refs continue to resolve, and the `Load` path returns the `ref` column as the user-facing `id`. Backend upsert correctly switched from `INSERT OR REPLACE` to `ON CONFLICT DO UPDATE` so the new `agents.backend ON DELETE RESTRICT` FK doesn't get tripped by the implicit delete. The post-write validation surface (workspace existence, repo existence for repo-scoped agents, agent/prompt visibility within scope) is materially tighter than before. Tests cover the pre-031 schema migration, FK rejection on delete, and the new public-ref vs internal-id distinction in import/load round-trips. The four `[low]`/`[info]` items are small documentation/comment improvements, not correctness defects.

PR #640, opened 2026-05-20
Fix Config Runtime workspace override selector

### LINKED ISSUE #639: Add workspace selector for runner image overrides in Config Runtime
## Problem

The Runtime tab in the Config section has a **Workspace override** form for `runner_image`, but the Config section no longer exposes a workspace selector. That matches the broader product decision that Config is mostly global and should not be filtered by the app-wide workspace selector.

However, the workspace runner image override is explicitly workspace-scoped. Today the UI still derives the target workspace from the hidden persisted selected workspace (`useSelectedWorkspace()`), so users can see and save a workspace override without an obvious way to choose which workspace they are editing. This is confusing and can lead to updating the wrong workspace.

The backend surfaces already model this correctly:

- REST: `PUT /workspaces/{workspace}/runtime` with `{ runner_image }`.
- MCP: workspace runner image update accepts an explicit `workspace` parameter.
- Store/domain: `workspaces.runner_image` is the per-workspace override.
- Runtime execution: workflow resolves global runtime settings and then applies `workspace.runner_image` when set.

The missing piece is the UI affordance inside Config > Runtime.

## Required behavior

Add an explicit workspace selector inside the **Workspace override** card in the Runtime tab.

The selector should:

- List all workspaces from the existing workspace data already loaded in config, or fetch `/workspaces` if that data is not available at render time.
- Be local to the Runtime tab / Workspace override card.
- Not reintroduce an app-wide workspace filter for the whole Config section.
- Make the selected workspace label visible in the card copy, input label, and save status/error messages where useful.
- Load the selected workspace's current `runner_image` into the override input when the selected workspace changes.
- Save with `PUT /workspaces/{selected_workspace}/runtime`.
- Keep the current behavior that an empty workspace override clears the override and falls back to the global runner image.

## UX expectations

The Runtime tab should be conceptually split like this:

1. **Global runner runtime settings**
   - Global runner image.
   - Container constraints.
   - Save global runtime.

2. **Workspace runner image override**
   - Workspace selector.
   - Current effective fallback hint from the global runner image.
   - Workspace-specific runner image input.
   - Save workspace override.

This should make it clear that runtime constraints remain global for now, while runner image can be overridden per workspace.

## Implementation notes

- Current UI code is in `internal/ui/src/app/config/page.tsx` around the runtime state and the `Workspace override` card.
- Avoid coupling this field to the hidden persisted global workspace value. A local `selectedRuntimeWorkspace` state is clearer.
- Initialize the local selector to the persisted workspace if available, otherwise `default`, otherwise the first workspace returned by the API.
- Keep `workspaceRunnerImage` synchronized with the selected runtime workspace, not with the app-wide workspace state.
- If the workspace list changes after a save/import, handle missing selected workspace by switching to the first available workspace and showing a small notice.
- Add/adjust UI tests if the project already has coverage for the Config Runtime tab. At minimum, add a focused test for changing the selected workspace and saving to the correct endpoint.

## Acceptance criteria

- A user can choose the workspace whose runner image override they are editing from Config > Runtime.
- Saving workspace A sends `PUT /workspaces/A/runtime` and does not affect workspace B.
- Switching from workspace A to workspace B updates the input to workspace B's current `runner_image` value.
- Empty input clears that workspace's override and shows the global image as fallback.
- The Config section remains globally oriented; only this Runtime override card contains a workspace selector.
- REST and MCP behavior remain unchanged except for any tests/docs needed to clarify the UI behavior.
- Documentation mentions that per-workspace runner image overrides are edited from Config > Runtime using the local workspace selector.

eloylp:
## Summary

- Adds a local workspace selector to Config -> Runtime's Workspace override section.
- Keeps workspace runner image editing independent from the app-wide workspace filter and saves to the selected workspace endpoint.
- Updates docs and adds a focused UI test for switching workspaces and saving the selected override.

Closes #639

## Verification

- `npm test -- src/app/config/page.test.tsx`
- `npm test`
- `npm run build`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -99,7 +99,7 @@ runtime:
 
 Runtime settings are fleet state because operators may need to switch runner images or constraints without rebuilding the daemon. They are stored in SQLite, returned by `/config`, included in import/export, and editable through Config -> Runtime, REST, and MCP. They are not startup env overrides; use the database-backed surfaces for runner image and container policy changes.
 
-The global runner image applies to every run unless a workspace sets `runner_image`. Constraints are passed to Docker where supported: CPU, memory, PID limit, timeout, and network mode. Advanced egress filtering is not part of the v1 runtime settings.
+The global runner image applies to every run unless a workspace sets `runner_image`. In the dashboard, workspace runner image overrides are edited from Config -> Runtime with the local workspace selector in the Workspace override section. Constraints are passed to Docker where supported: CPU, memory, PID limit, timeout, and network mode. Advanced egress filtering is not part of the v1 runtime settings.
 When patching global runtime settings through REST or MCP, omitted fields are preserved. Empty string clears string constraints such as `cpus`, `memory`, and `network_mode`; empty `runner_image` resets to the built-in default because the global image cannot be unset.
 
 ## `backends`
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -101,7 +101,7 @@ Raw agent memory markdown per `(workspace, agent, repo)` key. Useful for inspect
 
 ### Config
 
-Current fleet config snapshot. Includes YAML import/export for shareable fleet strategy. The Runtime tab controls the global runner image, basic Docker constraints, and the selected workspace's optional runner image override. Secret values are not shown in the dashboard; Claude, Codex, GitHub MCP, `gh` credentials, and runner git identity come from daemon environment variables and are injected into each ephemeral runner container.
+Current fleet config snapshot. Includes YAML import/export for shareable fleet strategy. The Runtime tab controls the global runner image and basic Docker constraints. Per-workspace runner image overrides are edited from Config -> Runtime with the local workspace selector in the Workspace override section. Secret values are not shown in the dashboard; Claude, Codex, GitHub MCP, `gh` credentials, and runner git identity come from daemon environment variables and are injected into each ephemeral runner container.
 
 ![Config inspector, webhook_secret rendered as `[redacted]`](img/config.png)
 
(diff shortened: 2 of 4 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[low]** `internal/ui/src/app/config/page.tsx:404` — `runtimeWorkspaces` is recomputed on every render (`configWorkspaces.length > 0 ? configWorkspaces : workspaceCatalog`), so the workspace-selection effect's dependency reference changes each render and the effect re-runs. The `hasSelection` early-return keeps it a no-op after stabilisation, so it isn't a correctness bug, but wrapping the derivation in `useMemo` would keep the dependency stable and make the intent clearer.
- **[low]** `internal/ui/src/app/config/page.tsx:413` — Initial-load edge case: when `useSelectedWorkspace()` initially returns `workspace='default'` and `workspaces=[]`, the effect eagerly sets `selectedRuntimeWorkspace='default'`. If the workspace catalog later arrives without a `default` entry, the next pass falls into the `selectedRuntimeWorkspace` branch and surfaces a "Workspace default is no longer available." notice on first paint, even though the user never picked it. `useSelectedWorkspace` already shows the same notice at the app level, so the runtime card duplicates the signal. Consider deferring the initial `setSelectedRuntimeWorkspace` until the catalog is non-empty, or skipping the notice when the previous value matches the unresolved default sentinel.
- **[nit]** `internal/ui/src/app/config/page.tsx:403` — `workspaceCatalog as WorkspaceRuntime[]` is a loose cast: `Workspace` from `@/lib/workspace` does not declare `runner_image`. It works at runtime because the missing field reads back as `undefined` → `''`, but the cast obscures the shape mismatch. A narrower mapping (e.g. `workspaceCatalog.map(w => ({ ...w, runner_image: '' }))`) would document the fallback assumption.

### Summary
Local `selectedRuntimeWorkspace` cleanly decouples the override save from the app-wide workspace, the workspaceRunnerImage effect now keys off the local selector (so switching workspaces refreshes the input), the PUT body and empty-input fallback contract are unchanged, docs were updated, and the new focused vitest passes against head and fails against base (non-vacuous). CI green.

PR #642, opened 2026-05-20
refactor: inline workflow prompt helper

eloylp:
## Summary

- Inlines the single-use `testConfigHasPrompt` helper in workflow test setup.
- Uses `slices.ContainsFunc` directly where the prompt lookup is needed, keeping the same seeded prompt behavior with less local test scaffolding.

## Verification

- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/workflow/test_helpers_test.go
+++ b/internal/workflow/test_helpers_test.go
@@ -3,6 +3,7 @@ package workflow
 import (
 	"fmt"
 	"path/filepath"
+	"slices"
 	"testing"
 
 	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Tiny, behavior-preserving test-helper refactor: the single-use `testConfigHasPrompt(cfg, cfg.Agents[i].PromptRef)` is replaced by an inline `slices.ContainsFunc(cfg.Prompts, func(p fleet.Prompt) bool { return p.Name == cfg.Agents[i].PromptRef })`. Both iterate `cfg.Prompts` and return true iff any element's `Name` equals the agent's `PromptRef`, so semantics are preserved. The `slices` import is added cleanly, the now-dead helper is removed, scope stays test-only, and CI is green. Auto-merging as a contained refactor.

PR #651, opened 2026-05-21
refactor: compare raw messages as bytes

eloylp:
## Summary

- Replaces a `json.RawMessage` string conversion in the Anthropic proxy test with `bytes.Equal`.
- Keeps the assertion byte-for-byte equivalent while avoiding a temporary string allocation.

## Verification

- `go test ./internal/anthropic_proxy`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/anthropic_proxy/translate_test.go
+++ b/internal/anthropic_proxy/translate_test.go
@@ -1,6 +1,7 @@
 package anthropicproxy
 
 import (
+	"bytes"
 	"encoding/json"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/anthropic_proxy/translate_test.go:375` — Swap of `string(block.Input) != \`{"location":"Tokyo"}\`` for `!bytes.Equal(block.Input, []byte(\`{"location":"Tokyo"}\`))` is behavior-equivalent. `block.Input` (`json.RawMessage`, underlying `[]byte`) is assignable to the `[]byte` parameter, and `bytes.Equal` returns the same boolean for this non-empty comparison. The nil-vs-empty edge is irrelevant here (target is non-empty) and `bytes.Equal` matches `string()==string()` semantics anyway.
- **[info]** `internal/anthropic_proxy/translate_test.go:4` — Adds the `bytes` import (package did not previously import it). Necessary and benign; `%s` in the `t.Errorf` format still renders the byte slice correctly, so the failure message is unchanged.

### Summary
Test-only, byte-for-byte equivalent stdlib swap; CI green. Approving.

PR #653, opened 2026-05-21
refactor: compare observe test messages as bytes

eloylp:
## Summary

- Replaces two byte-slice equality checks in observe SSE tests with bytes.Equal.
- Keeps the existing failure messages unchanged while avoiding temporary string conversion for the comparison.

## Verification

- go test ./internal/observe
- go test ./...
- git diff --check

### DIFF
--- a/internal/observe/observe_test.go
+++ b/internal/observe/observe_test.go
@@ -41,7 +41,7 @@ func TestSSEHubPublishDelivers(t *testing.T) {
 
 	select {
 	case msg := <-ch:
-		if string(msg) != "hello" {
+		if !bytes.Equal(msg, []byte("hello")) {
 			t.Fatalf("want 'hello', got %q", string(msg))
 		}
 	default:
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/observe/observe_test.go:44,62` — `string(msg) != "x"` → `!bytes.Equal(msg, []byte("x"))` is behavior-preserving. `msg` is `[]byte` from the SSE hub channel, so both operands are byte slices. `bytes.Equal` and the previous `string()` comparison agree on every input (including nil-vs-empty, which `bytes.Equal` treats as equal — irrelevant here since both literals are non-empty).
- **[info]** Failure diagnostics unchanged — both `t.Fatalf` calls still format with `%q` over `string(msg)`, so failure output is identical to before. `bytes` is already imported; no new dependency.

### Summary
Tiny, test-only equality swap that preserves both comparison semantics and failure messages. CI green.

PR #654, opened 2026-05-21
refactor: compare runtime network mode directly

eloylp:
## Summary

- Removes an unnecessary string conversion when checking Docker network mode in the runtime host config test.
- Keeps the same typed value in the failure message.

## Verification

- go test ./internal/runtime
- go test ./...
- git diff --check

### DIFF
--- a/internal/runtime/docker_test.go
+++ b/internal/runtime/docker_test.go
@@ -20,7 +20,7 @@ func TestHostConfigAppliesRuntimeConstraints(t *testing.T) {
 	if err != nil {
 		t.Fatalf("hostConfig: %v", err)
 	}
-	if string(cfg.NetworkMode) != "none" {
+	if cfg.NetworkMode != "none" {
 		t.Fatalf("NetworkMode = %q, want none", cfg.NetworkMode)
 	}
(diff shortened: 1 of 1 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- **[info]** `internal/runtime/docker_test.go:23` — Dropping the `string()` conversion is behavior-preserving. `cfg.NetworkMode` is the Docker SDK `container.NetworkMode`, a defined `string` type; comparing it to the untyped constant `"none"` converts the constant to `NetworkMode` and compares the identical underlying bytes — exactly what `string(cfg.NetworkMode) != "none"` did. No nil/empty edge case since these are strings, not slices.
- **[info]** Failure message `t.Fatalf("NetworkMode = %q, want none", cfg.NetworkMode)` is unchanged; `%q` formats a string-based type correctly, so diagnostics are preserved.
- **[info]** Change is test-only (`docker_test.go`); production `hostConfig` and all default/disabled/enabled mode handling are untouched. CI `test` job green confirms compilation and a still-passing test.

### Summary
Clean, behavior-preserving removal of a redundant string conversion in a test comparison — no concerns.

PR #668, opened 2026-05-30
ci: extend main race test timeout

eloylp:
## Summary
- increase the main-branch race test package timeout from Go's default 10m to 20m
- keep the race suite otherwise unchanged

## Context
The main CI run for merge commit 2c1d8bf failed in the push-only race step because internal/daemon hit the default 600s package timeout. The log did not show a race detector report or assertion failure; the package timed out under the full ./... race run.

## Validation
- not run; workflow-only change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,4 +49,4 @@ jobs:
 
       - name: go test race
         if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-        run: go test ./... -race -count=1
+        run: go test ./... -race -count=1 -timeout=20m


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The workflow-only timeout increase is an acceptable CI-only mitigation for the main-branch race job: it keeps the same `./... -race -count=1` coverage, only extends Go's per-package timeout from the 10m default to 20m, and does not suppress race detector output or test failures. Given the reported failure was a package timeout in `internal/daemon` without a race report or assertion failure, this does not appear to mask a correctness regression; a future timeout under the higher bound would still fail loudly and should be investigated.

PR #680, opened 2026-06-01
feat: add improvement clarification loop

eloylp:
## Summary
- add one editable clarification field per self-improvement recommendation
- add REST and MCP clarify actions that store the latest clarification and enqueue a fresh `agents.improvement` run
- pass prior recommendation and clarification context into the self-improvement analyst
- add a dashboard Clarify modal showing original feedback, attribution, target context, and the editable clarification
- document the clarification REST/MCP flow

## Tests
- `GOCACHE=/tmp/go-build-agents go test ./internal/store ./internal/workflow ./internal/daemon/observe ./internal/daemon ./internal/mcp`
  - `internal/store`, `internal/workflow`, and `internal/mcp` passed
  - `internal/daemon/observe` and `internal/daemon` needed rerun outside the sandbox because `httptest` local port binding is blocked inside the sandbox
- `GOCACHE=/tmp/go-build-agents go test ./internal/daemon/observe ./internal/daemon` outside sandbox
- `git diff --check`

## Not run
- `npm run build` in `internal/ui`: this clean worktree has no `node_modules`, so `next` is not installed locally.


eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[medium]** `internal/ui/src/app/improvements/page.tsx:114` — The Clarify modal is opened from the `/improvements/recommendations` list row, but that endpoint uses `ListSelfImprovementRecommendations`, which scans recommendations with `includeFeedback=false` (`internal/store/self_improvement.go:565`). As a result `clarifying.feedback` is always absent in the modal, so the operator sees `No feedback body available.` and unresolved linked agent/prompt/skill/guardrail context instead of the original feedback and attribution metadata promised by the PR/docs. This makes the maintainer clarification flow lose the exact evidence the user is supposed to clarify; fetch the detail endpoint before opening the modal or include the feedback snapshot in the list response.

### Summary
The Go store/workflow/API/MCP paths are mostly coherent and the targeted Go suite passes, but the dashboard clarification flow does not actually show the original feedback context it requires.

eloylp:
Pushed `83846c7` to address the clarification modal review finding. The Clarify action now loads the recommendation detail endpoint before opening the modal, so original feedback body, source URL, and attribution metadata come from the detailed payload instead of the compact list row.

Verified locally with `npm ci`, `npm test`, `npm run build`, and `git diff --check`.

PR #693, opened 2026-06-16
feat(observe): resolve self-improvement attribution via GitHub artifact ownership

### LINKED ISSUE #692: Resolve self-improvement attribution through GitHub artifact ownership
## Problem

Self-improvement feedback works well when `/agents improve` is written in the same GitHub body that contains signed `agents-run` metadata. It fails in a common review workflow: a maintainer leaves `/agents improve ...` as an inline PR review/diff comment. That human comment usually does not carry hidden attribution metadata, even though the surrounding review/comment/thread/commit may have been produced by an agent and may contain signed metadata.

Today `run_attributions` knows the exact run facts per `span_id`, but the daemon loses the reverse mapping from GitHub artifacts back to the span that produced them. As a result, inline feedback often becomes `needs_user_input`, or loose time-window inference can point at the wrong run, including internal analyst runs.

## Goal

Make normal GitHub review feedback actionable without requiring humans to paste or preserve attribution metadata. The daemon should deterministically walk GitHub artifact ownership context before falling back to conservative inference.

Desired model:

```text
human /agents improve comment
  -> direct signed metadata, if present
  -> parent/review/thread/PR artifact ownership
  -> span_id
  -> run_attributions
  -> exact agent + prompt/skill/guardrail versions for the analyst
```

## Current implementation points

Relevant code paths observed before filing this issue:

- `internal/observe/attribution.go`
  - `ResolveRunAttribution` currently tries same-body exact metadata, then `inferRunAttributions` against `run_attributions`.
  - `AttributionQuery` currently has body/commit message, workspace/repo/number/head SHA/time, but no GitHub artifact identity or review ancestry fields.
  - `run_attributions` is the canonical run snapshot. Keep it that way.
- `internal/webhook/handler.go`
  - `captureFeedback` calls `observe.ResolveRunAttribution` only when a `/agents improve` marker exists.
  - `webhookComment` currently captures `id`, `html_url`, `body`, `path`, `line`, `side`, `diff_hunk`, `commit_id`, timestamps.
  - It does not currently capture review ancestry fields such as `in_reply_to_id` or `pull_request_review_id`.
  - `pull_request_review_comment` passes only the human comment body to attribution resolution, so inline comments with no footer cannot resolve exact ownership.
- `internal/store/migrations/036_run_attributions.sql`
  - `run_attributions` already stores: `span_id`, workspace/repo/issue-or-PR number, event ids, agent/backend identity, prompt ref/version, skill version ids, guardrail version ids, head SHA, branch, created_at.
- `internal/store/migrations/037_self_improvement_feedback.sql`
  - `self_improvement_feedback` stores the human evidence row plus copied `linked_*` attribution fields resolved at ingestion.

## Design

Add a new narrow reverse-index table for GitHub artifacts that carry valid signed agent attribution metadata.

This table must not duplicate `agent_id`, `prompt_version_id`, skills, guardrails, backend, etc. Those remain owned by `run_attributions`. The new table should map GitHub object identity to `span_id`.

Suggested table name: `run_attribution_artifacts`.

Suggested fields:

```sql
id INTEGER PRIMARY KEY AUTOINCREMENT,
workspace_id TEXT NOT NULL DEFAULT 'default',
repo_owner TEXT NOT NULL,
repo_name TEXT NOT NULL,
issue_or_pr_number INTEGER NOT NULL DEFAULT 0,
source_type TEXT NOT NULL,
github_comment_id INTEGER NOT NULL DEFAULT 0,
github_review_id INTEGER NOT NULL DEFAULT 0,
github_review_comment_id INTEGER NOT NULL DEFAULT 0,
github_parent_comment_id INTEGER NOT NULL DEFAULT 0,
github_delivery_id TEXT NOT NULL DEFAULT '',
source_url TEXT NOT NULL DEFAULT '',
author_login TEXT NOT NULL DEFAULT '',
file_path TEXT NOT NULL DEFAULT '',
line INTEGER NOT NULL DEFAULT 0,
side TEXT NOT NULL DEFAULT '',
commit_sha TEXT NOT NULL DEFAULT '',
span_id TEXT NOT NULL,
metadata_json TEXT NOT NULL DEFAULT '',
github_created_at TIMESTAMP NULL,
github_updated_at TIMESTAMP NULL,
observed_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
```

Indexes/constraints should support:

- lookup by `(workspace_id, repo_owner, repo_name, source_type, github_comment_id)`
- lookup by `(workspace_id, repo_owner, repo_name, github_review_id)`
- lookup by `(workspace_id, repo_owner, repo_name, github_review_comment_id)`
- lookup by `(workspace_id, repo_owner, repo_name, github_parent_comment_id)` if useful
- lookup by `(workspace_id, repo_owner, repo_name, issue_or_pr_number, file_path, commit_sha, observed_at)` for conservative PR/thread context
- joining `span_id` to `run_attributions(span_id)`; use an FK if it fits existing migration constraints, otherwise keep a logical reference but always load canonical details from `run_attributions`.

## Capture behavior

Webhook handling should inspect relevant GitHub artifacts for signed metadata even when they do not contain `/agents improve`.

Capture valid signed artifacts from:

- `issue_comment`
- `pull_request_review`
- `pull_request_review_comment`
- push commit trailers when available in the webhook payload, if practical in this PR

For each artifact:

1. Extract `agents-run` hidden comments or `Agents-Attribution` trailers using the existing decode/verify logic.
2. Require valid signature/context when `AGENTS_ATTRIBUTION_SIGNING_SECRET` is configured.
3. Require metadata workspace/repo/issue-or-PR number to match the webhook context when those fields are present.
4. Store only valid authoritative artifact rows, or store invalid rows only as non-authoritative diagnostics if that is useful. Invalid/unsigned/malformed/foreign-instance/wrong-context metadata must not be used for exact attribution.
5. Do not mutate GitHub. The daemon remains read-only against GitHub.

Extend webhook payload structs as needed, especially for PR review comments:

- `comment.in_reply_to_id`
- `comment.pull_request_review_id`
- keep `comment.id`, `path`, `line`, `side`, `diff_hunk`, `commit_id`

## Resolution behavior

Update `observe.ResolveRunAttribution` and `AttributionQuery` to support artifact-aware lookup.

Recommended resolution order:

1. Direct signed metadata in the `/agents improve` body.
2. For `pull_request_review_comment`, exact artifact lookup by the current comment id, if it itself carries metadata.
3. Parent review comment lookup through `in_reply_to_id`.
4. Owning review lookup through `pull_request_review_id` / `github_review_id`.
5. Same PR/thread-ish artifact lookup using repo + PR number + path + commit SHA, only if it produces exactly one valid candidate.
6. Commit trailer/artifact lookup when available.
7. Existing `run_attributions` time-window inference, but keep it conservative.
8. Unresolved.

Exact attribution should only come from valid signed metadata or from an artifact row that was created from valid signed metadata.

## Avoid false positives

Please tighten the fallback behavior while working here:

- Do not infer self-improvement feedback to internal analyst runs such as `internal-catalog-analyst` / the self-improvement analyst.
- If multiple artifact candidates match, return unresolved/ambiguous instead of choosing one.
- Prefer artifact ownership over time-window inference.
- If artifact lookup only matches broad PR context without parent/review/path/commit evidence, keep it unresolved unless the match is uniquely strong.

## Store/API shape

Add store primitives in `internal/store` for:

- upserting a run attribution artifact
- resolving artifact by exact comment/review/review-comment identity
- resolving artifact ancestry for feedback queries
- loading the canonical `run_attributions` row by `span_id`

`self_improvement_feedback` should continue storing copied `linked_*` fields as ingestion-time evidence. It should not become dependent on joining the artifact table at read time.

Consider adding a diagnostic/mode value so feedback can say whether attribution came from:

- `direct`
- `artifact_comment`
- `artifact_parent_comment`
- `artifact_review`
- `artifact_pr_context`
- `commit_trailer`
- `inferred`
- `unresolved`

If schema changes are needed for this, keep them focused.

## Tests

Add focused tests for:

- Direct hidden metadata still resolves exact.
- A PR review comment with valid signed metadata is stored as an artifact even without `/agents improve`.
- A human inline reply with `/agents improve` and `in_reply_to_id` resolves through the parent artifact.
- A human inline review comment resolves through owning `pull_request_review_id` when the review artifact has valid signed metadata.
- Copied signed metadata from another repo/PR is rejected.
- Invalid signatures are logged/ignored and do not become exact attribution.
- Multiple matching PR artifact candidates produce ambiguous/unresolved attribution.
- Internal analyst runs are ignored by loose fallback inference.
- Feedback still stores and enqueues analysis with unresolved attribution when no valid ownership exists.

Recommended packages:

```bash
go test ./internal/observe ./internal/webhook ./internal/store ./internal/daemon
```

Use more focused package tests during development, but the final PR should pass the relevant self-improvement/webhook/store tests.

## Docs

Update docs after implementation:

- `docs/self-improvement-feedback.md`
- `docs/run-attribution.md`
- `README.md` if the feature wording needs one sentence of clarification
- `CLAUDE.md` / `AGENTS.md` code map or operational notes if a new table or resolver behavior is added

Docs should say that maintainers can use `/agents improve` in normal PR review comments, and the daemon resolves attribution by walking signed GitHub artifact ownership before falling back to inference.

## Acceptance criteria

- Inline PR review feedback can resolve exact attribution through a signed parent/review artifact without the human comment containing metadata.
- `run_attributions` remains the canonical source for agent/catalog/run facts.
- The new artifact table is a narrow GitHub artifact -> span reverse index.
- The analyst receives exact prompt/skill/guardrail version context when artifact ancestry resolves.
- Existing unresolved/clarification behavior remains for genuinely unattributed feedback.
- No daemon GitHub writes are introduced.
- Tests cover direct, artifact ancestry, invalid signature, ambiguity, and unresolved paths.

eloylp:
## Summary

- Adds `run_attribution_artifacts` table (migration 056) as a narrow reverse-index from GitHub artifact identity (comment_id, review_id, review_comment_id) to `span_id`, keeping `run_attributions` as the canonical run-fact source
- Extended `AttributionQuery` with `CommentID`, `ReviewCommentID`, `ReviewID`, `InReplyToID`, `PullRequestReviewID`, `FilePath` to carry artifact ancestry context into the resolver
- `ResolveRunAttribution` now walks the artifact ownership chain (comment → parent reply → owning review → conservative PR context) before falling back to time-window inference; added `mode` constants (`direct`, `artifact_comment`, `artifact_parent_comment`, `artifact_review`, `artifact_pr_context`, `commit_trailer`, `inferred`, `unresolved`)
- `inferRunAttributions` now excludes `internal-catalog-analyst` runs from loose inference — only exact signed metadata or artifact ancestry can resolve to those agents
- `webhookComment` extended with `InReplyToID` and `PullRequestReviewID`; `CaptureArtifact` is called for every `issue_comment`, `pull_request_review`, and `pull_request_review_comment` delivery (not just `/agents improve`), storing valid signed artifacts in the reverse-index
- 9 new focused tests covering all required paths: direct metadata, artifact chain (comment, parent reply, review), wrong-repo rejection, invalid signature rejection, ambiguous PR context, internal analyst exclusion from inference, and unresolved fallback

## Test plan

- [ ] `go test ./internal/observe/...` — all 9 new artifact tests pass
- [ ] `go test ./internal/webhook/...` — existing webhook tests still pass
- [ ] `go test ./internal/store/...` — migration applied, schema validates
- [ ] `go test ./...` — full suite green

Closes #692


eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- **[low]** `internal/webhook/handler.go:792-807` — `captureArtifact` unconditionally overwrites `in.WorkspaceID`, `in.RepoOwner`, and `in.RepoName` from the `repo` argument, but all three call sites already populate those same fields in the `RunAttributionArtifactInput` literal they pass in. One of the two should do the work; right now both do, silently. The override in the helper is a safe guard but the call-site initializations are dead writes. Could be simplified by either having the helper set them (and callers omit them) or vice versa.

- **[low]** `internal/observe/attribution.go:131-170` — The inline step comments (`// Step 2`, `// Step 3`, `// Step 4`, `// Step 6`) skip numbering the `ReviewID` direct-check block that sits between "Step 4" and "Step 5" (`q.ReviewID > 0`). The code is functionally correct, but the numbering doesn't line up with the design doc or the `resolveArtifactAttribution` doc comment's "2–6" range. A `// Step 5 (ReviewID direct)` comment or adjusting the labels would make the walk easier to audit later.

- **[low]** `internal/observe/attribution_artifacts_test.go:342` — The tamper expression `validComment[len(validComment)-5+5:]` evaluates to `validComment[len(validComment):]` which is always `""`. The intent (replace last 5 chars) is correct and the test does invalidate the signature, but the trailing slice expression is dead code. `tampered := validComment[:len(validComment)-5] + "XXXXX"` is equivalent and clearer.

### What's correct

- **SQL injection**: Every query uses `?` placeholders throughout `attribution_artifacts.go`. No string interpolation of external input anywhere.
- **Concurrency safety**: `selfImprovementInternalAgentNames` is a package-level read-only map (safe). All DB writes go through `s.db.Exec`/`s.db.QueryRow` which are safe for concurrent use (SQLite WAL). No unprotected shared mutable state introduced.
- **Unique index design**: The three partial unique indexes cleanly distinguish issue comments (`source_type, github_comment_id > 0`), PR reviews (`github_review_id > 0 AND github_review_comment_id = 0`), and PR review comments (`github_review_comment_id > 0`). `ON CONFLICT DO NOTHING` idempotency is correct because all three indexes enforce the right keys and the filters prevent cross-type collisions.
- **`runAttributionArtifactByReviewID` filter**: `AND github_review_comment_id=0` correctly isolates PR review rows from review comment rows that happen to share the same `github_review_id` — crucial for the step 4/ReviewID path.
- **Ambiguous PR context short-circuit**: `resolveArtifactAttribution` returns `(resolution, true)` when PR context yields >1 candidate, which correctly prevents the caller from falling through to time-window inference for an ambiguous result.
- **`artifactResolution` unknown span**: Returns `(AttributionUnresolved, true)` rather than `(_, false)` when the artifact exists but the span isn't in `run_attributions`. This correctly short-circuits inference (the artifact reference is definitive evidence that inference would be wrong).
- **Internal analyst exclusion**: The filter in `inferRunAttributions` plus the map guard in attribution.go ensures `internal-catalog-analyst` is only reachable via exact paths; the `TestInferenceSkipsInternalAnalystRuns` two-part test (one mixed, one analyst-only) covers both branches of the switch.
- **Issue #692 acceptance criteria**: All items satisfied — artifact ancestry resolution, `run_attributions` stays canonical, no GitHub writes, wrong-repo rejection, invalid-sig rejection, ambiguity → unresolved, and all nine test scenarios from the issue checklist are present.

### Summary

Solid, well-scoped implementation of artifact-based attribution resolution. The three minor findings are cosmetic; the logic, SQL posture, concurrency model, and test coverage are all correct.

eloylp:
Addressed the three low review cleanup findings in `8576706`:

- removed duplicate `WorkspaceID` / repo fields from the artifact capture call sites, leaving `captureArtifact` as the single owner for those fields
- clarified the artifact resolution step comments around direct review lookup
- simplified the invalid-signature tamper expression

Verified locally:

- `go test ./internal/observe/... ./internal/webhook/... ./internal/store/...`
- `go test ./...`

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings

- None.

### Summary

Cleanup commit `8576706` addresses the prior low findings; the artifact reverse-index implementation still satisfies issue #692 and CI is green. Local verification: `go test -race ./internal/observe/... ./internal/webhook/...` passed as part of the focused run; `go test ./internal/store/...` passed. The combined `go test -race ./internal/observe/... ./internal/webhook/... ./internal/store/...` run timed out in `internal/store` at Go's 10-minute package timeout after observe/webhook had passed.

eloylp:
Requested change before merge: please move the new raw SQL/persistence operations for `run_attribution_artifacts` into `internal/store`.

The feature shape is correct, but the current implementation puts new SQL in `internal/observe/attribution_artifacts.go`. We have intentionally been keeping SQL centralized in `internal/store` for consistency of schema ownership, migrations, scans, transaction behavior, and future refactors. I agree that packages owning their own persistence can be a valid design in general, but that is not the architecture we want to extend right now. Existing SQL outside `internal/store` is historical convenience debt, not precedent.

Please adjust this PR as follows:

- Keep the resolver/domain behavior in `internal/observe` if that is still the clearest owner.
- Move the artifact table row types and SQL primitives into `internal/store`:
  - upsert artifact
  - lookup by issue comment id
  - lookup by PR review comment id
  - lookup by PR review id
  - lookup by PR/file/commit context
  - row scanning
- `internal/observe` should call typed store APIs instead of using `s.db.Exec`, `s.db.Query`, or `s.db.QueryRow` for the new artifact table.
- Keep `run_attributions` canonical; the artifact table remains only a GitHub artifact -> `span_id` reverse index.
- Preserve the existing tests and add/adjust tests at the store boundary if needed.
- Update `AGENTS.md` and `CLAUDE.md` in this PR to record the rule: all new SQL belongs in `internal/store`; non-store SQL is legacy debt, not a pattern to extend.

This is release-blocking because the self-improvement feature is about to become a flagship surface, and we do not want to introduce more persistence-boundary drift right before launch.

eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[blocking]** `internal/observe/attribution_artifacts.go:1` — Maintainer guidance in the latest conversation comment requires the new raw SQL and persistence primitives for `run_attribution_artifacts` to move into `internal/store`, with `internal/observe` calling typed store APIs instead of owning `s.db.Exec`, `s.db.Query`, or `s.db.QueryRow` for the artifact table. The current PR still adds the artifact row types, upsert, lookup, and scanning SQL in `internal/observe/attribution_artifacts.go`, and the docs update currently reinforces `observe` as the owner. That misses the release-blocking architecture requirement.

### Summary
The feature behavior remains on target, but the latest maintainer request narrows the acceptable implementation boundary; this PR needs the artifact persistence moved into `internal/store` and the code map docs updated before it is merge-ready.

eloylp:
Addressed the release-blocking persistence-boundary request in `ef8d55a`:

- moved `run_attribution_artifacts` row types, upsert, lookups, and scanning into `internal/store`
- kept `internal/observe` responsible for signature/context validation and artifact-chain resolution while calling typed store APIs for the new table
- added store-boundary coverage for idempotent upsert, comment/review/review-comment lookups, PR context lookup, and review-vs-review-comment isolation
- updated `AGENTS.md` and `CLAUDE.md` to record that new SQL belongs in `internal/store` and non-store SQL is legacy debt, not precedent

Verified locally:

- `go test ./internal/store/... ./internal/observe/... ./internal/webhook/...`
- `go test ./...`
- `go test -race ./internal/observe/... ./internal/store/...` passed `internal/observe`, then timed out in the existing `internal/store` suite at Go's 10-minute package timeout
- `go test -race ./internal/store -run TestRunAttributionArtifactUpsertAndLookups`

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The requested persistence-boundary refactor is resolved: `run_attribution_artifacts` row types, scans, upsert, and lookup SQL now live in `internal/store`, while `internal/observe` validates attribution metadata and calls typed store APIs. The store-boundary docs and tests are in place, and issue #692's artifact-ancestry goals remain covered.

Verified locally:
- `go test ./internal/store/... ./internal/observe/... ./internal/webhook/...`
- `go test -race ./internal/store -run TestRunAttributionArtifactUpsertAndLookups`

PR #700, opened 2026-06-18
fix: prefer scoped skill catalog refs

### LINKED ISSUE #699: Fix ambiguous catalog skill resolution for scoped assets
## Problem

While preparing the self-improvement demo workspace, `demo-coder` was configured with `skills: ["go-api"]` in workspace `self-improvement-demo`.

There are two visible skills named/ref-like `go-api`:

- global `go-api`, which contains the handler-shape rule: one handler per HTTP method, no method branching
- workspace-scoped `self-improvement-demo/go-api`, which intentionally omits that rule for the demo baseline

At runtime, the composed prompt used the global `go-api` skill instead of the workspace-scoped demo skill. That made the demo agent already know the improvement we intended the self-improvement loop to discover.

## Expected behavior

Catalog resolution must be deterministic and safe when multiple visible assets share a display name/ref.

For agent skill references, the system should either:

1. Prefer the most specific visible scoped asset for the agent context, e.g. repo-scoped > workspace-scoped > global, or
2. Reject ambiguous name-based references and require a stable skill id / explicit scope.

The chosen behavior should be consistent across:

- runtime prompt rendering
- REST CRUD surfaces
- MCP CRUD surfaces
- UI agent editor selectors
- config import/export

## Why this matters

The intelligence catalog supports global, workspace, and repo-scoped assets. If a workspace-scoped skill can be silently shadowed by a global skill with the same name, users can believe an agent is using one instruction set while runtime composes another.

For self-improvement, this can cause wrong attribution and bad recommendations because the analyst sees behavior produced by an unintended prompt/skill version.

## Reproduction notes

Current live demo state before workaround:

- workspace: `self-improvement-demo`
- agent: `demo-coder`
- agent skills were configured as `["go-api"]`
- global `go-api` contains the method-specific handler rule
- workspace-scoped `go-api` does not
- trace `9066a442a0aa93c7` showed the composed prompt included the global `go-api` rule

Workaround applied for the demo:

- created workspace skill `demo-go-api-baseline`
- updated `demo-coder` to use stable skill id `skill_self-improvement-demo_demo-go-api-baseline`

The workaround should not replace a proper resolver fix.

## Implementation guidance

Please inspect the current resolver path for agent skill refs and the UI/MCP/REST surfaces that save those refs.

A good implementation should include tests for at least:

- global and workspace skill share the same name; workspace agent resolves the workspace skill if specificity-preference is chosen
- repo-scoped skill wins over workspace/global in repo context if specificity-preference is chosen
- ambiguous references produce a clear validation error if explicit-scope mode is chosen
- import/export round-trips without losing the selected asset identity
- runtime composed prompt includes the intended skill version

Avoid ad-hoc resolver logic outside the existing catalog/service/store boundaries. Keep SQL operations inside `internal/store` as currently required by project guidance.

eloylp:
## Summary

- resolve agent skill refs across visible id, public ref, and display-name matches in one scoped query
- prefer the most specific visible skill scope: repo > workspace > global
- add regressions for workspace/global name collisions, repo-specific precedence, and a workspace display name shadowed by a global ref

## Verification

- go test ./internal/store
- go test ./...

Closes #699

### DIFF
--- a/internal/store/catalog_resolution.go
+++ b/internal/store/catalog_resolution.go
@@ -193,31 +193,35 @@ func catalogCandidatesInScope(q querier, table, selector, workspaceID, repo stri
 		SELECT id, ref, name, COALESCE(workspace_id, ''), COALESCE(repo, '')
 		FROM ` + table + `
 		WHERE (id = ? OR ref = ? OR name = ?)
-		  AND workspace_id IS ?
-		  AND repo IS ?`
-	var workspaceArg, repoArg any
+		  AND workspace_id ` + catalogScopePredicate(workspaceID) + `
+		  AND repo ` + catalogScopePredicate(repo)
+	args := []any{selector, selector, selector}
 	if workspaceID != "" {
-		query = strings.Replace(query, "workspace_id IS ?", "workspace_id = ?", 1)
-		workspaceArg = workspaceID
+		args = append(args, workspaceID)
 	}
 	if repo != "" {
-		query = strings.Replace(query, "repo IS ?", "repo = ?", 1)
-		repoArg = repo
+		args = append(args, repo)
 	}
 	if mode == catalogScopeWorkspace {
 		query = `
 		SELECT id, ref, name, COALESCE(workspace_id, ''), ''
 		FROM ` + table + `
 		WHERE (id = ? OR ref = ? OR name = ?)
-		  AND workspace_id IS ?`
-		workspaceArg = nil
+		  AND workspace_id ` + catalogScopePredicate(workspaceID)
+		args = []any{selector, selector, selector}
 		if workspaceID != "" {
-			query = strings.Replace(query, "workspace_id IS ?", "workspace_id = ?", 1)
-			workspaceArg = workspaceID
+			args = append(args, workspaceID)
 		}
-		return scanCatalogCandidates(q, query, selector, selector, selector, workspaceArg)
+		return scanCatalogCandidates(q, query, args...)
 	}
-	return scanCatalogCandidates(q, query, selector, selector, selector, workspaceArg, repoArg)
+	return scanCatalogCandidates(q, query, args...)
+}
+
+func catalogScopePredicate(value string) string {
+	if value == "" {
+		return "IS NULL"
+	}
+	return "= ?"
 }
 
 func scanCatalogCandidates(q querier, query string, args ...any) ([]catalogCandidate, error) {
--- a/internal/store/catalog_resolution.go
+++ b/internal/store/catalog_resolution.go
@@ -170,7 +170,7 @@ func resolveVisibleCatalogRef(q querier, table, ref, workspaceID, repo string) (
 		}
 		return "", sql.ErrNoRows
 	}
-	if exactID != "" && exactName != ref {
+	if exactID != "" && (exactName != ref || isGeneratedCatalogRef(table, ref)) {
 		return exactID, nil
 	}
 	if ambiguous {
@@ -178,6 +178,10 @@ func resolveVisibleCatalogRef(q querier, table, ref, workspaceID, repo string) (
 		return "", fmt.Errorf("ambiguous %s %q in workspace %q; use %s id", label, ref, workspaceID, label)
 	}
 	return bestID, nil
+}
+
+func isGeneratedCatalogRef(table, ref string) bool {
+	return strings.HasPrefix(ref, strings.TrimSuffix(table, "s")+"_")
 }
 
 func resolveVisibleCatalogExactRef(q querier, table, ref, workspaceID, repo string) (string, string, error) {
--- a/internal/store/catalog_resolution.go
+++ b/internal/store/catalog_resolution.go
@@ -4,7 +4,6 @@ import (
 	"crypto/rand"
 	"database/sql"
 	"encoding/hex"
-	"errors"
 	"fmt"
 	"strings"
 
@@ -114,14 +113,59 @@ func ensureCatalogScope(tx *sql.Tx, kind, workspaceID, repo string) error {
 }
 
 func resolveVisibleCatalogRef(q querier, table, ref, workspaceID, repo string) (string, error) {
-	id, err := resolveVisibleCatalogID(q, table, ref, workspaceID, repo)
-	if err == nil {
-		return id, nil
-	}
-	if !errors.Is(err, sql.ErrNoRows) {
+	workspaceID = fleet.NormalizeWorkspaceID(workspaceID)
+	repo = fleet.NormalizeRepoName(repo)
+	rows, err := q.Query(`
+		SELECT id,
+			CASE
+				WHEN workspace_id IS NULL THEN 0
+				WHEN repo IS NULL THEN 1
+				ELSE 2
+			END AS specificity
+		FROM `+table+`
+		WHERE (id = ? OR ref = ? OR name = ?)
+		  AND (
+			(workspace_id IS NULL AND repo IS NULL)
+			OR (workspace_id = ? AND repo IS NULL)
+			OR (? <> '' AND workspace_id = ? AND repo = ?)
+		  )
+		ORDER BY specificity DESC, id`,
+		ref, ref, ref, workspaceID, repo, workspaceID, repo,
+	)
+	if err != nil {
 		return "", err
 	}
-	return resolveVisibleCatalogName(q, table, strings.TrimSuffix(table, "s"), strings.TrimSuffix(table, "s")+" id", ref, workspaceID, repo)
+	defer rows.Close()
+
+	var bestID string
+	bestSpecificity := -1
+	ambiguous := false
+	for rows.Next() {
+		var id string
+		var specificity int
+		if err := rows.Scan(&id, &specificity); err != nil {
+			return "", err
+		}
+		if bestSpecificity == -1 {
+			bestID = id
+			bestSpecificity = specificity
+			continue
+		}
+		if specificity == bestSpecificity && id != bestID {
+			ambiguous = true
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return "", err
+	}
+	if bestSpecificity == -1 {
+		return "", sql.ErrNoRows
+	}
+	if ambiguous {
+		label := strings.TrimSuffix(table, "s")
+		return "", fmt.Errorf("ambiguous %s %q in workspace %q; use %s id", label, ref, workspaceID, label)
+	}
+	return bestID, nil
 }
 
 func resolveVisibleCatalogID(q querier, table, id, workspaceID, repo string) (string, error) {
--- a/internal/store/catalog_resolution.go
+++ b/internal/store/catalog_resolution.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"database/sql"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -115,6 +116,11 @@ func ensureCatalogScope(tx *sql.Tx, kind, workspaceID, repo string) error {
 func resolveVisibleCatalogRef(q querier, table, ref, workspaceID, repo string) (string, error) {
 	workspaceID = fleet.NormalizeWorkspaceID(workspaceID)
 	repo = fleet.NormalizeRepoName(repo)
+
+	exactID, exactName, err := resolveVisibleCatalogExactRef(q, table, ref, workspaceID, repo)
+	if err != nil && !errors.Is(err, sql.ErrNoRows) {
+		return "", err
+	}
 	rows, err := q.Query(`
 		SELECT id,
 			CASE
@@ -123,14 +129,14 @@ func resolveVisibleCatalogRef(q querier, table, ref, workspaceID, repo string) (
 				ELSE 2
 			END AS specificity
 		FROM `+table+`
-		WHERE (id = ? OR ref = ? OR name = ?)
+		WHERE name = ?
 		  AND (
 			(workspace_id IS NULL AND repo IS NULL)
 			OR (workspace_id = ? AND repo IS NULL)
 			OR (? <> '' AND workspace_id = ? AND repo = ?)
 		  )
 		ORDER BY specificity DESC, id`,
-		ref, ref, ref, workspaceID, repo, workspaceID, repo,
+		ref, workspaceID, repo, workspaceID, repo,
 	)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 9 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[major]** `internal/store/catalog_resolution.go:126` — This combines exact `id`/`ref` matches with display-name matches and then chooses by scope specificity, so an explicit stable skill reference can be redirected to a more-specific skill whose `name` happens to equal that reference. For example, a workspace agent configured with the exported/global skill ref `skill_global_shared` would resolve to a workspace skill named `skill_global_shared` if one exists, even though the stored/exported ref is meant to preserve the selected asset identity. Issue #699 requires import/export and REST/MCP saved refs to stay consistent with runtime resolution; this makes stable refs from those surfaces non-authoritative under a same-string name collision. The resolver needs to preserve exact stable-id/ref semantics while still preferring the scoped display-name match for the intended `go-api` collision case.

### Summary
The scoped-name preference is implemented and covered, but exact saved skill refs can now be shadowed by display names, which breaks the stable-id/import-export part of the issue goal.

eloylp:
Addressed the review finding in commit `45f481c`.

The resolver now checks visible exact `id`/`ref` matches separately from display-name matches. Exact stable refs remain authoritative when the matched catalog asset has a different display name, so a saved/exported ref like `skill_global_shared` cannot be shadowed by a scoped skill named `skill_global_shared`. Name-style references such as `go-api` still use the most-specific visible display-name match.

Added `TestAgentSkillStableRefBeatsMoreSpecificDisplayName` for the reviewer scenario.

Verification:
- `go test ./internal/store`
- `go test ./...`
- `go test ./internal/store -race -run 'TestAgentSkill(DisplayName|StableRef)'`

Note: `go test ./internal/store -race` was also attempted, but the full package race run hit Go’s 10-minute timeout in unrelated CRUD tests during SQLite migration setup; no race report was emitted before timeout.

eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[major]** `internal/store/catalog_resolution.go:173` — Exact saved refs are still not always authoritative. The new guard only returns the exact match when `exactName != ref`, but imported/exported skills commonly default `name` to the stable ref when no explicit display name is set (`importSkills` does this before saving). In that case a saved ref like `skill_global_shared` can still be redirected to a workspace skill whose display name is `skill_global_shared`, because `exactName == ref` falls through to the display-name winner. Issue #699 requires import/export round trips and saved REST/MCP refs to preserve the selected asset identity; this still breaks that case. The resolver needs a more direct way to distinguish stable id/ref inputs from name-style inputs such as `go-api`, rather than relying only on whether the exact match's display name differs from the ref.

### Summary
The previous shadowing case is narrowed and the focused store tests pass, but stable refs whose default display name equals the ref can still be shadowed by scoped display-name matches.

eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- None new. The current-SHA blocking finding already posted for `internal/store/catalog_resolution.go:173` remains unresolved; I verified the same path against `importSkills`, where empty skill names default to the stable ref.

### Summary
No additional issues found beyond the existing stable-ref shadowing blocker; focused verification passed with `go test ./internal/store -run 'TestAgentSkill(DisplayName|StableRef)'`.

eloylp:
Addressed the remaining stable-ref shadowing blocker in commit `1ea37fe`.

The resolver now treats generated catalog refs such as `skill_global_shared` as exact selectors even when their display name defaults to the same value, so import/export and saved REST/MCP refs remain authoritative under a same-string scoped display-name collision. Legacy name-style selectors such as `go-api` still use the scoped display-name precedence covered by the existing tests.

Added `TestAgentSkillGeneratedRefBeatsMoreSpecificDefaultDisplayName` for the import-default-name case.

Verification:
- `go test ./internal/store -run 'TestAgentSkill(DisplayName|StableRef|GeneratedRef)'`
- `go test ./...`
- `go test ./internal/store -race -run 'TestAgentSkill(DisplayName|StableRef|GeneratedRef)'`

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The latest commit preserves stable generated refs while keeping scoped display-name precedence for name-style skill references, with focused regressions for the previously blocking cases.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
Re-reviewed after the latest conversation update; the current resolver keeps generated stable refs authoritative while preserving scoped display-name precedence for name-style references. Verified with `go test ./internal/store -run 'TestAgentSkill(DisplayName|StableRef|GeneratedRef)' -timeout 120s`.\n\n

internal/store/catalog_resolution.go
```diff
@@ -114,14 +114,98 @@ func ensureCatalogScope(tx *sql.Tx, kind, workspaceID, repo string) error {
...
+			END AS specificity
+		FROM `+table+`
+		WHERE name = ?
+		  AND (
+			(workspace_id IS NULL AND repo IS NULL)
+			OR (workspace_id = ? AND repo IS NULL)
+			OR (? <> '' AND workspace_id = ? AND repo = ?)
+		  )
+		ORDER BY specificity DESC, id`,
+		ref, workspaceID, repo, workspaceID, repo,
```
eloylp: /agents improve , we should prefer the dabase to be as dumb as possible. Thus this selection logic should be in the code rather than SQL

eloylp:
This fix is on the right track, but for the release we should widen the scope so catalog scoping is consistent across the whole intelligence catalog surface, not only agent skill refs.

The underlying problem is that scoped catalog assets make name-based selectors ambiguous. A selector like `go-api` can refer to a global asset, a workspace asset, or a repo asset. Users need deterministic behavior across import, REST, MCP, UI saves, and runtime prompt rendering.

Please extend the implementation with a shared resolution policy for catalog assets:

1. Exact internal id / public ref wins if the asset is visible from the caller context.
2. Display-name selectors resolve by most specific visible scope.
3. Same-scope duplicate matches remain ambiguous and should fail with a clear validation error.
4. Explicit prompt scope, where present, remains authoritative and constrains lookup.
5. Runtime composition must use the same resolved asset identity that import/CRUD saved.

Scope policy by asset kind:

- Prompts: repo > workspace > global.
- Skills: repo > workspace > global.
- Guardrails: workspace > global. Guardrails are not repo-scoped today, so they do not need repo-specific resolution, but they should still follow the same exact-ref-first / deterministic display-name behavior.

Implementation preference: keep the database as dumb as possible. SQL should fetch visible candidates only. The domain decision should live in Go, with small testable helpers, for example:

```go
type catalogCandidate struct {
    ID          string
    Ref         string
    Name        string
    WorkspaceID string
    Repo        string
}

func chooseCatalogCandidate(candidates []catalogCandidate, selector, workspaceID, repo string, mode catalogScopeMode) (catalogCandidate, error)
func catalogSpecificity(c catalogCandidate, mode catalogScopeMode) int
```

Avoid encoding the actual policy in SQL with `CASE ... specificity`. It is harder to inspect, harder to reuse for prompts/guardrails, and easier for each resolver to drift. The query can still filter by visibility for efficiency, but Go should decide exact-vs-name precedence, specificity, and ambiguity.

Suggested shape:

- Keep domain-specific wrappers readable:
  - `resolveVisibleSkillRef(...)`
  - `resolveVisiblePromptRef(...)`
  - `resolveVisibleGuardrailRef(...)`
- Internally share candidate fetching/selection helpers.
- Add regression tests for prompts and guardrails in addition to the existing skill cases.
- Add at least one runtime/render-oriented assertion showing the composed prompt uses the intended scoped skill/prompt body, not only that import/load round-trips the ref.

This is release-critical because the self-improvement flow relies on attribution to the exact prompt/skill/guardrail version that produced behavior. If scoping silently picks the wrong catalog asset, the analyst will generate plausible but wrong recommendations.

eloylp:
I started addressing the release-critical shared catalog resolution request, but this runner is blocked before commit/push because the filesystem is full. The local checkout has the intended patch staged conceptually across prompt, skill, guardrail, and runtime prompt resolution, and `git diff --check` passed, but both Go tests and `git commit` fail with `no space left on device` before Git can update its index.

No remote changes were pushed from this run.

eloylp:
Addressed the release-scope catalog resolution feedback in commit `35f85ae`.

Changes:
- Added a shared Go-side catalog candidate chooser for visible prompt/skill/guardrail candidates.
- Kept SQL to candidate retrieval and moved exact-ref, display-name specificity, and same-scope ambiguity policy into `chooseCatalogCandidate` / `catalogSpecificity`.
- Applied the shared policy to prompt refs, skill refs, explicit prompt scope lookup, and workspace guardrail refs.
- Added regressions for prompt and guardrail collisions plus a render-time assertion that composed prompts use the resolved scoped prompt/skill bodies.

Verification:
- `gofmt -w internal/store/catalog_resolution.go internal/store/agents.go internal/store/workspaces_import_load.go internal/store/store_test.go`
- `git diff --check`

Blocked locally:
- `go test ./internal/store -run 'TestAgent(Skill|Prompt)|TestWorkspaceGuardrailReference'` could not run because the container filesystem is 100% full and Go cannot create a build cache (`no space left on device`).

eloylp:
Fixed the CI failures from commit `35f85ae` in follow-up commit `597dc2a`.

The failing store test still expected global+workspace prompt display-name matches to be ambiguous; the release policy now says display-name selectors use the most specific visible scope, so the test now asserts workspace selection. The MCP replace-import fixture now also carries the required prompt catalog refs in its config fixture.

Verification:
- `gofmt -w internal/store/store_test.go internal/mcp/tools_test.go`
- `git diff --check`

Full local tests remain blocked by the runner filesystem/cache space issue; GitHub CI is the source of truth for this follow-up.

eloylp:
Fixed the remaining PR #700 CI failure in commit `1452ba8`.

Root cause: exported agents carry `prompt_scope: global`, which uses the exact-scope catalog lookup. That query used `workspace_id IS ?` / `repo IS ?` with nil parameters for global scope, and the replace-import path failed to find the global prompt. The query now emits explicit `IS NULL` predicates for empty scope components.

Verification:
- `gofmt -w internal/store/catalog_resolution.go`
- `git diff --check`

eloylp:
Added commit `e132dc2` for the remaining MCP replace-import failure.

Export now omits redundant `prompt_scope: global` for globally scoped prompt refs. Global prompt lookup is already the default visible lookup; workspace and repo prompt scopes are still exported explicitly.

Verification:
- `gofmt -w internal/daemon/config/config.go`
- `git diff --check`

eloylp:
Daemon smoke testing found one remaining catalog-resolution edge outside the main agent prompt/skill resolver.

Important product constraint to preserve: exported config should not export historical catalog versions anymore. Since we removed version pinning, export/import should only carry the current/latest asset body for prompts, skills, and guardrails. That keeps config export aligned with the current live catalog and avoids reopening the old version-pinning can of worms.

What passed locally on the PR branch:
- started an isolated daemon from PR #700 on a clean DB
- imported global/workspace/repo prompt + skill name collisions
- repo-scoped agent resolved correctly to the repo-specific prompt and skill
- export then replace-import round-tripped for that repo-scoped current-body case

Remaining issue found with an exact global-ref collision:
- global skill ref: 
- workspace skill display name: 
- initial import works
- exported YAML strips  correctly
- replace-importing the exported YAML failed with 

The exported YAML showed the global skill carrying a version snapshot ID from the workspace-collision skill. This points at the version/export helper path, not the main agent resolver:  and  in  still resolve with , so stable refs can be shadowed by colliding display names. If we keep any version listing helpers for UI diff/history, they need exact id/ref-first semantics too. For export specifically, I think the cleaner fix is to stop exporting  entirely and only export the current/latest asset body.

eloylp:
Correction to my previous comment: the shell interpreted the inline backticks before `gh` sent the body. Here is the intended finding.

Daemon smoke testing found one remaining catalog-resolution edge outside the main agent prompt/skill resolver.

Important product constraint to preserve: exported config should not export historical catalog versions anymore. Since we removed version pinning, export/import should only carry the current/latest asset body for prompts, skills, and guardrails. That keeps config export aligned with the current live catalog and avoids reopening the old version-pinning can of worms.

What passed locally on the PR branch:
- Started an isolated daemon from PR #700 on a clean DB.
- Imported global/workspace/repo prompt + skill name collisions.
- Repo-scoped agent resolved correctly to the repo-specific prompt and skill.
- Export then replace-import round-tripped for that repo-scoped current-body case.

Remaining issue found with an exact global-ref collision:
- Global skill ref: `skill_global_policy`.
- Workspace skill display name: `skill_global_policy`.
- Initial import works.
- Exported YAML strips `prompt_scope: global` correctly.
- Replace-importing the exported YAML failed with `UNIQUE constraint failed: skill_versions.id`.

The exported YAML showed the global skill carrying a version snapshot ID from the workspace-collision skill. This points at the version/export helper path, not the main agent resolver: `skillInternalID` and `guardrailInternalID` in `internal/store/catalog_versions.go` still resolve with `WHERE id=? OR ref=? OR name=?`, so stable refs can be shadowed by colliding display names. If we keep any version listing helpers for UI diff/history, they need exact id/ref-first semantics too.

For export specifically, I think the cleaner fix is to stop exporting `versions:` entirely and only export the current/latest asset body.

eloylp:
Additional smoke result after the earlier export note:

- Runtime resolution was tested by firing an on-demand agent against a synthetic catalog with colliding `shared` prompt and skill names at global, workspace, and repo scope.
- The persisted trace prompt included only the repo-scoped bodies:
  - `REPO PROMPT BODY SHOULD APPEAR`
  - `REPO SKILL BODY SHOULD APPEAR`
- The trace prompt did not include the global/workspace prompt or skill bodies, so the actual run-time renderer resolves the scoped catalog refs correctly.

Product rule to keep explicit in this PR: past catalog versions should not be exportable. Since version pinning was removed, export/import should deal only with the current/latest asset body and metadata. Historical versions are UI/runtime audit data, not portable config.

The remaining issue I found is therefore on the export/version-listing side, not the runtime renderer: exported skill snapshots can still be pulled through the old `id OR ref OR name` resolver path when display names collide. The clean release fix is to stop exporting historical `versions:` at all, and keep any UI history helpers on exact id/ref-first resolution.

eloylp:
Addressed the remaining export/version-listing catalog feedback in commit `5da21e9`.

Changes:
- Config export no longer serializes historical catalog `versions:` snapshots; exported YAML carries the current/latest prompt, skill, and guardrail bodies plus current metadata only.
- Skill and guardrail version-listing helpers now resolve exact internal ids/public refs before falling back to display-name lookup, so a stable ref cannot be shadowed by a colliding scoped display name.
- Added a regression for exact skill/guardrail refs under display-name collisions and updated export tests for the current-only product rule.

Verification:
- `go test ./internal/store -run 'TestCatalogVersionSnapshotsUseExactRefBeforeDisplayName|TestAgent(Skill|Prompt)|TestWorkspaceGuardrailReference'`
- `go test ./internal/daemon -run 'TestStoreImportYAMLPreservesCatalogVersions|TestToolExportConfigReturnsYAML|TestStoreExportReturnsYAML'`
- `go test ./...`
- `go test ./internal/store -race -run 'TestCatalogVersionSnapshotsUseExactRefBeforeDisplayName|TestAgent(Skill|Prompt)|TestWorkspaceGuardrailReference'`

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The export/version-listing follow-up satisfies the maintainer constraint: config export no longer serializes historical catalog `versions:`, while skill and guardrail version helpers resolve exact internal ids/public refs before display-name fallback under scoped display-name collisions. Verified locally with `go test ./internal/store -run 'TestCatalogVersionSnapshotsUseExactRefBeforeDisplayName|TestAgent(Skill|Prompt)|TestWorkspaceGuardrailReference' -timeout 120s` and `go test ./internal/daemon -run 'TestStoreImportYAMLPreservesCatalogVersions|TestToolExportConfigReturnsYAML|TestStoreExportReturnsYAML' -timeout 180s`.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The updated PR satisfies the expanded catalog-resolution scope: exact ids/refs win, display-name selectors choose the most specific visible scope for prompts/skills/guardrails, same-scope ambiguity is handled, export is current-only, and runtime prompt rendering uses the resolved scoped bodies. Verified locally with focused store/export tests and `go test ./...`.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The current #700 head satisfies the expanded catalog-resolution scope: exact ids/refs win, display-name selectors resolve by visible scope specificity for prompts, skills, and guardrails, same-scope ambiguity returns validation errors, explicit prompt scope constrains lookup, runtime rendering uses the resolved scoped bodies, and config export is current-only. Verified locally with `go test ./internal/store -run 'TestCatalogVersionSnapshotsUseExactRefBeforeDisplayName|TestAgent(Skill|Prompt)|TestWorkspaceGuardrailReference|TestRuntimePromptUsesResolvedScopedCatalogBodies' -timeout 120s` and `go test ./internal/daemon -run 'TestStoreImportYAMLPreservesCatalogVersions|TestToolExportConfigReturnsYAML|TestStoreExportReturnsYAML' -timeout 180s`; GitHub reports the `test` check green on `5da21e98407425887f2767caa265004b4e7bd41e`.

PR #708, opened 2026-06-18
fix: harden SQLite timestamp scans

### LINKED ISSUE #706: Audit SQLite time-field scanning and query robustness
## Problem

We recently found a production-impacting bug where a SQLite timestamp value stored as text could not be scanned into time.Time reliably. The failure made attribution artifact lookups silently miss existing rows, which then caused self-improvement attribution to degrade to unresolved/needs-user-input even though the artifact was present.

This may not be isolated. Any store query that scans SQLite date/time text directly into time.Time, or relies on inconsistent timestamp formats, can produce subtle false negatives or empty results.

## Goal

Audit the codebase for SQLite queries and scan paths that may mishandle time fields, then fix them consistently.

## Scope

- Search the store and daemon persistence code for every query that reads timestamp columns.
- Identify all rows.Scan / row.Scan calls where SQLite timestamp columns are scanned into time.Time, *time.Time, or shapes that could fail on text formats.
- Verify timestamp columns written by migrations and inserts use consistent formats.
- Fix scan helpers so nullable and non-nullable timestamps are decoded consistently from SQLite-supported representations.
- Add regression tests covering the specific malformed/driver-produced timestamp shapes that caused the attribution artifact miss.
- Prefer central helpers in internal/store for timestamp scanning/parsing. Do not scatter ad-hoc parsing across feature packages.

## Implementation notes

- Keep SQL operations inside the store boundary. Feature packages should not add direct SQLite queries.
- Check attribution-related tables first, especially run_attributions, run_attribution_artifacts, improvement feedback/recommendations/bundles, traces, runners/event queue, memory, and auth/session tables.
- The bug class to prevent is: row exists, but timestamp scan/parsing failure causes lookup/list APIs to behave as if the row did not exist.
- Where a timestamp is only for display, prefer returning a parsed zero-safe value or raw string intentionally rather than failing the whole business lookup.
- Add structured logs for non-happy parsing paths where the row can be safely skipped, but do not hide parser errors on data required for correctness.

## Acceptance criteria

- All SQLite timestamp scan paths are audited and either use shared parsing helpers or are documented as safe.
- Regression tests prove attribution artifact lookup works when SQLite returns timestamp text values like the one previously observed.
- Tests cover at least one nullable timestamp column and one non-null timestamp column.
- No new direct database logic is introduced outside internal/store.
- The issue leaves no known timestamp scan paths with silent false-negative behavior.

eloylp:
## Summary

- add a shared SQLite timestamp scanner for store/observe reads that accepts driver time.Time values, RFC3339 text, SQLite datetime text, and legacy zone-formatted strings
- use the scanner for auth, memory, runner, observe event/trace/dispatch, run attribution, and self-improvement timestamp reads
- normalize new attribution/event timestamp writes to RFC3339Nano text and keep event since filtering compatible with older timestamp text

## Verification

- `go test ./internal/store -run 'TestSQLiteTime|TestRunAttributionArtifact'`
- `go test ./internal/store ./internal/observe -race -run 'TestSQLiteTime|TestRunAttributionArtifact|TestReadWriteMemory|TestAuthBootstrapLoginAndAPITokenLifecycle|TestStoreListEventsSinceFilter'`
- `go test ./internal/mcp -run TestToolListEventsSinceFilter`
- `go test ./internal/observe -run TestStoreListEventsSinceFilter`
- `go test ./...`

Closes #706

### DIFF
--- a/internal/observe/attribution.go
+++ b/internal/observe/attribution.go
@@ -577,11 +577,15 @@ func (s *Store) inferRunAttributions(workspaceID string, q AttributionQuery) []A
 		FROM run_attributions
 		WHERE workspace_id=? AND repo_owner=? AND repo_name=? AND issue_or_pr_number=?
 		  AND (?='' OR head_sha=?)
-		  AND created_at BETWEEN ? AND ?
-		ORDER BY created_at DESC`,
+		  AND (
+			created_at BETWEEN ? AND ?
+			OR replace(substr(created_at, 1, 19), 'T', ' ') BETWEEN ? AND ?
+		  )
+		ORDER BY replace(substr(created_at, 1, 19), 'T', ' ') DESC, created_at DESC`,
 		workspaceID, q.RepoOwner, q.RepoName, q.IssueOrPRNumber,
 		strings.TrimSpace(q.HeadSHA), strings.TrimSpace(q.HeadSHA),
 		from.UTC().Format(time.RFC3339Nano), to.UTC().Format(time.RFC3339Nano),
+		from.UTC().Format(time.DateTime), to.UTC().Format(time.DateTime),
 	)
 	if err != nil {
 		return nil
--- a/internal/observe/attribution.go
+++ b/internal/observe/attribution.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/eloylp/agents/internal/fleet"
+	storepkg "github.com/eloylp/agents/internal/store"
 	"github.com/eloylp/agents/internal/workflow"
 )
 
@@ -513,7 +514,7 @@ func (s *Store) recordRunAttribution(in workflow.SpanInput, createdAt time.Time)
 		a.SpanID, a.WorkspaceID, a.RepoOwner, a.RepoName, a.IssueOrPRNumber,
 		a.EventID, a.EventQueueID, a.AgentID, a.AgentName, a.BackendID, a.BackendName,
 		nullString(a.PromptVersionID), a.PromptRef, strings.Join(a.SkillVersionIDs, ","), strings.Join(a.GuardrailVersionIDs, ","),
-		a.HeadSHA, a.Branch, a.CreatedAt,
+		a.HeadSHA, a.Branch, a.CreatedAt.UTC().Format(time.RFC3339Nano),
 	)
 	if err != nil {
 		log.Printf("observe: persist run attribution %s: %v", in.SpanID, err)
@@ -580,7 +581,7 @@ func (s *Store) inferRunAttributions(workspaceID string, q AttributionQuery) []A
 		ORDER BY created_at DESC`,
 		workspaceID, q.RepoOwner, q.RepoName, q.IssueOrPRNumber,
 		strings.TrimSpace(q.HeadSHA), strings.TrimSpace(q.HeadSHA),
-		from, to,
+		from.UTC().Format(time.RFC3339Nano), to.UTC().Format(time.RFC3339Nano),
 	)
 	if err != nil {
 		return nil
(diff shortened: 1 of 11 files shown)


eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[major]** `internal/observe/attribution.go:580` — Issue #706 requires mixed SQLite timestamp formats to avoid silent false negatives, but loose run-attribution inference still filters `run_attributions.created_at` with `created_at BETWEEN ? AND ?` while binding RFC3339Nano bounds. Legacy rows written as SQLite datetime text on the same date can sort before the lower bound because the stored value has a space at byte 11 and the bound has `T` (`"2026-06-18 12:00:00" < "2026-06-18T11:00:00Z"` lexically), so a row inside the attribution window can be excluded before `SQLiteTime` gets a chance to parse it. This leaves the same unresolved-attribution failure mode for older `run_attributions` rows even though new writes are normalized; the query needs the same kind of mixed-format normalization added for event `since` filtering, plus a regression covering legacy `created_at` text inside a same-day inference window.

### Summary
The shared scanner and artifact/event coverage are in good shape, but the attribution inference query still has a mixed-format timestamp false-negative path that blocks the linked issue goal.

eloylp:
Addressed the attribution inference review finding in commit `bf58bd6`.

Loose run attribution inference now checks both the RFC3339 range and a normalized SQLite datetime range for `run_attributions.created_at`, so legacy same-day `YYYY-MM-DD HH:MM:SS` rows are not filtered out before `SQLiteTime` can parse them. Added `TestResolveRunAttributionInfersLegacySQLiteDatetimeText` for the reviewer scenario.

Verification:
- `go test ./internal/observe -run 'TestResolveRunAttribution(Inferred|InfersLegacy)'`
- `go test ./internal/observe -race -run 'TestResolveRunAttribution(Inferred|InfersLegacy)'`
- `go test ./...`

eloylp:
## PR Review — pr-reviewer

**Verdict:** REQUEST_CHANGES

### Findings
- **[major]** `internal/store/runners.go:331` — Issue #706 calls out the runner/event queue timestamp paths and requires leaving no known SQLite timestamp comparison with mixed-format false-negative behavior, but cleanup still compares `event_queue.completed_at < ?` against an RFC3339Nano bound. A legacy completed row stored as SQLite datetime text on the same day can compare before an earlier RFC3339 cutoff purely because `' '` sorts before `'T'` (`"2026-06-18 15:00:00" < "2026-06-18T14:00:00Z"`), so `DeleteCompletedEventsBefore` can prune a completed event that is newer than the retention cutoff. This needs the same normalized mixed-format predicate used for event listing / run-attribution inference, plus a regression with a same-day legacy `completed_at` value that should survive cleanup.

### Summary
The previous attribution-inference blocker is fixed and the focused tests pass, but the event-queue cleanup path still has a mixed-format timestamp comparison that can delete rows outside the cutoff.

Verification:
- `go test ./internal/store -run 'TestSQLiteTime|TestRunner|TestEventQueue|TestReadWriteMemory|TestAuthBootstrapLoginAndAPITokenLifecycle'`
- `go test ./internal/observe -run 'TestResolveRunAttribution(Inferred|InfersLegacy)|TestStoreListEventsSinceFilter'`

eloylp:
Addressed the event queue cleanup review finding in commit `843cab8`.

`DeleteCompletedEventsBefore` now compares `completed_at` through the same normalized `YYYY-MM-DD HH:MM:SS` shape used for mixed SQLite/RFC3339 timestamp filtering, so a same-day legacy SQLite datetime row newer than the cutoff is not pruned just because `' '` sorts before `'T'`. Added `TestDeleteCompletedEventsBeforeKeepsSameDayLegacyDatetimeAfterCutoff` to cover legacy and RFC3339 rows on both sides of the cutoff.

Verification:
- `go test ./internal/store -run 'TestDeleteCompletedEventsBefore|TestSQLiteTime'`
- `go test ./internal/store -run 'TestSQLiteTime|TestRunner|TestEventQueue|TestDeleteCompletedEventsBefore|TestReadWriteMemory|TestAuthBootstrapLoginAndAPITokenLifecycle'`
- `go test ./internal/observe -run 'TestResolveRunAttribution(Inferred|InfersLegacy)|TestStoreListEventsSinceFilter'`
- `go test ./internal/store -race -run 'TestDeleteCompletedEventsBefore|TestSQLiteTime'`
- `go test ./...`

CI is pending on the updated head at the time of this comment.

eloylp:
## PR Review — pr-reviewer

**Verdict:** APPROVE

### Findings
- None.

### Summary
The latest head fixes the remaining event-queue cleanup mixed-format comparison, and the PR now covers the linked #706 timestamp scan/query goals with centralized parsing plus focused regressions for nullable and non-null timestamp paths.

Verification:
- `go test ./internal/store -run 'TestDeleteCompletedEventsBefore|TestSQLiteTime|TestRunAttributionArtifact|TestReadWriteMemory|TestAuthBootstrapLoginAndAPITokenLifecycle'`
- `go test ./internal/observe -run 'TestResolveRunAttribution(Inferred|InfersLegacy)|TestStoreListEventsSinceFilter'`
- `go test ./internal/mcp -run TestToolListEventsSinceFilter`
- `go test ./internal/store -race -run 'TestDeleteCompletedEventsBefore|TestSQLiteTime'`

